PR #175, opened 2024-09-19
feat: splashscreen

Llandy3d:
Adds a splashscreen to show while the app is loading.

A new event `splashscreen:close` is created that can be sent from the frontend via `window.studio.app.closeSplashscreen()`.

To complete this we need the splashscreen and a way to detect from the frontend that the app is loaded so that we can close it and show the main window.


cristianoventura:
The splash screen handles light and dark mode. However, it shows up pretty quickly since it's only waiting for the main window and the project structure to be configured. We can add an extra time out if we want it to show for a longer period.

You can uncomment L158 of `main.ts` to see how it looks like.

https://github.com/user-attachments/assets/aa419561-98b0-45c1-9055-6c50046cfe15

src/main.ts
```diff
@@ -127,6 +159,10 @@ app.whenReady().then(async () => {
   watcher.on('unlink', (filePath) => {
     mainWindow.webContents.send('ui:remove-file', path.basename(filePath))
   })
+
+  splashscreenWindow.close()
+  mainWindow.show()
+  mainWindow.focus()
```
Llandy3d, line 165: This won't work, we will need to use the `splashscreen:close` event to make sure we show a ready page.
So we will need to detect that the react page is loaded and then we can send the event from there to trigger the hiding of the splashscreen and the showing of the main window

e-fisher (approved):
LGTM! 🚀

PR #222, opened 2024-09-25
fix: Requests falsely treated as static assets

e-fisher:
This PR fixes 2 issues:

1. Inconsistency of URL path between data from proxy and loaded har file: proxy would return path including query params, while `harToProxyData` returned path only. 
2. The `isNonStaticAssetResponse` not striping query params from path and falsely treating URLs as static assets, for example: test.k6.io/home?foo=bar.js

How to test:

- Use the attached har file to verify the request to `https://test.k6.io/?foo=bar.js` is no longer treated as a static asset.
- Create a generator and verify it's not treated as static assets in Validator either

[2024-09-25_06-32-30.har.zip](https://github.com/user-attachments/files/17126179/2024-09-25_06-32-30.har.zip)

### DIFF
--- a/src/components/WebLogView/Row.tsx
+++ b/src/components/WebLogView/Row.tsx
@@ -4,7 +4,6 @@ import { ProxyData } from '@/types'
 
 import { MethodBadge } from '../MethodBadge'
 import { ResponseStatusBadge } from '../ResponseStatusBadge'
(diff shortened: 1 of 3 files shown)


Llandy3d (commented):
~In the validator it still seems to be treated as a static file~ 🤔

It works correctly, I tested the wrong branch 💤 

image

Llandy3d (approved):
🙌 😴

Issue #207, opened 2024-09-23
Request inspection does not work in rule preview for the replaced responses


Llandy3d:
Inspecting a request where a value is extracted in the `rule preview` pane with a correlation rule works without issues.
The same is not valid for the responses that are replaced, clicking on them does not show the detail view for the response

![Image](https://github.com/user-attachments/assets/212ef97c-a4cd-46c3-96d7-e506724579be)

PR #223, opened 2024-09-25
fix: Inspect requests matched by correlation

e-fisher:
Resolves https://github.com/grafana/k6-studio/issues/207

- Click a request in correlation rule preview will open inspector
- Sync `selectedRequest` to show changes when editing correlation rule

CleanShot 2024-09-25 at 15 43 27

### DIFF
--- a/src/components/WebLogView/WebLogView.tsx
+++ b/src/components/WebLogView/WebLogView.tsx
@@ -4,6 +4,8 @@ import { Group as GroupType, ProxyData } from '@/types'
 import { Row } from './Row'
 import { Group } from './Group'
 import { Table } from '@/components/Table'
+import { useMemo } from 'react'
+import { useDeepCompareEffect } from 'react-use'
 
 interface WebLogViewProps {
   requests: ProxyData[]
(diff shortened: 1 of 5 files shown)


src/components/WebLogView/WebLogView.tsx
```diff
@@ -21,6 +23,20 @@ export function WebLogView({
...
+  )
+
+  // Sync selectedRequest when requests change to show updates in correlation preview
+  useDeepCompareEffect(() => {
+    if (!selectedRequest) {
+      return
+    }
+
+    onSelectRequest(selectedRequest)
+  }, [selectedRequest, onSelectRequest])
```
e-fisher, line 38: When a request is clicked, we copy it to `selectedRequest` state. This is fine if request doesn't change, but when inspecting a request and editing a correlation rule the `requests` object will change and `selectedRequest` in parent component would become stale. This effect addresses it by syncing `selectedRequest` when `requests` change. It's not the ideal solution, but since it's likely the rule preview will change in the future, I didn't invest time in refactoring all consumers of `WebLogView`

cristianoventura (approved):
LGTM!

Llandy3d (approved):
🙌

Issue #206, opened 2024-09-23
Requests generated for 302 redirects chains


Llandy3d:
When the script gets generated, it also generates the caught requests from the redirect chain making multiple calls and in case of verification rule present, it would check the status code of each request.
This should be fixed because `k6` handles automatically redirects and returns the response from the last call showing a single request.

When generating the k6 script, omit the redirect chains and generate code to just perform the first request that would be checked with the status code of the last removed request.

PR #232, opened 2024-10-02
feat: merge redirect chains

cristianoventura:
## Description

Currently, when exporting a script, the Studio generates checks for all requests in a redirect chain. For example, the URL `http://www.k6.io` does 3 redirects.

```sh
$ curl -v -L http://www.k6.io 2>&1 | grep -i "^< location:"
< Location: https://www.k6.io/
< location: http://k6.io/index.html
< Location: https://k6.io/index.html
```

Right now, one check for each request above is generated when exporting the script, which is unexpected behaviour. 

This PR introduces logic responsible for identifying and resolving a redirect chain, ensuring that only the initial request and the check for the final response are present.

Example: Request to `http://www.k6.io` (first request in the chain) checks for the response in `https://k6.io/index.html` (final response in the chain).

## How to Test

- Create a recording and browse to a website that does at least one redirect, such as `http://www.k6.io` (note the HTTP protocol)
- Check the redirect chain in the request list
image
- Export a script and verify that only one check is present (checking for the HTTP status of the last request in the chain)

## Checklist

- [X] I have added tests for my changes.

## Related PR(s)/Issue(s)

 Resolves https://github.com/grafana/k6-studio/issues/206

### DIFF
--- a/src/codegen/codegen.utils.ts
+++ b/src/codegen/codegen.utils.ts
@@ -1,3 +1,6 @@
+import { ProxyData } from '@/types'
+import { getLocationHeader } from '@/utils/headers'
+
 // TODO: find a well-maintained library for this
 export function stringify(value: unknown): string {
   if (typeof value === 'string') {
@@ -18,4 +21,66 @@ export function stringify(value: unknown): string {
   }
 
   return `${value}`
+}
+
+function isRedirect(response: ProxyData['response']) {
+  if (!response) return false
+  return response.statusCode >= 300 && response.statusCode < 400
+}
+
+function findNextUrl(response: ProxyData['response']) {
+  if (!response || !isRedirect(response)) return undefined
+  return getLocationHeader(response.headers)
+}
+
+export function mergeRedirects(recording: ProxyData[]) {
+  const result = []
+  const processed = new Set()
+
+  for (const item of recording) {
+    const { response, id } = item
+
+    // Skip requests that have been processed
+    if (processed.has(id)) {
+      continue
+    }
+
+    // Requests without response don't need to be merged
+    if (!response) {
+      result.push(item)
+      processed.add(id)
+      continue
+    }
+
+    // Requests that are not redirects don't need to be merged
+    if (!isRedirect(response)) {
+      result.push(item)
+      processed.add(id)
+      continue
+    }
+
+    let finalResponse: ProxyData['response'] = response
+    let nextUrl = findNextUrl(response)
+
+    // Find the final response by following the redirect chain
+    while (nextUrl) {
+      const nextRequest = recording.find((req) => req.request.url === nextUrl)
+      if (nextRequest) {
+        processed.add(nextRequest.id)
+        finalResponse = nextRequest.response
+        nextUrl = findNextUrl(finalResponse)
+      } else {
+        break
+      }
+    }
+
+    result.push({
+      ...item,
+      response: finalResponse,
+    })
+
+    processed.add(id)
+  }
+
+  return result
 }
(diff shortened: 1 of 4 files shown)


Llandy3d (changes requested):
Trying with the recording below, both the `POST` and `GET` request are still being made while only the post one should be left, also the check is still checking for a `302` while it should be checking on the last request in the chain status code

I wonder if it's related to the use of the `Location` instead of the `Referer` 🤔 
I've attached the recording for easier testing

[Register User.har.zip](https://github.com/user-attachments/files/17242949/Register.User.har.zip)

src/codegen/codegen.utils.ts
```diff
@@ -19,3 +22,65 @@ export function stringify(value: unknown): string {
...
+    result.push({
+      ...item,
+      response: finalResponse,
+    })
+
+    processed.add(id)
+  }
+
+  return result
+}
```
e-fisher, line 86: Here's a simplified version, converted to functional programming approach. 
It resolves @Llandy3d's issue as well: the Location header in har file doesn't include host (e.g. "/") so next request's URL doesn't match. This solution skips URL checking. 

**One gotcha:**
This solution assumes all redirect chain's requests are in consecutive order. I wasn't to break this during my testing, but in theory could be possible in this scenario:

1. User clicks a link and a request is sent
2. A background fetch is triggered in JS to track user's click on a link
3. Response to 1. request is received, which is a redirect. 

In that case, the redirect chain would be interrupted by request in step no. 2.

To address this we would need to scan all requests and find the one matching the redirect, no longer assuming they are in order, and convert redirect header URLs to absolute when comparing. Though I'm not sure if that edge case needs to be covered at the current stage, WDYT @Llandy3d @cristianoventura ?

```suggestion
export function mergeRedirects(recording: ProxyData[]) {
  return recording.reduce((acc, request) => {
    const previousRequest = acc[acc.length - 1]

    // If the previous request is a redirect copy the response from the current
    // request to the previous request
    if (previousRequest && isRedirect(previousRequest?.response)) {
      return [
        ...acc.slice(0, -1),
        {
          ...previousRequest,
          response: request.response,
        },
      ]
    }

    return [...acc, request]
  }, [] as ProxyData[])
}

```

cristianoventura: @e-fisher I like how simple your solution is! 🤩 My initial implementation of using a `Set` was to keep track of what URLs were visited so the redirect chain could be resolved regardless of the order. We may run into issues if we assume they're always in order (which may not always be the case as you pointed out).

> and convert redirect header URLs to absolute when comparing
The current comparison uses the absolute URL when trying to find the next request, but this isn't always the case, which is why it's failing for the recording @Llandy3d  submitted.

```js
const nextRequest = recording.find((req) => req.request.url === nextUrl)
```

I'll implement a fix for this!

Llandy3d: I think it might make sense to try to dedicate some more time now to harden the checking so that we can also handle out of order redirects.
It's uncommon now for our tests to find that scenario but as this gets more adoption the probabilities of hitting it become higher 🤔

cristianoventura:
@Llandy3d The new changes now handle redirects with a relative path and address cases similar to the one you submitted!

> I wonder if it's related to the use of the Location instead of the Referer

The Referrer header lives on the client side and it's an optional header so it's unreliable to identify the redirect chain with it.
The `Location` response header on the other hand should always be present with a redirect response so it's more reliable in this case.

cristianoventura:
@e-fisher @Llandy3d  I also included tests to ensure this implementation handles redirects that are out of order.

Llandy3d (approved):
🙌

e-fisher (approved):
LGTM! 🚀

PR #253, opened 2024-10-15
feat: Error Boundary

going-confetti:
## Description

Add an error element for the top-level route error boundary

## How to Test

* Create a correlation rule
* Select `Begin-end` in the extractor type
* Type `\` in the `Begin` input field - you will see the error view
* Click `Go back` and verify that you've been redirected to the home view

## Checklist

image
image


src/ErrorElement.tsx
```diff
@@ -0,0 +1,72 @@
...
+  )
+}
+
+const handleOpenLogs = () => {
+  window.studio.app.openApplicationLog()
+}
+
+const handleRestart = () => {
+  window.location.replace('/')
+}
```
going-confetti, line 25: It might be a bit hacky: this essentially reloads the react app, which also erases the zustand stores making sure any erroneous input is reset.

Llandy3d: I would add this note in the code as a comment as well 👍

src/ErrorElement.tsx
```diff
@@ -0,0 +1,72 @@
...
+        />
+        <Text
+          size="2"
+          color="gray"
+          align="center"
+          css={css`
+            max-width: 560px;
+          `}
+        >
+          We apologize for the inconvenience. Please help us improve our service
```
Llandy3d, line 54: should it be `service` -> `application` ? 🤔

Llandy3d (dismissed):
🚀 🔌

src/ErrorElement.tsx
going-confetti: Yes, sounds better!

Issue #236, opened 2024-10-04
Persist user preferences


cristianoventura:
### Description

We need a way to persist user preferences in the application. This will enable us to provide a more personalized experience, allowing users to customize some settings. These preferences should be remembered next time the application launches.

For now, the settings we'd like to persist are:

- change [proxy mode](https://docs.mitmproxy.org/stable/concepts-modes/#upstream-proxy) from `regular` to `upstream`
- what port should the proxy run on

PR #254, opened 2024-10-15
feat: persist proxy preferences

cristianoventura:
## Description

This PR adds a "Settings" page where users can set some preferences. The following was implemented:

### Recorder

- browser path (let users input an alternative path to where Chrome executable is located)

### Proxy

- mode (regular or upstream)
- port
- automatically find port (let k6 studio find an available port if `port` is busy)
- upstream server and authentication

## How to Test

- Launch k6 studio and check that the settings JSON file was created in (`/Users/username/Library/Application Support/k6 Studio`)
- Change the browser path and attempt to launch the recorder
- Change the proxy port and check that a new process is running on that port (example: `lsof -i tcp:6001`)
- Check that all changes are saved in the JSON file

## Checklist

image

## Related PR(s)/Issue(s)

 Resolves https://github.com/grafana/k6-studio/issues/236


src/main.ts
```diff
@@ -56,8 +59,8 @@ const proxyEmitter = new eventEmitter()
 // Used mainly to avoid starting a new proxy when closing the active one on shutdown
 let appShuttingDown: boolean = false
 let currentProxyProcess: ProxyProcess | null
-let proxyReady = false
-export let proxyPort = 6000
+let proxyStatus: ProxyStatus = 'offline'
+export let appSettings: AppSettings
```
cristianoventura, line 63: - Proxy status will be tracked across its life cycle so we can _observe_ it in the UI.
- All the app configuration now lives in `appSettings`. This variable is dynamic and will change as settings are changed.

src/main.ts
```diff
@@ -152,11 +155,19 @@ const createWindow = async () => {
   }
 
   mainWindow.once('ready-to-show', () => configureWatcher(mainWindow))
+  proxyEmitter.on('status:change', (statusName: ProxyStatus) => {
+    proxyStatus = statusName
+    mainWindow.webContents.send('proxy:status:change', statusName)
+  })
```
cristianoventura, line 161: Since we already have an EventEmitter for the proxy I used it as a wrapper for changes related to the proxy status.

src/main.ts
```diff
@@ -152,11 +155,19 @@ const createWindow = async () => {
...
+  mainWindow.on('closed', () =>
+    proxyEmitter.removeAllListeners('status:change')
+  )
 
   return mainWindow
 }
 
 app.whenReady().then(async () => {
+  appSettings = await getSettings()
```
cristianoventura, line 170: Retrieving settings is the first event when the app launches. `getSettings` is an upsert so it'll create a configuration file if none is found.

src/main.ts
```diff
@@ -506,6 +517,52 @@ ipcMain.handle('app:open-log', () => {
   openLogFolder()
 })
 
+ipcMain.handle('settings:get', async () => {
+  console.info('settings:get event received')
+  return await getSettings()
+})
```
cristianoventura, line 523: Event handler to retrieve settings in the UI.

src/main.ts
```diff
@@ -506,6 +517,52 @@ ipcMain.handle('app:open-log', () => {
...
+
+ipcMain.handle('settings:save', async (event, data: AppSettings) => {
+  console.info('settings:save event received')
+
+  const browserWindow = browserWindowFromEvent(event)
+  try {
+    const diff = await saveSettings(data)
+    applySettings(diff, browserWindow)
```
cristianoventura, line 531: Event handler for saving settings from the UI. `saveSettings` returns the diff with only settings that have been modified. `applySettings` is responsible for overwriting and performing specific changes.

src/proxy.ts
```diff
@@ -50,10 +51,17 @@ export const launchProxy = (
     proxyScript,
     '--set',
     `confdir=${certificatesPath}`,
+    '--listen-port',
+    `${proxySettings.port}`,
```
cristianoventura, line 55: The proxy should always listen on the port from the settings file.

src/proxy.ts
```diff
@@ -50,10 +51,17 @@ export const launchProxy = (
...
     '--mode',
-    `regular@${port}`,
+    getProxyMode(proxySettings),
   ]
 
+  if (proxyRequiresAuth(proxySettings)) {
+    const { username, password } = proxySettings.upstream
+    proxyArgs.push('--upstream-auth', `${username}:${password}`)
+  }
```
cristianoventura, line 63: Appends authentication argument when upstream mode requires auth.

src/main.ts
```diff
@@ -506,6 +517,52 @@ ipcMain.handle('app:open-log', () => {
...
+ipcMain.handle('proxy:status:get', async () => {
+  console.info('proxy:status:get event received')
+  return proxyStatus
+})
+
+async function applySettings(
+  diff: Partial<AppSettings>,
+  browserWindow: BrowserWindow
+) {
+  if (diff.proxy) {
```
cristianoventura, line 555: This is where configuration changes when settings are saved.

src/views/Settings/ProxySettings.tsx
```diff
@@ -0,0 +1,123 @@
...
+        />
+      </FieldGroup>
+
+      {proxy && proxy.mode === 'upstream' && <UpstreamProxySettings />}
+
+      <Flex gap="2" mt="5">
+        <Text size="2">
+          Proxy status: <ProxyStatusIndicator status={proxyStatus} />
+        </Text>
+      </Flex>
```
cristianoventura, line 97: This proxy status can be moved to the footer/activity bar once we implement it.

src/settings.ts
```diff
@@ -0,0 +1,91 @@
+import { app } from 'electron'
+import { writeFile, open } from 'fs/promises'
+import path from 'node:path'
+import { AppSettings } from './schemas/appSettings'
+import { version } from '../package.json'
+
+const defaultSettings: AppSettings = {
+  appVersion: version,
```
cristianoventura, line 8: This file contains helpers for handling the settings JSON, including the default configuration that will be set when no file is found.

Currently, we don't support any type of _migration_ that allows the settings file to be backward compatible in case we introduce breaking changes. I'm including `appVersion` here so we can keep track of the schema between releases.

src/components/Layout/ActivityBar/SettingsButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Tooltip, IconButton } from '@radix-ui/themes'
+import { useNavigate } from 'react-router-dom'
+
+export function SettingsButton() {
+  const navigate = useNavigate()
+
+  return (
+    <Tooltip content="Settings" side="right">
+      <IconButton
+        are-label="Settings"
```
going-confetti, line 12:
```suggestion
        area-label="Settings"
```

src/schemas/appSettings.ts
```diff
@@ -0,0 +1,71 @@
...
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: 'Browser path is required',
+        path: ['browserPath'],
+      })
+    }
+  })
+
+export const AppSettingsSchema = z.object({
+  appVersion: z.string(),
```
going-confetti, line 65: Should this be `schemaVersion` instead? I presume breaking changes will be rather rare, so we can bump the schema version when needed + add migration logic

src/schemas/appSettings.ts
```diff
@@ -0,0 +1,71 @@
...
+      .min(1)
+      .max(65535, { message: 'Port number must be between 1 and 65535' }),
+    findPort: z.boolean(),
+    upstream: z.object({
+      url: z.string().url({ message: 'Invalid URL' }).or(z.literal('')),
+      requireAuth: z.boolean(),
+      username: z.string().optional(),
+      password: z.string().optional(),
+    }),
+  })
```
going-confetti, line 18: Not sure about storing the upstream URL in settings if the mode is set to `regular'. Would a discriminated union work better here?
TS example:
```ts
interface RegularModeSettings {
  mode: 'regular'
  port: number
  findPort: boolean
}

interface UpstreamModeSettings {
  mode: 'upstream'
  port: number
  findPort: boolean
  url: string
  requireAuth: boolean
  username: string
  password: string
}

type ProxySettings = RegularModeSettings | UpstreamModeSettings
```

There

src/schemas/appSettings.ts
```diff
@@ -0,0 +1,71 @@
...
+        code: z.ZodIssueCode.custom,
+        message: 'Password is required',
+        path: ['upstream.password'],
+      })
+    }
+  })
+
+export const RecorderSettingsSchema = z
+  .object({
+    detectBrowserPath: z.boolean(),
```
going-confetti, line 50: Similar to the other comment: if `detectBrowserPath` is `true`, there's no need to store `browserPath`

src/schemas/appSettings.ts
going-confetti: I don't think we plan on making app settings shareable, so it's probably not _too important_, but I think of a setting file as a representation of the current app settings, so having stuff that doesn't affect the app there seems odd to me.

Remembering previous choices may or may not be nice, depending on the setting IMO. For example, if I switch between two different configs often, then it makes a lot of sense, but if it's a setting I change once a year or so, then I'd find it confusing.

going-confetti: > For example, if I switch between two different configs often, then it makes a lot of sense

Although in this case, maybe the UX would need to be different: let the user create multiple proxy configurations and add a way to choose the active one in the recorder

cristianoventura: I ended up getting a `discriminatedUnion` to work. Now we're not persistent upstream configuration if proxy mode is regular. Same with the browser path when it's checked to detect it automatically!

going-confetti (dismissed):
LGTM

cristianoventura:
@going-confetti @Llandy3d I made changes to the PR that include:

- not storing upstream settings when `mode` is `regular`
- not storing browser path when `detectBrowserPath` is `true`
- moved the settings page to a dialog
- implemented a file picker for the browser path

https://github.com/user-attachments/assets/6cf5cd3e-d1a4-493d-9dc9-3cef3bc018fa

src/schemas/appSettings.ts
```diff
@@ -0,0 +1,88 @@
...
+  automaticallyFindPort: z.boolean(),
+})
+
+export const UpstreamProxySettingsSchema = RegularProxySettingsSchema.extend({
+  mode: z.literal('upstream'),
+  url: z.string().url({ message: 'Invalid URL' }).or(z.literal('')),
+  requiresAuth: z.boolean(),
+  username: z.string().optional(),
+  password: z.string().optional(),
+})
```
Llandy3d, line 19: small nit: I would have probably created a `BaseProxySettingsSchema` to be shared between all the proxy modes and then have each of them extend that since defining the port is a common setting 🤔  (but it does not matter much here and can be a future refactor 🙇 )

Llandy3d (approved):
🚀 🚀 🚀

going-confetti (approved):

Issue #241, opened 2024-10-07
WebSocket support


ppcano:
### Brief summary

WebSocket requests are converted into k6 http requests

### k6 Studio version

latest

### OS

macOS

### Steps to reproduce the problem

Recording a quickpizza flow https://quickpizza.grafana.com/

### Expected behaviour

Skip WebSocket requests until they are implemented, or alternatively, use the k6 WS API.

### Actual behaviour

Websocket request

![Image](https://github.com/user-attachments/assets/fa25b08b-2363-4d23-8bd5-32333de7ee35)

is converted into k6 http request:

```js
  url = http.url`https://quickpizza.grafana.com/ws`;
    resp = http.request("GET", url, null, params);

    check(resp, {
      "status matches 101": (r) => r.status === 101,
    });
```

it then fails:

![Image](https://github.com/user-attachments/assets/41014d91-2655-46f4-ab60-1fb2c1b1084e)

Llandy3d:
Thanks for reporting this issue!

PR #255, opened 2024-10-16
fix: remove websocket requests from recording

Llandy3d:
## Description

Websocket requests are now removed from the generated script file to avoid issues due to websockets not yet supported

## How to Test

Record a website with websockets, try to validate it and see no issue related to the protocol and the request missing for the `101` switching protocol.

## Checklist

## Related PR(s)/Issue(s)

Fix: #241

### DIFF
--- a/src/codegen/codegen.utils.ts
+++ b/src/codegen/codegen.utils.ts
@@ -1,3 +1,4 @@
+import { flow } from 'lodash-es'
 import { ProxyData } from '@/types'
 import { getLocationHeader, getUpgradeHeader } from '@/utils/headers'
 
@@ -116,8 +117,4 @@ export const removeWebsocketRequests = (recording: ProxyData[]) => {
   })
 }
 
-export const cleanupRecording = (recording: ProxyData[]) => {
-  let cleanedRecording = mergeRedirects(recording)
-  cleanedRecording = removeWebsocketRequests(cleanedRecording)
-  return cleanedRecording
-}
+export const cleanupRecording = flow([mergeRedirects, removeWebsocketRequests])
--- a/src/codegen/codegen.utils.ts
+++ b/src/codegen/codegen.utils.ts
@@ -116,7 +116,7 @@ export const removeWebsocketRequests = (recording: ProxyData[]) => {
   })
 }
 
-export const recordingCleanup = (recording: ProxyData[]) => {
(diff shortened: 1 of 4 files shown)


src/codegen/codegen.utils.ts
```diff
@@ -109,3 +109,15 @@ export function mergeRedirects(recording: ProxyData[]) {
...
   return result
 }
+
+export const removeWebsocketRequests = (recording: ProxyData[]) => {
+  return recording.filter((data) => {
+    return getUpgradeHeader(data.request.headers) !== 'websocket'
+  })
+}
+
+export const recordingCleanup = (recording: ProxyData[]) => {
```
going-confetti, line 119: Maybe `cleanupRecording`? Just so the naming is consistent across functions and the function name describes an action

src/codegen/codegen.utils.ts
```diff
@@ -109,3 +109,15 @@ export function mergeRedirects(recording: ProxyData[]) {
...
+  let cleanedRecording = mergeRedirects(recording)
```
going-confetti, line 120: It's not cleaned up at this point yet, so this doesn't seem like the best name.
Maybe we can simply do the following:
```ts
removeWebsocketRequests(mergeRedirects(recording))
```
it's not a long term solution though, we may want to use function composition or find another way to keep it readable as the number of steps grows 🤔

e-fisher (dismissed):
Agree with going-confetti's comments, otherwise LGTM!

src/codegen/codegen.utils.ts
Llandy3d: I wouldn't be in favor of chaining the calls because it will be less readable 🤔 

From my experience it was common to have the final name of the item you are going to return and overwrite it in multiple steps, it makes it readable what actions you are taking on it and the name reflects the value that get returned at the end of the function 🤔 
(technically even after the first step is cleaned because applied a "cleanup" step 😄 )

going-confetti: What about something like this:
```ts
import { flow } from 'lodash'

// ...

const cleanRecording = flow(
  mergeRedirects,
  removeWebsocketRequests
)(recording)

return cleanRecording
```
?

going-confetti (approved):
LGTM 🚀

Issue #247, opened 2024-10-15
Implement Parameterization Rule


Llandy3d:
Implement the Parameterization rule with the ability to take the source value from Test Variables.

We have a set of replacer functions ready to be used for replacing value in requests.
For this ticket the parameterization rule has two parts, the first one is a selector where you can choose which data you want to be used in the replacement, at this point this will be a value selected from the available Test Variables.

The second component is a selector for the replacer part, similar to the Correlation rule you can select either head/body/url/headers/cookie to be replaced.

PR #257, opened 2024-10-18
feat: Add parameterization rule UI

e-fisher:
## Description

Add parameterization rule UI. Allows to create and edit parameterization rules, view matched request's changes. 
Additionally, I've updated inline rule preview in rule list, badges had poor contrast in tooltips and text overflow didn't look great. I've improvised some icons with text (regexp = (.*), json = { }) since we didn't have those in currently used icon library, but they can be replaced in the future. 

This wasn't included in product DNA, but I've added support for custom values. I think it's useful for the cases when you need to override some value quickly, without creating a variable for it. 

## Current limitations
- Custom code currently is not supported, will be added in future PR
- The matched request preview is not great, there are no highlights what was changed, payload variable replacement looks to be wrapped in quotes - will be working on improved preview and open separate PR

## How to Test

- Try parameterizing different targets: body, url, headers
- Use both custom value and variable reference
- Verify variables cannot be removed or renamed when in use

## Checklist

CleanShot 2024-10-18 at 09 58 04

CleanShot 2024-10-17 at 14 41 58

CleanShot 2024-10-17 at 14 42 10

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/247


going-confetti (changes requested):
It looks like a parameterization rule is applied to the original request and not to the "actual" request modified by the previous rules. As an example, here I added 2 parameterization rules:
1. Replace a string between `Main` and `Page` in the URL with `hello` (one request to `en.wikipedia.org/wiki/MainPage` matched, as the result, the URL is transformed into `en.wikipedia.org/wiki/MainhelloPage`)
2. Replace a string between `hello` and `Page` in the URL with `world` (no requests matched, although in theory it should match the previous request and transform the URL into `en.wikipedia.org/wiki/MainhelloworldPage`)
image

src/views/Generator/TestOptions/VariablesEditor.tsx
```diff
@@ -113,3 +105,57 @@ export function VariablesEditor() {
...
+  remove,
+}: {
+  field: FieldArrayWithId<TestData, 'variables', 'id'>
+  index: number
+  register: UseFormRegister<TestData>
+  errors: FieldErrors<TestData>
+  remove: UseFieldArrayRemove
+}) {
+  const isVariableInUse = useGeneratorStore(
+    useShallow((state) =>
```
going-confetti, line 123: Nit: `useShallow` is redundant here as well (if I understand it correctly)

going-confetti (approved):
LGTM! Added 2 minor nits

going-confetti (approved):

PR #261, opened 2024-10-21
feat: URL prompt in Recorder

going-confetti:
## Description

image

## How to Test
1. Go to Recorder and enter a valid URL - verify that it's loaded when the browser starts
2. Go to Recorder and enter an invalid URL\some gibberish - verify that the browser starts and shows an error instead of the page

## Checklist

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -28,7 +28,10 @@ function getBrowserPath() {
   return recorder.browserPath as string
 }
 
-export const launchBrowser = async (browserWindow: BrowserWindow) => {
+export const launchBrowser = async (
+  browserWindow: BrowserWindow,
+  url?: string
+) => {
   const path = getBrowserPath()
   console.info(`browser path: ${path}`)
 
@@ -67,6 +70,7 @@ export const launchBrowser = async (browserWindow: BrowserWindow) => {
       `--proxy-server=http://localhost:${appSettings.proxy.port}`,
       `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
       disableChromeOptimizations,
+      url ?? '',
     ],
     onExit: sendBrowserClosedEvent,
   })
--- a/src/views/Recorder/EmptyState.tsx
+++ b/src/views/Recorder/EmptyState.tsx
@@ -1,4 +1,3 @@
-import { useEffect, useRef } from 'react'
 import { css } from '@emotion/react'
 import { zodResolver } from '@hookform/resolvers/zod'
 import { DiscIcon } from '@radix-ui/react-icons'
@@ -20,8 +19,6 @@ const RecorderEmptyStateSchema = z.object({
 type RecorderEmptyStateFields = z.infer<typeof RecorderEmptyStateSchema>
 
 export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
-  const inputRef = useRef<HTMLInputElement | null>(null)
-
   const {
     register,
     handleSubmit,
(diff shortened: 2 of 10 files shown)


src/browser.ts
```diff
@@ -67,6 +70,7 @@ export const launchBrowser = async (browserWindow: BrowserWindow) => {
       `--proxy-server=http://localhost:${appSettings.proxy.port}`,
       `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
       disableChromeOptimizations,
+      url ?? '',
```
going-confetti, line 73: This looks sketchy tbh, but I couldn't find a better alternative. Do you think it should somehow be sanitized before getting to this line?
The decision not to include any validation was deliberate in that it's much easier to correct erroneous input in the browser and continue the recording than to read a validation message and re-submit the form.

Llandy3d: I tried to input a command like `&& echo "HELLO"` but it stills opens it in chrome as url, personally I think it looks good 🙌

src/views/Recorder/Recorder.tsx
```diff
@@ -48,32 +48,32 @@ export function Recorder() {
...
-      })
-      log.error(error)
-    }
-  }, [resetProxyData, showToast])
+  const handleStartRecording = useCallback(
+    async (url?: string) => {
+      try {
+        resetProxyData()
+        setRecorderState('starting')
+        console.log('url', url)
```
Llandy3d, line 62: might be a leftover log line

src/views/Recorder/EmptyState.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <form
+        onSubmit={handleSubmit(onSubmit)}
+        css={css`
+          margin-top: var(--space-6);
+        `}
+      >
+        <FieldGroup
+          name="url"
+          label="Starting URL"
+          hint="Enter the URL of the website or service you want to test"
```
Llandy3d, line 71: Should we maybe mention `Optionally enter....` ? 🤔

going-confetti: @Llandy3d I think it's better to start with the action, so I'd maybe suggest appending `(optional)` to the hint or the label, but I'm not sure it's really needed. Let's wait for the final design and discuss it again

Llandy3d: Makes sense!
It was not clear on first impact that the field could be empty but looking forward to the final design 👀

src/browser.ts
cristianoventura: This seems to be sanitized behind the scenes by Chrome. I tested a few cases including `javascript:alert('')`.

We may want to sanitize this URL if we end up storing it in settings or something, but it looks good!

cristianoventura (commented):
LGTM! 🙌

cristianoventura (approved):

src/views/Recorder/EmptyState.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const onSubmit = ({ url }: RecorderEmptyStateFields) => {
+    onStart(url)
+  }
+
+  useEffect(() => {
+    window.requestAnimationFrame(() => {
+      inputRef.current?.focus()
+    })
+  }, [])
```
e-fisher, line 47: Looks like we can get way with the good old `autoFocus` prop on the `TextField` in this case, since input is not added to dom conditionally. I did some testing and seemed to work as expected, but let me know if I missed something

e-fisher (approved):
LGTM! 🚀 Left a NIT comment on `useEffect` to autofocus

e-fisher (approved):
🙌

PR #270, opened 2024-10-23
feat: enable usage report

cristianoventura:
## Description

This PR introduces changes to collect [usage report data when running a k6 script](https://grafana.com/docs/k6/latest/set-up/usage-collection/) as well as the following when running k6 Studio:

- os
- arch
- version
- timestamp

The user can opt-out of the usage report by disabling it in the Settings dialog.

## How to Test

- Check that `usageReport.enabled` property is properly saved
- Run a script from k6 Studio and check that the option `--no-usage-report` is passed to the k6 binary (you can do that by logging the `k6Args` array in the `script.ts` file.

## Checklist

image

### DIFF
--- a/src/components/Settings/UsageReportSettings.tsx
+++ b/src/components/Settings/UsageReportSettings.tsx
@@ -26,8 +26,8 @@ export const UsageReportSettings = () => {
                 checked={field.value}
                 onCheckedChange={field.onChange}
               />{' '}
-              I consent to the anonymous collection and use of data to improve
-              k6 Studio.{' '}
+              Please send my anonymous usage data to Grafana to aid in
+              development of k6 Studio.{' '}
               <Link href="" onClick={handleLinkClick}>
                 Learn more.
               </Link>
--- a/src/components/Settings/UsageReportSettings.tsx
+++ b/src/components/Settings/UsageReportSettings.tsx
@@ -22,12 +22,11 @@ export const UsageReportSettings = () => {
             <Text size="2" as="label">
               <Checkbox
                 {...register('usageReport.enabled')}
-                defaultChecked
                 checked={field.value}
                 onCheckedChange={field.onChange}
               />{' '}
-              Please send my anonymous usage data to Grafana to aid in
-              development of k6 Studio.{' '}
+              Send my anonymous usage data to Grafana to aid in development of
+              k6 Studio.{' '}
               <Link href="" onClick={handleLinkClick}>
                 Learn more.
               </Link>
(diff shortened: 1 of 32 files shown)


src/usageReport.ts
```diff
@@ -0,0 +1,55 @@
+import { v4 as uuidv4 } from 'uuid'
```
going-confetti, line 1: We use `self.crypto.randomUUID` in others places in k6 studio, I don't think a separate module is needed

src/components/Settings/UsageReportSettings.tsx
```diff
@@ -0,0 +1,40 @@
...
+          render={({ field }) => (
+            <Text size="2" as="label">
+              <Checkbox
+                {...register('usageReport.enabled')}
+                defaultChecked
+                checked={field.value}
+                onCheckedChange={field.onChange}
+              />{' '}
+              Please send my anonymous usage data to Grafana to aid in
+              development of k6 Studio.{' '}
```
going-confetti, line 30: I'd avoid using "please" (as mentioned in [Grafana's UX writing guide](https://grafana.com/docs/writers-toolkit/write/style-guide/ux-writing/#use-a-voice-thats-friendly-and-not-too-formal))

src/components/Settings/UsageReportSettings.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <SettingsSection title="Usage Report">
+      <Flex gap="2">
+        <Controller
+          control={control}
+          name="usageReport.enabled"
+          render={({ field }) => (
+            <Text size="2" as="label">
+              <Checkbox
+                {...register('usageReport.enabled')}
+                defaultChecked
```
going-confetti, line 25: IMO this is redundant: the field should reflect the current settings: if they are saved in the file, the user will see the value from the file; if not, the user will see the default value

src/usageReport.ts
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    await fetch(URL, {
+      method: 'post',
+      body: JSON.stringify(data),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+  } catch (e) {
+    console.log('Error sending report', e)
```
going-confetti, line 53: Since we don't send the report when running the app in dev env, this line isn't really needed. I think we should either log it in the app logs, or fail silently instead

src/usageReport.ts
Llandy3d: That's on me!
I didn't notice :)

PR #272, opened 2024-10-24
fix: prevent infinite error loop on proxy restart

cristianoventura:
## Description

This PR introduces logic that prevents the proxy from being stuck on an infinite restarting loop. This is known to happen in two occasions:

- When the option to automatically find a port is set to `false` and the currently selected port is busy
- On Windows when there are instances of the proxy still running before new settings can be applied

proxy-loop-error-windows

## How to Test

- Launch an instance of k6 Studio and set the proxy port
- Launch a separate instance of k6 Studio, set the option to automatically find a port to `false` and use the same port as the previous instance.

## Checklist

Screenshot 2024-10-24 at 11 38 50 AM

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/273

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -577,7 +577,7 @@ async function applySettings(
   browserWindow: BrowserWindow
 ) {
   if (modifiedSettings.proxy) {
-    stopProxyProcess()
+    await stopProxyProcess()
     appSettings.proxy = modifiedSettings.proxy
     proxyEmitter.emit('status:change', 'restarting')
     currentProxyProcess = await launchProxyAndAttachEmitter(browserWindow)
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -615,6 +617,22 @@ const launchProxyAndAttachEmitter = async (browserWindow: BrowserWindow) => {
         // don't restart the proxy if the app is shutting down or if it's already restarting
         return
       }
+
+      if (proxyRetryCount === PROXY_RETRY_LIMIT && !automaticallyFindPort) {
+        proxyRetryCount = 0
+        proxyEmitter.emit('status:change', 'offline')
```
cristianoventura, line 623: Setting a max retry attempt if the user opts not to find an available port automatically.

src/main.ts
```diff
@@ -685,11 +703,15 @@ function getFilePathFromName(name: string) {
...
+const stopProxyProcess = async () => {
   if (currentProxyProcess) {
-    // NOTE: this might not kill the second spawned process on windows
     currentProxyProcess.kill()
     currentProxyProcess = null
+
+    // kill remaining proxies if any, this might happen on windows
+    if (getPlatform() === 'win') {
+      await cleanUpProxies()
+    }
```
cristianoventura, line 714: Ensures that other instances of the proxy are killed. This is the same logic used during launch

going-confetti (approved):
LGTM! There should probably we a more prominent error message for when the max number of retries is reached, but it's out of scope for this PR, I think

PR #282, opened 2024-10-28
feat: Add parameterization custom code logic

e-fisher:
## Description

Add logic to support parameterization rule custom code value. The snippet will be wrapped in a function `getCorrelationValueN`. This is different to custom code rule where the snippet is dropped into script without wrapping it, but I went with this approach as it more versatile and helps avoid variable name collision. Also, in the future this could be extended to pass request details to the custom code function to allow more customization, e.g.,: `getCorrelationValue0(params, headers, body)`.

The UI for adding custom code parameterization will be added in a separate PR. 

## How to Test

- Manually add parameterization rule with custom code
- Verify snippet is added to script
- Verify snippet is referenced in request
- Verify snippet function is added only once when referenced in multiple requests

## Checklist

CleanShot 2024-10-28 at 11 01 43

### DIFF
--- a/src/rules/rules.ts
+++ b/src/rules/rules.ts
@@ -7,10 +7,21 @@ import { createParameterizationRuleInstance } from './parameterization'
 import { ProxyData, RequestSnippetSchema } from '@/types'
 import { generateSequentialInt } from './utils'
 
+function createSequentialIdPool() {
+  const currentId: Record<TestRule['type'], Generator<number>> = {
+    correlation: generateSequentialInt(),
+    parameterization: generateSequentialInt(),
+    verification: generateSequentialInt(),
+    customCode: generateSequentialInt(),
+  }
+
+  return (type: TestRule['type']) => currentId[type]
+}
+
 export function applyRules(recording: ProxyData[], rules: TestRule[]) {
-  const idGenerator = generateSequentialInt()
+  const idGenerator = createSequentialIdPool()
   const ruleInstances = rules.map((rule) =>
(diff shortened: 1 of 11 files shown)


src/rules/shared.ts
```diff
@@ -170,7 +170,7 @@ export const replaceBeginEndBody = (
     selector.begin,
     selector.end
   )
-  if (!valueToReplace) return request
+  if (!valueToReplace) return
```
e-fisher, line 173: Previously, we had no way of knowing if the request matched when the value replaced is the same previous value. While it does not change script, it still should be displayed as matched request in rule preview. 

Returning `undefined` when there's no match allows consumers to know when rule matched.

src/rules/rules.ts
```diff
@@ -7,10 +7,21 @@ import { createParameterizationRuleInstance } from './parameterization'
...
 import { generateSequentialInt } from './utils'
 
+function createSequentialIdPool() {
+  const currentId: Record<TestRule['type'], Generator<number>> = {
+    correlation: generateSequentialInt(),
+    parameterization: generateSequentialInt(),
+    verification: generateSequentialInt(),
+    customCode: generateSequentialInt(),
+  }
+
```
e-fisher, line 17: Independent ID sequence for every rule type

cristianoventura (dismissed):
I manually changed the generator to use `customCode` and the script generates as expected! 🚀

going-confetti (approved):
LGTM 🤩

PR #292, opened 2024-10-30
fix: Show allowlist dialog as modal when action is required

e-fisher:
## Description

We've recently updated allowlist dialog to be displayed as an inline popover to match test options' dialog style. Although this approach is effective, we would like to direct the user's attention towards the allowlist when the user's input is required. This includes when creating a new generator, opening a generator without a set allowlist, or switching between recordings that contain different hosts. With this change, allowlist could be displayed both as popover or modal depending on this condition:

If allowlist is triggered by indirect action (open generator, switch recording) → show as modal.
If allowlist button is clicked inside generator → show as popover.

Could use some feedback whether we should disable "Continue" button when no hosts are selected. Currently, it's not blocking you from continuing without any hosts because neither does the allowlist popover. 

## How to Test

Verify allowlist is displayed as a popover when clicking the allowlist button, and as modal in other cases. 

## Checklist

CleanShot 2024-10-30 at 16 17 31

CleanShot 2024-10-30 at 16 17 43

### DIFF
--- a/src/views/Generator/Allowlist/Allowlist.tsx
+++ b/src/views/Generator/Allowlist/Allowlist.tsx
@@ -1,15 +1,18 @@
-import { Button } from '@radix-ui/themes'
+import { Button, Dialog, Flex } from '@radix-ui/themes'
 import { useGeneratorStore } from '@/store/generator'
 import { AllowlistDialog } from './AllowlistDialog'
 import { extractUniqueHosts } from '@/store/generator/slices/recording.utils'
 import { GlobeIcon } from '@radix-ui/react-icons'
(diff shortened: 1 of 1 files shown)


Llandy3d (approved):
🚀 🚀 🚀 

> Could use some feedback whether we should disable "Continue" button when no hosts are selected. Currently, it's not blocking you from continuing without any hosts because neither does the allowlist popover.

Personally I think it's fine, the continue is there to go to the next step. It's true that it doesn't make too much sense to proceed with no host selected but the user is not locked out, they can still select the host from the popover. So overall for this PR I think it's fine 🙌

PR #294, opened 2024-10-31
fix: Import external HAR

e-fisher:
## Description

Currently, when trying to import externally recorded HAR, you'll get an error. This is caused by trying to decode a non-base64-encoded response content. Using `safeAtob` will allow us to support both encoded and non-encoded HARs. 

## How to Test

1. Using your browser (I've used Chrome) visit a website with devtools opened, save the HAR from network tab
2. Import HAR into k6 studio

## Checklist


Llandy3d (approved):
🚀

PR #301, opened 2024-11-05
feat: organize settings dialog with tabs

cristianoventura:
## Description

This PR changes the Settings dialog by separating sections into tabs. This provides a cleaner user interface as we add more functionalities to the Studio and enable users with custom settings.

## How to Test

- Make the state of the settings dialog invalid
- Attempt to submit
- Check that tabs with errors are automatically selected and, inputs with errors automatically get focus

## Checklist

https://github.com/user-attachments/assets/a01c999c-a401-4a8e-a255-2ec6c0097c92

### DIFF
--- a/src/components/Settings/SettingsDialog.tsx
+++ b/src/components/Settings/SettingsDialog.tsx
@@ -122,7 +122,7 @@ export const SettingsDialog = ({ open, onOpenChange }: SettingsDialogProps) => {
               </Tabs.List>
             </Box>
 
-            <Box minHeight="0" maxHeight="480px" flexGrow="1">
+            <Box minHeight="0" maxHeight="460px" mb="4" flexGrow="1">
               <ScrollArea>
                 {tabs.map((tab) => (
                   <Tabs.Content key={tab.value} value={tab.value}>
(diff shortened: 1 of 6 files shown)


src/components/Settings/SettingsDialog.tsx
```diff
@@ -1,23 +1,36 @@
...
+import { ExclamationTriangleIcon } from '@radix-ui/react-icons'
+import { ButtonWithTooltip } from '../ButtonWithTooltip'
+import { findIndex, sortBy } from 'lodash-es'
 
 type SettingsDialogProps = {
   open: boolean
   onOpenChange: (open: boolean) => void
 }
 
+const tabs = [
```
cristianoventura, line 20: I extracted the content of the tabs to an array, so we can loop through and select which one to focus on after an error happens, following the order they are rendered in the UI. This ensures that the user is taken to the very first error at the top of the form. 

This loop happens in the `onInvalid` method on L72.

going-confetti: > so we can loop through and select which one to focus on after an error happens

Looks like saving on blur/change can help us avoid the problem altogether: we won't need to show the warning icon and focus on a broken tab, and the user won't need to remember to save when the changes they made are on a different tab

Llandy3d (dismissed):
Looks good!

I wonder if it makes sense for the `Recorder` section to show greyed out the inputs for selecting a specific browser... it feels like the information that you can actually do that it's hidden away 🤔

going-confetti (approved):
LGTM!
@cristianoventura and I talked about auto-saving changes and it looks like it'll be a separate improvement later on, because of how settings are designed (for example, changing the proxy mode changes how the form should be validated)

PR #302, opened 2024-11-07
feat: capture additional logs for proxy

cristianoventura:
## Description

When the proxy fails to start, the only line outputted to the log file is the following:

```.log
[2024-11-07 10:53:51.755] [error] Error logged during startup, exiting...
```

This happens because mitmproxy outputs some errors to stdout and, we are not capturing them.

With the changes in this PR, the log file will look like this when the proxy port is already in use, but additional information will be captured regardless of the error:

```log
[2024-11-07 10:48:52.998] [error] Error logged during startup, exiting...
[2024-11-07 10:48:52.999] [error] [10:48:52.998] [Errno 48] HTTP(S) proxy (upstream mode) failed to listen on *:6001 with [Errno 48] error while attempting to bind on address ('0.0.0.0', 6001): address already in use
[2024-11-07 10:48:53.000] [error] Try specifying a different port by using `--mode upstream:localhost:200@6003`.
[2024-11-07 10:48:53.000] [error] Proxy failed to start on port 6001, restarting...
```

## How to Test

- launch a separate instance of k6 Studio (such as a production build using a valid port or, start mitmproxy on your terminal `mitmproxy --mode regular@6001`)
- launch k6 studio using this branch and change the port to the same as the other instance. Make sure the "Automatically find port" option is disabled
- the proxy should fail to launch due to the port being busy
- check the log file

## Checklist

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -671,8 +671,10 @@ const launchProxyAndAttachEmitter = async (browserWindow: BrowserWindow) => {
       proxyEmitter.emit('status:change', 'restarting')
       currentProxyProcess = await launchProxyAndAttachEmitter(browserWindow)
 
+      const errorMessage = `Proxy failed to start on port ${proxyPort}, restarting...`
+      log.error(errorMessage)
       sendToast(browserWindow.webContents, {
-        title: 'Proxy failed to start, restarting...',
+        title: errorMessage,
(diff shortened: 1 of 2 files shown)


Llandy3d (approved):
LGTM!

PR #303, opened 2024-11-08
refactor: Avoid param reassignment in group_snippet

going-confetti:
## How to Test
⚠️ Please check that Validator works with different test scripts

## Checklist

### DIFF
--- a/resources/group_snippet.js
+++ b/resources/group_snippet.js
@@ -1,24 +1,18 @@
 import execution from 'k6/execution'
 
 // saving the original values
-const request = http.request
-const asyncRequest = http.asyncRequest
+const httpRequest = http.request
+const httpAsyncRequest = http.asyncRequest
 
 class Client {
   // request instruments the http module's request function with the group header
   request(method, url, ...args) {
-    const group = { 'X-k6-group': trimAndRemovePrefix(execution.vu.tags.group) }
-    args = instrumentArguments(group, ...args)
-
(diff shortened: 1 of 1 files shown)


resources/group_snippet.js
```diff
@@ -45,44 +39,23 @@ class Client {
...
+function instrumentArguments(args) {
+  const [body = null, params = {}] = args
+  const groupHeaders = {
+    'X-k6-group': trimAndRemovePrefix(execution.vu.tags.group),
   }
 
-  return args
+  const updatedParams = Object.assign({}, params, {
+    headers: Object.assign({}, params.headers || {}, groupHeaders),
+  })
```
going-confetti, line 53: TIL that our version of k6 doesn't support the spread operator. It was added to a newer version (https://github.com/grafana/k6/pull/3456)

resources/group_snippet.js
going-confetti, line 1: I'm not quite happy with this snippet "polluting" the global namespace, essentially making it impossible to create user-defined variables with names like `Client`, `trimAndRemovePrefix`, etc - it's only possible in custom code snippets (and I'm still not sure whether those should have their own scope - let's have a separate discussion on that)

e-fisher (approved):
LGTM 👌

resources/group_snippet.js
Llandy3d: good point!
We should definitely not reserve those names 🤔

PR #346, opened 2024-11-12
feat: Documentation link

going-confetti:
## Description

* New dropdown menu in the bottom left corner that groups both the documentation link and the linkt to github issues
* Documentation option in the Help menu

## Checklist

image
image

### DIFF
--- a/src/components/Layout/ActivityBar/ActivityBar.tsx
+++ b/src/components/Layout/ActivityBar/ActivityBar.tsx
@@ -10,6 +10,7 @@ import { NavIconButton } from './NavIconButton'
 import { ApplicationLogButton } from './ApplicationLogButton'
 import { SettingsButton } from './SettingsButton'
 import { ProxyStatusIndicator } from './ProxyStatusIndicator'
+import { HelpButton } from './HelpButton'
 
 export function ActivityBar() {
   return (
@@ -46,6 +47,7 @@ export function ActivityBar() {
           <ProxyStatusIndicator />
           <SettingsButton />
           <ApplicationLogButton />
+          <HelpButton />
           <VersionLabel />
         </Flex>
       </Box>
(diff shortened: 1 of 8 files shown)


cristianoventura (dismissed):
👌

Llandy3d (commented):
Noticed a small bug where after selecting the icon and clicking away it stays highlighted
image

I wonder if it makes sense to have them grouped here or they should be distinct 🤔  (or just find docs under the help section)

going-confetti:
> Noticed a small bug where after selecting the icon and clicking away it stays highlighted image

This is not a bug, it's by design, so that the focus isn't lost. It works the same in other places across the app.

> I wonder if it makes sense to have them grouped here or they should be distinct 🤔 (or just find docs under the help section)

What would be the downside of grouping them together? We discussed this with Maja and it seemed like a good solution, but maybe we're missing something

Llandy3d:
> What would be the downside of grouping them together? We discussed this with Maja and it seemed like a good solution, but maybe we're missing something

Might just be personal opinion (and not on the strong side), the first impact was that we were now hiding the `report issue` button behind a generic info button and another detail is that there is no tooltip, so you don't know what happens until you press the button, but this button behaves differently from the others, where the others do an action and this one doesn't. (and the user cannot possibly know that 🤔 )

> This is not a bug, it's by design, so that the focus isn't lost. It works the same in other places across the app.
I feel like here seems odd to have focus on a button 🤔 
But regardless I think we have a bug because if you press on the `theme toggle` and then try to press again on this new info button, it's not going to be highlighted anymore! 😄

going-confetti:
> But regardless I think we have a bug because if you press on the theme toggle and then try to press again on this new info button, it's not going to be highlighted anymore! 😄

Still not a bug. This is because of the difference between `:focus` and `:focus-visible` in CSS. The idea is that you may want to only have focus outlines when interacting with controls with keyboard or setting focus programmatically.
For example, when you edit a comment on GitHub, you can tab into the `B` button to make the text bold - the outline will be displayed, but will disappear if you click on any other button with your mouse cursor

cristianoventura (dismissed):

Llandy3d (dismissed):
LGTM!

Llandy3d (approved):
🚀

PR #347, opened 2024-11-12
feat: Persistent theme settings

going-confetti:
## How to Test
* Try changing the default setting
* Close the app
* Open it again - the setting should persist

## Checklist

image


src/components/ThemeSwitcher.tsx
going-confetti, line 1: Goodbye 😭

going-confetti:
> Oh man, I loved that switch in activity bar 😞 It removes the friction to test both themes when developing, can we keep it in activity bar additionally to settings? We could also wrap in dev mode condition if there's a reason we don't want to have it there in prod builds.

Let's make it dev-only for now 👍

src/components/ThemeSwitcher.tsx
```diff
@@ -4,6 +4,12 @@ import { IconButton, Tooltip } from '@radix-ui/themes'
 
 export function ThemeSwitcher() {
   const theme = useTheme()
+  // @ts-expect-error we have commonjs set as module option
+  const isDev = import.meta.env.DEV
```
cristianoventura, line 8: Would `process.env.NODE_ENV === 'development'` work here?

cristianoventura (dismissed):
LGTM! 🙌

src/components/ThemeSwitcher.tsx
going-confetti: Thanks for the suggestion. I was under the impression it wasn't available in the render process 👀. Replaced with `process.env.NODE_ENV === 'development'` 👍

cristianoventura (approved):

Issue #237, opened 2024-10-04
Toggle word-wrap in Monaco editor


cristianoventura:
### Description

Some instances of the Monaco editor contain long lines that create horizontal scrolling, particularly when opening the Script Preview or inspecting HTTP responses. To enhance usability, we'd like to add a word-wrap toggle that allows users to easily manage line length and improve readability.

PR #355, opened 2024-11-18
feat: word-wrap editor

cristianoventura:
## Description

This PR enables word-wrap for the Monaco editor in Script Preview and the Content tab of a response. The state for both are persisted individually in the settings file.

## How to Test

- open Script Preview and toggle the word-wrap switch
- open the response content of a JSON request and toggle word-wrap
- re-open the app and check that the previous state was persisted

## Checklist

- [] I have added tests for my changes.

https://github.com/user-attachments/assets/3b93ffa2-ed78-458d-9ae9-108ab4c01f58

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/237


src/hooks/useEditorWordWrapSetting.ts
```diff
@@ -0,0 +1,33 @@
+import { EditorSettings } from '@/types/settings'
+import { useState, useEffect } from 'react'
+import { useDebounce } from 'react-use'
+
+export function useEditorWordWrapSetting(key: keyof EditorSettings) {
```
cristianoventura, line 5: This hook can be used to handle each setting independently based on the `key` (`wordWrapScriptPreview` or `wordWrapResponseContent`)

e-fisher (commented):
I feel like word wrap toggle should be a part of CodeEditor component, perhaps a Toolbar component inside, with option to control it through props. 

Also, I probably wouldn't go as far as having this option saved per every component, just drop it in local storage to remember my preferred view and apply in all places we show editor. It's only one click away to toggle, anyway. IMO, added complexity doesn't outweigh benefit added, but would love to hear what others think.

cristianoventura (commented):
I've followed @e-fisher's suggestion to keep the state in local storage. This indeed minimizes complexity. I also implemented a `EditorToolbar` component to hold the word wrap switch.

image

cc @going-confetti

Llandy3d (approved):
Looks good! 🙌 

I wonder if it makes sense to have it on a smaller size on the right side of the bar UI wise because I assume it is a "set and forget" element 🤔

e-fisher (commented):
Agree with Llandy on moving it to the right. Maybe not just make it smaller, but use an icon with a tooltip instead? 

Found a weird bug when toggle doesn't when viewing `raw` request response, here's a video:

https://github.com/user-attachments/assets/4ea1207a-df73-46ef-b1ea-81d4043f9535

Otherwise looks great!

cristianoventura:
I've moved the word wrap switch to an IconButton aligned at the end of the toolbar.

image

@e-fisher I noticed that this only happens for JSON >= than 10000 characters. It could be related to internal performance so I'm still trying to figure it out.

src/components/Monaco/ReactMonacoEditor.tsx
```diff
@@ -36,14 +35,32 @@ export function ReactMonacoEditor({
...
 }: ReactMonacoEditorProps) {
   const theme = useTheme()
+  const [editor, setEditor] = useState<monaco.editor.IStandaloneCodeEditor>()
   const [toolbarState, setToolbarState] = useState<ToolbarState>({
     wordWrap: 'off',
   })
 
+  // Monaco automatically applies word wrap if the content length of a line is >= 10000 characters
+  // In this case, we disable the word wrap button so the internal state is respected
+  const shouldEnableWordWrapButton = () => {
```
cristianoventura, line 45: @e-fisher This was an interesting find. If any line has a content length >= 10000 characters, Monaco overrides the `wordWrap` option. This is likely related to [stopRenderingLineAfter](https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IEditorConstructionOptions.html#stopRenderingLineAfter) but changing this option doesn't seem to have any effect on this case specifically other than hiding the rest of the line content.

This was the solution I came up with by disabling the word wrap button and respecting Monaco's internal state. Please let me know what you think.

Llandy3d (dismissed):

e-fisher (dismissed):
Great find, and sorry for taking you down the rabbit hole! 😅 

I would wrap the line calculation in `useMemo` to avoid running on every re-render, but otherwise looks good!

Llandy3d (approved):

PR #357, opened 2024-11-20
feat: bundle latest k6 on build

### LINKED ISSUE #327: bundle latest k6 version on build
We are currently shipping a pinned version of k6, we should improve the build process to bundle the latest version of k6 on release.

Llandy3d:
## Description

The binaries for `k6` have been removed and now the specified version of `k6` is retrieved from the k6 github repository during the release ci process. The version is specified by a repository variable called `K6_VERSION`.

Due to this, the binary is also missing in `dev` so the `npm install` has been augmented with a script to retrieve the binary if it's missing.

_note: The version for `dev` is specified inside of the `install-k6.js` file at this time, so the version is specified in 2 places._

## How to Test

- Clone the repository and `npm install` making sure you have `k6` installed after it.
- Trigger a manual release from this branch and test that it includes k6.

## Checklist

## Related PR(s)/Issue(s)

Closes #327

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,43 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # get k6 binaries macos
+      - name: get latest k6 binary macos
+        if: startsWith(matrix.platform, 'macos-')
+        run: |
+          K6_PATH_AMD=k6-${{ vars.K6_VERSION }}-macos-amd64
+K6_PATH_ARM=k6-${{ vars.K6_VERSION }}-macos-arm64
+
+          # download binaries
+          curl -LO https://github.com/grafana/k6/releases/download/${{ vars.K6_VERSION }}/$K6_PATH_AMD.zip
+          curl -LO https://github.com/grafana/k6/releases/download/${{ vars.K6_VERSION }}/$K6_PATH_ARM.zip
+
+          # unzip & smoke test
+          unzip $K6_PATH_AMD.zip
+          unzip $K6_PATH_ARM.zip
+          $K6_PATH_AMD/k6 version
+          $K6_PATH_ARM/k6 version
+
+          # move to resource folder
+          mv $K6_PATH_AMD/k6 resources/mac/x86_64
+          mv $K6_PATH_ARM/k6 resources/mac/arm64
+
+      # get k6 binaries windows
+      - name: get latest k6 binary windows
+        if: startsWith(matrix.platform, 'windows-')
+        run: |
+          $K6_PATH_AMD="k6-${{ vars.K6_VERSION }}-windows-amd64"
+
+          # download binaries
+          Invoke-WebRequest -Uri "https://github.com/grafana/k6/releases/download/${{ vars.K6_VERSION }}/$K6_PATH_AMD.zip" -OutFile "k6-windows-amd64.zip"
+
+          # unzip & smoke test
+          Expand-Archive -Path "k6-windows-amd64.zip" -DestinationPath "."
+          & $K6_PATH_AMD\k6.exe version
+
+          # move to resource folder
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -18,6 +18,43 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # get k6 binaries macos
+      - name: get latest k6 binary macos
```
going-confetti, line 22: Can we run the `install-k6` script here so that the code isn't duplicated?

Llandy3d: Great suggestion!
I built them in the opposite order so I didn't think about it, but if it works in the action and both locally for windows as well we might be able to just use the single script 👀

install-k6.js
```diff
@@ -0,0 +1,91 @@
...
+      getMacOSK6Binary()
+      console.log('k6 binary download completed')
+    }
+    break
+  case 'win32':
+    if (!existsK6('win', 'x86_64')) {
+      console.log('k6 binary not found')
+      console.log('downloading k6... this might take some time...')
+      getWindowsK6Binary()
+      console.log('k6 binary download completed')
```
cristianoventura, line 86: I noticed a race condition between downloading the binaries and showing the success message while it was still in progress. We can solve this by making `executeCommand` resolve or reject a Promise or, we can move the success message inside the `exec` callback.

image

install-k6.js
```diff
@@ -0,0 +1,91 @@
...
+# download binaries
+Invoke-WebRequest -Uri "https://github.com/grafana/k6/releases/download/${K6_VERSION}/${K6_PATH_WIN_AMD}.zip" -OutFile "k6-windows-amd64.zip"
+
+# unzip & smoke test
+Expand-Archive -Path "k6-windows-amd64.zip" -DestinationPath "."
+${K6_PATH_WIN_AMD}\\k6.exe version
+
+# move to resource folder
+Move-Item -Path "${K6_PATH_WIN_AMD}\\k6.exe" -Destination resources\\win\\x86_64
+`
```
cristianoventura, line 50: I noticed the temp files were still there after the download was complete (they were also showing up in `git status`). We can add a clean up function here similar to what we have for macOS


```suggestion
# clean up
del k6-windows-amd64.zip
Remove-Item -Recurse -Force ${K6_PATH_WIN_AMD}
`
```

install-k6.js
```diff
@@ -0,0 +1,91 @@
+const { exec } = require('child_process')
+const { existsSync } = require('fs')
+
+const K6_VERSION = 'v0.55.0'
+const K6_PATH_MAC_AMD = `k6-${K6_VERSION}-macos-amd64`
+const K6_PATH_MAC_ARM = `k6-${K6_VERSION}-macos-arm64`
+const K6_PATH_WIN_AMD = `k6-${K6_VERSION}-windows-amd64`
+
+const existsK6 = (os, arch) => {
+  return existsSync(`resources/${os}/${arch}/k6`)
```
cristianoventura, line 10: This isn't working on Windows because of the missing extension (`.exe`) 😞 We may have to do the os check inside this function instead.

.github/workflows/release.yml
cristianoventura: I agree with running `node install-k6.js` instead to remove duplication.

cristianoventura (commented):
We need some adjustments on Windows but it's looking good so far!

install-k6.js
Llandy3d: Noticed this one as well!
I'm now making use of `execSync` instead of `exec` and that fixes the message issue as well 🙌

install-k6.js
cristianoventura: We can remove this function now that we are using `existsSync` directly.

cristianoventura (commented):
It looks good on both Mac and Windows 🚀  Just one minor thing left.

cristianoventura (approved):

Issue #370, opened 2024-12-02
Generator schema migration


cristianoventura:
Using the pattern adopted in https://github.com/grafana/k6-studio/issues/265, we need to implement a migration strategy for the Generator schema.

PR #371, opened 2024-12-02
feat: settings migration

cristianoventura:
## Description

This PR adds migrations to the settings schema, ensuring that changes to its structure remain backward-compatible. This allows the settings to work properly even when the schema is updated.

The same pattern will be applied to the Generator schema on a separate PR.

## How to Test

This PR contains one extra commit that is intended to be used for testing purposes and will be reverted before merging.

Before testing this branch, make sure your local settings file (`k6-studio-dev.json`) contains `usageReport.enabled` `false` and `appearance.theme` as `light` so you can validate the changes.

```json
"version": "1.0",
...
"usageReport": { "enabled": false },
"appearance": { "theme": "light" }
```

- Launch k6 studio from this branch
- Check that the theme and usage report are still the same
- Make changes to any other settings in the Settings dialog and check that the previous changes were migrated

Once the settings are saved, the file structure should look like this instead:

```json
"version": "1.0",
...
"general": {
   "usageReport": { "enabled": false },
   "appearance": { "theme": "light" }
}
```

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have added tests for my changes.
- [x] I have run linter locally (`npm run lint`) and all checks pass.
- [x] I have run tests locally (`npm test`) and all tests pass.
- [x] I have commented on my code, particularly in hard-to-understand areas.

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/265

PR #373, opened 2024-12-04
feat: generator schema migration

cristianoventura:
## Description

This PR adds migrations to the generator schema, ensuring that changes to its structure remain backward-compatible. This allows the generator to work properly even when the schema is updated.

The pattern adopted in this PR is based on the settings migration https://github.com/grafana/k6-studio/pull/371.

## How to Test

- make changes to the latest version of the schema `/schemas/generator/v1/index.ts` such as renaming a property.
- change the `migrate` function in `/schemas/generator/v0/index.ts` to return the new schema.
- make appropriate changes to the remaining implementation so the new schema is used. This may require changing the slices and several other parts of the code.
- ensure the Generator file has `version: '0'` and all old properties before testing.
- open the old generator file and check that options were migrated.

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/370

### DIFF
--- a/src/schemas/generator/README.md
+++ b/src/schemas/generator/README.md
@@ -14,10 +14,10 @@ For when a new migration is needed:
 4. Update `/schemas/generators/index.ts` to use the new version:
 
 ```ts
-function migrate(settings: z.infer<typeof AnySettingSchema>) {
+function migrate(generator: z.infer<typeof AnyGeneratorSchema>) {
 
 case '1.0':
-  return migrate(v1.migrate(settings))
+  return migrate(v1.migrate(generator))
 
 }
 ```
--- a/src/schemas/generator/README.md
+++ b/src/schemas/generator/README.md
@@ -1,0 +1,25 @@
+# Generator migration
+
+Migrations are needed when changing the structure of existing generators file such as renaming or deleting keys.
+
+If you're simply adding new options to the generator file, a migration may not be needed. In this case, extend the latest schema available.
+
+## Creating a new migration
+
+For when a new migration is needed:
+
+1. Create a directory for the new version (e.g. `/v2`).
+2. Declare the new schema with appropriate changes.
+3. In `/v1/index.ts`, implement a `migrate` function that takes a `v1` schema and returns a `v2` schema.
+4. Update `/schemas/generators/index.ts` to use the new version:
+
+```ts
+function migrate(settings: z.infer<typeof AnySettingSchema>) {
+
+case '1.0':
+  return migrate(v1.migrate(settings))
+
+}
+```
+
+5. Make changes to the remaining implementation to use the new schema.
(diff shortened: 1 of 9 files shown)


allansson (changes requested):
Migration logic works! 💯 

One thing though. The following files are all a part of the schema so they should be moved to the `v1` folder:

```
src/schemas/rules.ts
src/schemas/testData.ts
src/schemas/testOptions.ts
```

And the instructions should be to copy the entire schema in the `v1` folder and rename it `v2`. If you don't copy, then you run the risk of changing previous versions of the schema e.g. by making changes to the `TestOptionsSchema`.

cristianoventura:
@allansson Nice catch, I've missed those. I've now updated the schemas + the readme.

allansson (approved):

src/schemas/generator/v1/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  testData: TestDataSchema,
+  rules: TestRuleSchema.array(),
+  allowlist: z.string().array(),
+  includeStaticAssets: z.boolean(),
+  scriptName: z.string().default('my-script.js'),
+})
+
+export type GeneratorSchema = z.infer<typeof GeneratorFileDataSchema>
+
+// TODO: Migrate generator to the next version
```
going-confetti, line 19: `v1` is essentially the same as `v0` here, right? I don't understand what do we need the TODO comment for - is it for future non-breaking changes to `v1`? If so, I'd prefer a regular descriptive comment, so that TODOs are reserved for debt

src/schemas/generator/README.md
```diff
@@ -0,0 +1,25 @@
...
+Migrations are needed when changing the structure of existing generators file such as renaming or deleting keys.
+
+If you're simply adding new options to the generator file, a migration may not be needed. In this case, extend the latest schema available.
+
+## Creating a new migration
+
+For when a new migration is needed:
+
+1. Create a directory for the new version (e.g. `/v2`).
+2. Copy files from the previous schema into this directory and make appropriate changes.
```
going-confetti, line 12: While out of scope for this PR, I think we should explore autogenerating new schema versions in future - instead of manually creating a copy of the `vN` folder and its content, we'd be able to run something like `whatever-code-gen-is-trendy-rn studio:generator` and have it do it automatically.

going-confetti (changes requested):
LGTM with a minor remark.
Apologies for `changes requested` - I only chose it because we recently merged an update to ESLint config and I wanted to make sure the branch is updated before merging.

src/schemas/generator/v1/index.ts
cristianoventura: They are the same schema, but the implementation of the `migrate` function is incomplete because there is nothing to migrate to since `v1` is the latest version.

The `TODO` comment here indicates that the implementation needs to change when another version exists (to migrate from `v1` to `v2`).

src/schemas/generator/README.md
cristianoventura: We can explore that separately and probably do the same for settings too! 👍

Issue #313, opened 2024-11-10
Ability to disable a Rule


Llandy3d:
We can create a new rule, move it and delete it. We are missing the disabling operation.

It could be useful to disable temporarily a specific rule when working with the application, we should introduce the option of disabling a single rule.

PR #375, opened 2024-12-05
feat: ability to disable a rule

cristianoventura:
## Description

This PR introduces the ability to disable/enable a test rule. Disable rules will not be generated by the script.

## How to Test

- Create a new Test Rule
- Disable the rule and check that it was not generated by the script
- Re-enable the same rule and check that it is now generated by the script

## Checklist

image

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/313

### DIFF
--- a/src/codegen/codegen.test.ts
+++ b/src/codegen/codegen.test.ts
@@ -157,11 +157,12 @@ describe('Code generation', () => {
       ).toBe(expectedResult.replace(/\s/g, ''))
     })
 
-    it('should replace correlated values', async () => {
+    describe('Correlation', () => {
       const rules: TestRule[] = [
         {
           type: 'correlation',
           id: '1',
+          enabled: true,
           extractor: {
             filter: { path: '/login' },
             selector: {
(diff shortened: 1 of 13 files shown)


e-fisher (commented):
Works as advertised! 🙌 

The verification rule doesn't have a status label, but I was thinking - maybe instead of added a label we could drop opacity to 0.5 and add hover tooltip "disabled" when rule is not enabled. 
The reasoning behind this: in most cases all my rules will be enabled and seeing "enabled" label repeated just adds noise. Here's how it could look like:

Screenshot 2024-12-06 at 14 49 02

cristianoventura:
> Works as advertised! 🙌
> 
> The verification rule doesn't have a status label, but I was thinking - maybe instead of added a label we could drop opacity to 0.5 and add hover tooltip "disabled" when rule is not enabled. The reasoning behind this: in most cases all my rules will be enabled and seeing "enabled" label repeated just adds noise. Here's how it could look like:

I like the idea of reducing noise! I've made the changes you suggested @e-fisher

Issue #277, opened 2024-10-25
Update eslint to use recommended-type-checked config and fix type errors


e-fisher:
Currently, we are not using the correct config for type checked project, which allows implicit `any` and other type-unsafe operations. To fix this:

1. Use `plugin:@typescript-eslint/recommended-type-checked` eslint config instead of `plugin:@typescript-eslint/recommended`  https://typescript-eslint.io/getting-started/typed-linting/

2. Fix type errors - as of writing, it shows 167 errors. 

This is needed to prevent hard to find bugs in the future and keep the project type-safe.

PR #383, opened 2024-12-17
chore: Stricter ESLint config

going-confetti:
Relevant issue: https://github.com/grafana/k6-studio/issues/277

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -7,7 +7,7 @@ module.exports = {
   extends: [
     'eslint:recommended',
     'plugin:@typescript-eslint/eslint-recommended',
-    'plugin:@typescript-eslint/recommended',
+    'plugin:@typescript-eslint/recommended-type-checked',
     'plugin:import/recommended',
     'plugin:import/electron',
     'plugin:import/typescript',
@@ -18,7 +18,11 @@ module.exports = {
     'prettier',
   ],
   parser: '@typescript-eslint/parser',
-  ignorePatterns: ['resources/group_snippet.js', 'install-k6.js'],
+  ignorePatterns: [
+    'resources/group_snippet.js',
+    'install-k6.js',
+    '.eslintrc.cjs',
+  ],
   plugins: [
     'import',
     'unused-imports',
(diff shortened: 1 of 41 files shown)


.eslintrc.cjs
```diff
@@ -43,11 +47,8 @@ module.exports = {
         argsIgnorePattern: '^_',
       },
     ],
-    // TODO: remove warnings when plugin:@typescript-eslint/recommended-type-checked is enabled
-    '@typescript-eslint/no-unsafe-assignment': 'warn',
-    '@typescript-eslint/no-unsafe-return': 'warn',
-    '@typescript-eslint/no-unsafe-argument': 'warn',
-    '@typescript-eslint/no-unsafe-member-access': 'warn',
+    '@typescript-eslint/no-floating-promises': 'off',
+    '@typescript-eslint/no-misused-promises': 'off',
```
going-confetti, line 51: I suggest turning those on later. It highlighted some cases where we pass async functions as e.g. click handlers

src/codegen/codegen.utils.ts
```diff
@@ -24,6 +24,8 @@ export function stringify(value: unknown): string {
     return `{${properties}}`
   }
 
+  // TODO: https://github.com/grafana/k6-studio/issues/277
```
going-confetti, line 27: Marked ignore comments with a link to the issue (I'll update the issue description once this PR is merged).
While it does look annoying, the pros of enabling stricter rules for everything else outweigh the cons

src/views/Validator/Validator.hooks.ts
```diff
@@ -2,7 +2,8 @@ import { useLocation, useParams } from 'react-router-dom'
 
 export function useScriptPath() {
   const { fileName } = useParams()
-  const { state } = useLocation()
+  // TODO(router): useLocation is not type-safe. Refactor this route to avoid using it.
+  const { state } = useLocation() as { state: { externalScriptPath: string } }
```
going-confetti, line 6: @allansson now I see what you meant, and I barely scratched the surface...

allansson: The same goes for getting route params via `useParams` or search params via `useSearchParams`. You're just casting things hoping that your routes are structured in a certain way.

e-fisher (dismissed):

.eslintrc.cjs
going-confetti: Initially, I was reluctant to do this, as the number of warning seemed to be too high with quite a few of them not being useful. I then discussed possible solutions with @e-fisher and we found a way to enable these rules with a few tweaks.

PR #384, opened 2024-12-18
chore: Use tanstack query for settings

going-confetti:
## How to Test
1. Open the settings dialog
2. Check if tabs still work
3. Try changing a setting and saving it

## Checklist

### DIFF
--- a/src/components/Settings/LogsSettings.tsx
+++ b/src/components/Settings/LogsSettings.tsx
@@ -43,14 +43,14 @@ export function LogsSettings() {
   return (
     <SettingsSection>
       <Flex mb="4" justify="between" align="center">
-        <Flex>
+        <Flex align="center">
           <Spinner mr="2" />
-          <Text size="2">
-            k6 Studio logs in this screen are updated in real-time.
+          <Text size="2" color="gray">
+            Application logs are updated in real-time.
           </Text>
         </Flex>
         <Button onClick={handleOpenLogClick} variant="outline">
-          Open log location
+          Open logs folder
         </Button>
       </Flex>
 
--- a/src/components/Settings/SettingsDialog.tsx
+++ b/src/components/Settings/SettingsDialog.tsx
@@ -3,7 +3,7 @@ import { zodResolver } from '@hookform/resolvers/zod'
 import { Box, Button, Dialog, Flex, ScrollArea, Tabs } from '@radix-ui/themes'
 import { ExclamationTriangleIcon } from '@radix-ui/react-icons'
 import { findIndex, sortBy } from 'lodash-es'
-import { useEffect, useState } from 'react'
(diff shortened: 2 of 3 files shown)


src/components/Settings/LogsSettings.tsx
```diff
@@ -45,14 +45,14 @@ export function LogsSettings() {
...
       <Flex mb="4" justify="between" align="center">
-        <Flex>
+        <Flex align="center">
           <Spinner mr="2" />
-          <Text size="2">
-            k6 Studio logs in this screen are updated in real-time.
+          <Text size="2" color="gray">
+            Application logs are updated in real-time.
           </Text>
         </Flex>
```
going-confetti, line 53: I made it a bit more subtle, but I'm not quite happy with the spinner here: IMO it gives an indication that something is constantly fetching, even though it's not the case.

src/components/Settings/SettingsDialog.tsx
```diff
@@ -65,23 +60,6 @@ export const SettingsDialog = ({ open, onOpenChange }: SettingsDialogProps) => {
     formState: { isDirty, errors },
   } = formMethods
 
-  const onSubmit = async (data: AppSettings) => {
```
going-confetti, line 68: using tanstack query is more ergonomic and allows us to reused cached settings across different components

src/components/Settings/LogsSettings.tsx
going-confetti, line 1: Not related to this PR, but I think it'd be really nice to add basic syntax highlighting to logs
image

allansson (approved):
Left a suggestion, but otherwise it's legit. 💯

src/components/Settings/SettingsDialog.tsx
```diff
@@ -41,17 +42,12 @@ const tabs = [
 ]
 
 export const SettingsDialog = ({ open, onOpenChange }: SettingsDialogProps) => {
-  const [settings, setSettings] = useState<AppSettings>()
-  const [submitting, setSubmitting] = useState(false)
+  const { data: settings } = useSettings()
+  const { mutateAsync: saveSettings, isPending } = useSaveSettings(() => {
+    onOpenChange(false)
+  })
```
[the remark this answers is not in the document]
going-confetti: It's a bit weird, because of how the save handler in main doesn't throw an error, but returns `false` instead. This means that a save error will still be considered "success", but at the same time we don't want the dialog to close if it wasn't actually saved. 

I think the handler should just re-throw the error, but I also don't want to make any changes to it now, because we have the same pattern in a few other places :\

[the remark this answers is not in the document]
allansson: I see. Well, you have my blessing! 👍

PR #388, opened 2024-12-19
fix(windows): Chrome folder is shown when no starting URL is provided

going-confetti:
## Description

On Windows, if the user doesn't provide a URL before starting a recording, the browser shows the Chrome folder.
image

## How to Test

On a windows machine:
1. Open the app
2. Go to Recorder
3. Click `Start recording` while leaving the `Starting URL` blank
4. Verify that the browser is launched with an empty page

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -70,7 +70,7 @@ export const launchBrowser = async (
       `--proxy-server=http://localhost:${appSettings.proxy.port}`,
       `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
       disableChromeOptimizations,
-      url ?? 'about:blank',
+      url || 'about:blank',
     ],
     onExit: sendBrowserClosedEvent,
   })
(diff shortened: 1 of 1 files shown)


src/browser.ts
```diff
@@ -70,7 +70,7 @@ export const launchBrowser = async (
       `--proxy-server=http://localhost:${appSettings.proxy.port}`,
       `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
       disableChromeOptimizations,
-      url ?? '',
+      url ?? 'about:blank',
```
e-fisher, line 73: I'm still able to reproduce the issue on Windows. My guess `url` is an empty string and doesn't trigger nullish operator fallback, the `||` operator should do the trick. 

You can also verify this on macos: about:blank should open as an empty page, currently it opens to chrome homepage with google search.

e-fisher (changes requested):

src/browser.ts
allansson: Might be worth to also to `url.trim()` to handle the user entering a bunch of spaces.

going-confetti: You're right, I didn't even notice this on MacOS 🤦‍♂️

e-fisher (approved):
🚀

PR #389, opened 2024-12-19
fix: Last item scripts list isn't always accessible

going-confetti:
## Description
See attached screenshots

## How to Test
* Resize the window to the min supported height
* Make sure you have enough files for the scrollbar to appear
* Verify that the last item is clearly visible

Before
image

After
image


e-fisher (approved):
LGTM 🙌

PR #390, opened 2024-12-20
feat: Save and load browser events

allansson:
## Description

This PR implements saving and loading of browser events. For now it saves the events as an extended property `_browserEvents` on the `log` object of the HAR file. While we're looking in to creating a new format to store both request and browser events, this is a short term solution.

The PR also includes a simple list view of the browser events. When the browser event recording is enabled, the recording/preview views will have two tabs: one for requests and another for the browser events.

Currently, only a dummy event sent on an interval is being recorded.

## How to Test

1. Enable browser feature.
2. Start a recording.
3. You should see a tabbed view.
4. Check the Browser Events tab.
5. Make sure events trickle in.
6. Stop recording.
7. You should still see the events when reloading the recording.
8. Disable browser feature.
9. Check that the app works like it would in main.

## Checklist

Screenshot 2024-12-20 at 15 07 59


src/components/WebLogView/WebLogView.tsx
```diff
@@ -10,7 +10,6 @@ import { useDeepCompareEffect } from 'react-use'
 interface WebLogViewProps {
   requests: ProxyDataWithMatches[]
   groups?: GroupType[]
-  activeGroup?: string
```
allansson, line 13: The `activeGroup` prop was not used anywhere so I removed it here and any place it was being passed down.

src/schemas/recording/index.ts
```diff
@@ -1 +1,2 @@
 export * from './v1/browser'
+export { Recording } from './v1/recording'
```
allansson, line 2: I'm pretending we have an actual format for recordings.

src/views/Recorder/Recorder.tsx
```diff
@@ -205,47 +194,33 @@ export function Recorder() {
...
-        {selectedRequest !== null && (
-          <Allotment.Pane minSize={300}>
-            <Details
-              selectedRequest={selectedRequest}
-              onSelectRequest={setSelectedRequest}
-            />
-          </Allotment.Pane>
+      {recorderState === 'idle' && (
+        <EmptyState isLoading={isLoading} onStart={handleStartRecording} />
+      )}
```
allansson, line 199: This is the view where you enter the start url. It was previously handled in `RequestsSection`. I moved it here because:

1. It's only applicable when recording.
2. With the tabbed interface, it would appear in the requests tab.

src/views/Recorder/Recorder.tsx
```diff
@@ -205,47 +194,33 @@ export function Recorder() {
...
+      {recorderState !== 'idle' &&
+        !settings?.recorder.enableBrowserRecorder && (
+          <RequestLog
+            recorderState={recorderState}
+            groups={groups}
+            requests={debouncedProxyData}
+            onUpdateGroup={handleUpdateGroup}
+            onResetRecording={handleResetRecording}
+            onCreateGroup={handleCreateGroup}
+          />
```
allansson, line 222: I'm trying to keep the normal UI when you disable the browser feature. Maybe it makes more sense to always show the tabbed interface and disable/hide the "Browser Events" tab?

src/views/Recorder/RequestLog.tsx
allansson, line 1: `Recorder` and `RecordingPreviewer` had the same layout with the exception of some recording related buttons and such. This is my attempt at having a re-usable component between the two.

src/views/RecordingPreviewer/RecordingPreviewer.tsx
```diff
@@ -113,25 +118,17 @@ export function RecordingPreviewer() {
...
-        {selectedRequest !== null && (
-          <Allotment.Pane minSize={300}>
-            <Details
-              selectedRequest={selectedRequest}
-              onSelectRequest={setSelectedRequest}
-            />
-          </Allotment.Pane>
-        )}
-      </Allotment>
+      {!isLoading && settings?.recorder.enableBrowserRecorder && (
```
allansson, line 121: We check `isLoading` here because otherwise there will be a flicker when switching between recordings (probably due to the way we do `setProxyData([])` in the `useEffect`-disposer).

src/views/RecordingPreviewer/RecordingPreviewer.tsx
```diff
@@ -10,19 +9,23 @@ import {
...
 
 export function RecordingPreviewer() {
+  const { data: settings } = useSettings()
+
   const [proxyData, setProxyData] = useState<ProxyData[]>([])
-  const [isLoading, setIsLoading] = useState(false)
-  const [selectedRequest, setSelectedRequest] = useState<ProxyData | null>(null)
+  const [browserEvents, setBrowserEvents] = useState<BrowserEvent[]>([])
+
+  const [isLoading, setIsLoading] = useState(true)
```
allansson, line 28: We'll always be in a loading state when the component mounts, so this should be `true` by default.

src/views/Recorder/BrowserEventLog.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+interface BrowserEventLogProps {
+  events: BrowserEvent[]
+}
+
+export function BrowserEventLog({ events }: BrowserEventLogProps) {
+  return (
+    <Flex direction="column" minHeight="0" height="100%">
+      <Allotment defaultSizes={[1]}>
```
allansson, line 46: Using allotment here because it sets a fixed height on the element and that makes the `ScrollArea` work correctly. Please let me know if you know any better way, but I've tried settings flex here and flex there and it just won't do it for me.

going-confetti (approved):
Looks good and works well 🚀

Issue #321, opened 2024-11-10
Caret not changing when collapsing groups in the inline validator


Llandy3d:
The caret icon is broken in the inline validator in the Generator view, the same icon is shown both on collapse and not.

![Image](https://github.com/user-attachments/assets/d2d83f95-e23b-416f-99b1-874f691fe385)

PR #391, opened 2024-12-24
fix: Caret icon doesn't change state in Validator dialog

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/321

## How to Test
* Open the validator dialog
* Run a script
* Verify that the group sections have expected caret state

image

### DIFF
--- a/src/components/Collapsible.tsx
+++ b/src/components/Collapsible.tsx
@@ -37,7 +37,7 @@ const Header = styled.div`
 const CaretIcon = styled(CaretRightIcon)`
   transform: rotate(0deg);
 
-  [data-state='open'] & {
+  button[data-state='open'] & {
     transform: rotate(90deg);
   }
 `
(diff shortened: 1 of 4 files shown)


src/components/Collapsible.tsx
```diff
@@ -37,7 +37,7 @@ const Header = styled.div`
 const CaretIcon = styled(CaretRightIcon)`
   transform: rotate(0deg);
 
-  [data-state='open'] & {
+  button[data-state='open'] & {
```
going-confetti, line 40: Since there's a dialog with `data-state="open"`, this was always true

src/components/WebLogView/WebLogView.tsx
```diff
@@ -10,7 +10,6 @@ import { useDeepCompareEffect } from 'react-use'
 interface WebLogViewProps {
   requests: ProxyDataWithMatches[]
   groups?: GroupType[]
-  activeGroup?: string
```
going-confetti, line 13: Unused prop

cristianoventura (approved):
👍

PR #365, opened 2024-11-27
feat: implement crash reporter

cristianoventura:
## Description

This PR integrates Sentry so we can collect information about crashes and other errors in the main process for production builds.

## How to Test

This integration only works in production. You can either:

- Option 1: test the artifacts for the temporary [0.8.1 release](https://github.com/grafana/k6-studio/releases/tag/untagged-48a1dccb85075cb63112). This release has [one commit ahead](https://github.com/grafana/k6-studio/commit/83fb0d83b00efb9e7baf9e7a8e97d384dba98087) of this branch that calls an undefined function on startup which will report to Sentry
- Option 2: change `main.ts` locally to work in development environment, start the app setting the `SENTRY_DSN` variable (which is available in the project's config in Sentry), and manually throw an error. (source maps will not be available in development mode)

Notes:

- Secrets and env variables are stored in this repo settings
- Sentry DSN is statically replaced during build time and shipped within the release artifacts
- Optionally test the integration locally with `SENTRY_DSN=xx npm start`
- You can simulate a crash using `process.crash()` anywhere in the main process

## Checklist

- [x] I have performed a self-review of my code.
- [ ] I have added tests for my changes.
- [x] I have run linter locally (`npm run lint`) and all checks pass.
- [x] I have run tests locally (`npm test`) and all tests pass.
- [ ] I have commented on my code, particularly in hard-to-understand areas.

## Screenshots (if appropriate):

image

## Related PR(s)/Issue(s)

 Resolves https://github.com/grafana/k6-studio/issues/349

Issue #392, opened 2024-12-27
Automate schema migrations


cristianoventura:
Currently, creating migrations for settings and generators requires a manual process where we copy the directory of the latest version and then implement the necessary changes from there.

We'd like to automate the creation of migrations for settings and generators through a script, which would:

- Identify the latest version directory.
- Automatically copy the relevant files or templates needed for migration.
- Set the latest version in the `migrate` function that lives in `schemas/settings/index.ts` and `schemas/generators/index.ts`

Issue #378, opened 2024-12-12
Split main.ts in separate modules


cristianoventura:
The current `main.ts` file has grown significantly in size and complexity, making it difficult to maintain and scale over time. As new features and changes are added, the file has become harder to navigate and troubleshoot. 

Following patterns adopted in [other open-source Electron projects](https://github.com/sindresorhus/awesome-electron?tab=readme-ov-file#apps), we'd like to split the logic into separate files/modules that have a clear and single responsibility.

Example:

```
├── src
│   ├─ main
│   ├───── browser.ts
│   ├───── ipc.ts
│   ├───── logger.ts
│   ├───── main.ts
│   ├───── mainWindow.ts
│   ├───── menu.ts
│   ├───── proxy.ts
│   ├───── script.ts
│   └───── splashscreenWindow.ts
```

PR #393, opened 2024-12-31
feat: crash reporter settings

cristianoventura:
## Description

This PR depends on https://github.com/grafana/k6-studio/pull/365. It adds a new option to the Settings dialog that allows users to opt out of k6 Studio sending error/crash reports to Sentry.

## How to Test

- manually throw an error anywhere in the application that is triggered on user click (e.g. toggle theme)
- toggle the setting and check that the error event is not sent when the setting is disabled

## Checklist

Screenshot 2024-12-30 at 11 04 28 AM


src/components/Settings/TelemetrySettings.tsx
```diff
@@ -33,6 +33,24 @@ export const UsageReportSettings = () => {
...
+          name="telemetry.errorReport"
+          render={({ field }) => (
+            <Text size="2" as="label">
+              <Checkbox
+                {...register('telemetry.errorReport')}
+                checked={field.value}
+                onCheckedChange={field.onChange}
+              />{' '}
+              Send crash reports and error data to Grafana to help improve k6
+              Studio. <Link href="">Learn more.</Link>
```
cristianoventura, line 49: We'll likely need a link here leading to a doc similar to the usage collection.

src/main.ts
```diff
@@ -66,6 +66,14 @@ if (process.env.NODE_ENV !== 'development') {
   Sentry.init({
     dsn: SENTRY_DSN,
     integrations: [Sentry.electronMinidumpIntegration()],
+
+    // conditionally send the event based on the user's settings
+    beforeSend: (event) => {
+      if (appSettings.telemetry.errorReport) {
+        return event
+      }
```
cristianoventura, line 74: Sentry is always initialized in Prod mode, but the event is dynamically sent based on the `errorReport` setting.

src/schemas/settings/v1/index.ts
```diff
@@ -117,7 +117,10 @@ export function migrate(
     proxy: settings.proxy,
     recorder: settings.recorder,
     windowState: settings.windowState,
-    usageReport: settings.usageReport,
+    telemetry: {
+      usageReport: settings.usageReport.enabled,
+      errorReport: true,
+    },
```
cristianoventura, line 123: A migration was needed since `usageReport` was renamed to `telemetry

e-fisher: Don't we need to bump schema version, copy schema to v3 and add the adjustments there? 
Everything works now, but if we introduce another change that depends on `useageReport` or `telemetry` and the user has skipped one version between updates - the migration won't work anymore because we lose that middle step.

e-fisher (commented):
Added comments about migration and empty link

src/schemas/settings/v3/index.ts
```diff
@@ -0,0 +1,36 @@
...
+export {
+  AppearanceSchema,
+  ProxySettingsSchema,
+  RecorderSettingsSchema,
+  TelemetrySchema,
+  WindowStateSchema,
+}
+
+export const AppSettingsSchema = z.object({
+  version: z.literal('3.0'),
```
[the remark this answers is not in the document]
cristianoventura: Since all migrations are required (and likely cause breaking changes) and automatically fast-forwarded, I don't think bumping the minor would ever be suitable 🤔. We could use timestamps instead, like other major frameworks do. This change could also be incorporated into https://github.com/grafana/k6-studio/issues/392 where the migration name and version are automatically generated.

src/main.ts
```diff
@@ -66,6 +66,14 @@ if (process.env.NODE_ENV !== 'development') {
   Sentry.init({
     dsn: SENTRY_DSN,
     integrations: [Sentry.electronMinidumpIntegration()],
+
+    // conditionally send the event based on the user's settings
+    beforeSend: (event) => {
+      if (appSettings.telemetry.errorReport) {
```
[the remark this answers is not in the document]
cristianoventura: Good catch! I chose to initialize the variable with the `defaultSettings` object instead. If we leave it as `undefined`, we'd need to perform additional type checks in multiple places, which could increase the overall maintenance effort.

e-fisher (approved):
LGTM 🚀

src/main.ts
```diff
@@ -79,7 +88,7 @@ const PROXY_RETRY_LIMIT = 5
 let proxyRetryCount = 0
 let currentClientRoute = '/'
 let wasAppClosedByClient = false
-export let appSettings: AppSettings
+export let appSettings = defaultSettings
```
going-confetti, line 91: Not sure how I feel about this, as it can lead to a situation when `defaultSettings` would be used instead of the actual settings. I see that it's only imported in `browser.ts` - can we add `detectBrowserPath` as a parameter to `launchBrowser` instead?

src/schemas/settings/v3/index.ts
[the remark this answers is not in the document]
going-confetti: > We could use timestamps instead, like other major frameworks do. This change could also be incorporated into https://github.com/grafana/k6-studio/issues/392 where the migration name and version are automatically generated.

I like this idea!

src/main.ts
cristianoventura: > Not sure how I feel about this, as it can lead to a situation when defaultSettings would be used instead of the actual settings

@going-confetti This already happens. The `getSettings` method returns the default object if the settings file can't be parsed. It's a fault tolerance mechanism. `appSettings` will be set at some point (either to the actual settings or failover to default)

https://github.com/grafana/k6-studio/blob/ff74970cf00262bea0ffa86f5b30b66f754c9f68/src/settings.ts#L68-L72


> I see that it's only imported in browser.ts - can we add detectBrowserPath as a parameter to launchBrowser instead?

This is a good idea! There will be probably additional places where we'll have to use `appSettings` with the separation of modules that we plan to work on https://github.com/grafana/k6-studio/issues/378, so I'll defer this change to when we get there.

going-confetti: > This already happens. The getSettings method returns the default object if the settings file can't be parsed. It's a fault tolerance mechanism. appSettings will be set at some point (either to the actual settings or failover to default)

IMO it could be a big deal if the settings can't be parsed and we use default settings as a fallback without informing the user. For example, if they have custom proxy configuration and stuff suddenly stops working.
Seems out of scope of this PR though

going-confetti (approved):

src/main.ts
cristianoventura: Agreed! We can work on something to let users know that the default config was loaded if the parsing fails.

Issue #396, opened 2025-01-07
Implement Sentry in the renderer process


cristianoventura:
Once #365 is implemented, we need to make sure crashes and errors that happen in the renderer process are also captured and sent to Sentry.

PR #397, opened 2025-01-07
feat: enable Sentry in renderer process

cristianoventura:
## Description

This PR enables the Sentry integration in the renderer process, accounting for the user's preference to opt out of it.  This PR is dependant on https://github.com/grafana/k6-studio/pull/393.

## How to Test

- Enable Sentry to run the development environment (both in `main.ts` and `renderer.ts`)
- Manually throw an error in the React code
- Run the app with the `SENTRY_DSN` env variable set (grab it from the project in Sentry) `SENTRY_DSN=xx npm start`
- Reproduce the error in runtime and check Sentry

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/396

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -29,8 +29,31 @@
 import { ProxyData } from './types'
 import './index.tsx'
 import { setMonacoEnv } from './components/Monaco/setMonacoEnv'
+import * as Sentry from '@sentry/electron/renderer'
 
 setMonacoEnv()
+
+// Sentry integration
+if (process.env.NODE_ENV !== 'development') {
+  Sentry.init({
+    integrations: [
+      Sentry.browserTracingIntegration(),
+      Sentry.replayIntegration(),
+    ],
+    tracesSampleRate: 1.0,
+    replaysSessionSampleRate: 0.1,
+    replaysOnErrorSampleRate: 1.0,
+
+    // conditionally send the event based on the user's settings
+    beforeSend: async (event) => {
+      const { telemetry } = await window.studio.settings.getSettings()
+      if (telemetry.errorReport) {
+        return event
+      }
+      return null
+    },
+  })
+}
 
 // Proxy
 


src/renderer.ts
```diff
@@ -29,9 +29,32 @@
...
+// Sentry integration
+if (process.env.NODE_ENV !== 'development') {
+  Sentry.init({
+    integrations: [
+      Sentry.browserTracingIntegration(),
+      Sentry.replayIntegration(),
+    ],
+    tracesSampleRate: 1.0,
+    replaysSessionSampleRate: 0.1,
+    replaysOnErrorSampleRate: 1.0,
```
cristianoventura, line 45: Enabled some basic integrations such as browser tracing and session replay, which may help us debug issues when they arise.

e-fisher (commented):
Unhandled promise rejections are not reported to sentry, to reproduce it: add a throw to an async method, I've used `validateAndSaveHarFile` in `views/Recorder/Recorder.tsx`. 

I've skimmed through docs and looks like `sentry/react` integration should handle this automatically if we add it additionally to `sentry/electron`, here's an example how to setup additional SDK: https://docs.sentry.io/platforms/javascript/guides/electron/#framework-specific-sdks

screencapture 2025-01-08 at 09 40 00

Scutua (approved):

src/renderer.ts
going-confetti: Should we consider filtering events to scrub sensitive data?

src/sentry.ts
```diff
@@ -0,0 +1,33 @@
...
+        SentryRenderer.browserTracingIntegration(),
+        SentryRenderer.replayIntegration(),
+      ],
+      tracesSampleRate: 1.0,
+      replaysSessionSampleRate: 0.1,
+      replaysOnErrorSampleRate: 1.0,
+
+      // conditionally send the event based on the user's settings
+      beforeSend: async (event) => {
+        const getSettings = getSettingsFn || window.studio.settings.getSettings
```
cristianoventura, line 19: `windows` isn't available in the preload context so I'm adding flexibility here to pass the `getSettings` function as an argument optionally.

cristianoventura, line 19: `window` isn't available in the preload context so I'm adding flexibility here to pass the getSettings function as an argument optionally.

cristianoventura (commented):
@e-fisher Unfortunately `@sentry/react` didn't work and after spending some time trying to figure out why I learned the following.

The unhandled rejection gets raised inside the callback of the `onBrowserClosed` subscription, which is where `validateAndSaveHarFile` is being called.

https://github.com/grafana/k6-studio/blob/898bde88d184d452a020114324ddf2afd02bfede/src/views/Recorder/Recorder.tsx#L153-L155

However, `onBrowserClosed` is declared in the preload context (not in the main `renderer`).

Electron creates two separate environments for the renderer process: the main renderer and the preload script. Sentry was only initialized in the `renderer.ts` file, which prevented it from capturing errors originating from functions or callbacks within the `preload` context.

The problem was solved by:

- initializing Sentry in the preload and renderer contexts
- adding an event listener for `unhandledrejection` to catch unhandled promise rejections

e-fisher (dismissed):
Looks like that did the trick! 🙌 
~I've noticed this error in the console when sentry is enabled, otherwise LGTM~ No longer reproducible

```
Refused to create a worker from 'blob:http://localhost:5173/45616690-949f-4daf-a53c-4fc7e0a2906f' because it violates the following Content Security Policy directive: "script-src 'self'". Note that 'worker-src' was not explicitly set, so 'script-src' is used as a fallback.
```

cristianoventura:
I've removed the integrations as we've discussed today!

e-fisher (approved):

Issue #399, opened 2025-01-09
Generator prompts for saving when no changes were made


cristianoventura:
### Brief summary

When a generator is set to version: "0" and no changes are made, the UI still prompts the user to save changes.

### k6 Studio version

main

### OS

macOS

### Steps to reproduce the problem

1. Set a generator's version to 0.
2. Don't make any changes to the generator.
3. Navigate away or try to exit the page.
4. Observe that the UI still prompts for changes, even though no modifications were made.

The "Save generator" button is also clickable as a result of the `isDirty` variable being `true`.

### Expected behaviour

After the state has been migrated from version 0 to 1, the file should contain the migrated of the schema.

### Actual behaviour

File content differs from state, resulting in changes being detected.

PR #400, opened 2025-01-09
fix: save generator after migration

cristianoventura:
## Description

This PR fixes an issue where the UI detects changes in the generator if it has been migrated during runtime.

## How to Test

- Manually set the `version` property of a generator to `0`
- Open the generator in the UI
- Check that no changes are detected
- Check that the file has been migrated to `version: 1.0`

## Checklist

https://github.com/user-attachments/assets/3fdb18ab-5da2-4419-a619-db5350b10168

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/399

### DIFF
--- a/src/schemas/generator/v0/index.ts
+++ b/src/schemas/generator/v0/index.ts
@@ -21,12 +21,12 @@ export type GeneratorSchema = z.infer<typeof GeneratorFileDataSchema>
 export function migrate(generator: GeneratorSchema): v1.GeneratorSchema {
   return {
     version: '1.0',
+    recordingPath: generator.recordingPath,
+    options: generator.options,
+    testData: generator.testData,
+    rules: generator.rules,
     allowlist: generator.allowlist,
     includeStaticAssets: generator.includeStaticAssets,
-    options: generator.options,
-    recordingPath: generator.recordingPath,
-    rules: generator.rules,
     scriptName: generator.scriptName,
-    testData: generator.testData,
   }
 }
--- a/src/views/Generator/Generator.utils.ts
+++ b/src/views/Generator/Generator.utils.ts
@@ -47,9 +47,30 @@ export const writeGeneratorToFile = (
   )
 }
 
-export const loadGeneratorFile = async (fileName: string) => {
+export const isGeneratorDirty = (
+  generatorStateData: Partial<GeneratorFileData>,
+  generatorFileData: Partial<GeneratorFileData>
+) => {
+  // Convert to JSON instead of doing deep equal to remove
+  // `property: undefined` values
+  return (
+    JSON.stringify(generatorStateData) !== JSON.stringify(generatorFileData)
+  )
+}
+
+export const loadGeneratorFile = async (
+  fileName: string,
+  migrate?: boolean
+) => {
   const generatorFile = await window.studio.generator.loadGenerator(fileName)
-  return GeneratorFileDataSchema.parse(generatorFile.content)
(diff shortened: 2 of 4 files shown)


src/views/Generator/Generator.utils.ts
```diff
@@ -47,9 +47,30 @@ export const writeGeneratorToFile = (
   )
 }
 
-export const loadGeneratorFile = async (fileName: string) => {
+export const isGeneratorDirty = (
+  generatorStateData: Partial<GeneratorFileData>,
+  generatorFileData: Partial<GeneratorFileData>
+) => {
+  // Convert to JSON instead of doing deep equal to remove
```
going-confetti, line 54: I have a question not related to this particular PR:
I think it shouldn't be a problem here, but since JSON.stringify does not guarantee the order of keys, can two equal objects be considered unequal?

going-confetti (approved):

src/views/Generator/Generator.utils.ts
e-fisher: That's possible. We might want to switch to `isEqual` if this proves to be a problem.

e-fisher: IIRC, I've used JSON.stringify because generatorFileData would include all the undefined values, but stored JSON wouldn't have them, and if would show up as false dirty. So we'd need to do some cleanup if we switch to `isEqual`

e-fisher (commented):
Works like a charm! 🙌 Though, I'm a bit worried about the "hidden" write operation when opening a generator. If we mess up a migration, could this lead to broken generators? Without this change it would still break them on save, but I imagine that would be less impactful, because failing to open a generator I would try others, corrupting them as well. 

As an alternative to consider: could we migrate in memory while doing the `isDirty` check, without writing anything to the file? And only save new migrated version to file after user modified and saved generator.

Llandy3d:
> Works like a charm! 🙌 Though, I'm a bit worried about the "hidden" write operation when opening a generator. If we mess up a migration, could this lead to broken generators? Without this change it would still break them on save, but I imagine that would be less impactful, because failing to open a generator I would try others, corrupting them as well.
> 
> As an alternative to consider: could we migrate in memory while doing the `isDirty` check, without writing anything to the file? And only save new migrated version to file after user modified and saved generator.

This sounds like a reasonable and neat solution 👀 
I wonder if we should also warn that the generator needs to be updated and possibly ask if checking and updating all of them is wanted, this last bit could possibly be something for the future

src/schemas/generator/v0/index.ts
```diff
@@ -21,12 +21,12 @@ export type GeneratorSchema = z.infer<typeof GeneratorFileDataSchema>
...
     version: '1.0',
-    allowlist: generator.allowlist,
-    includeStaticAssets: generator.includeStaticAssets,
-    options: generator.options,
     recordingPath: generator.recordingPath,
+    options: generator.options,
+    testData: generator.testData,
     rules: generator.rules,
+    allowlist: generator.allowlist,
+    includeStaticAssets: generator.includeStaticAssets,
```
cristianoventura, line 29: The issue was actually related to the order of the fields in the migration, which caused the `JSON.stringify` comparison to return `false`. I made a ticket to address this behaviour https://github.com/grafana/k6-studio/issues/403

cristianoventura (commented):
I've removed the hidden write operation and let Zod to parse the state internally.

PR #402, opened 2025-01-13
feat: Export skeleton browser script

allansson:
## Description

image

This PR adds the foundation for exporting a browser recording as a script. It currently only exports a skeleton script with the correct options and a log statement for every dummy event recorded. Future PR:s will be focused on capturing specific events and emitting the correct code.

## How to Test

1. Start a recording and let some dummy browser events trickle in.
2. Stop the recording.
3. Click the "Export browser script" in the top-right dotted menu.
4. Run the generated script using the validator.

You should get a script with a log statement per dummy event and the script should run successfully in the validator.

## Checklist

- [X] I have added tests for my changes.

## Related PR(s)/Issue(s)

https://github.com/grafana/k6-cloud/issues/3091


src/codegen/browser/types.ts
allansson, line 1: This schema is an intermediate representation of a `Test`. The idea is that a stream of events gets converted into this representation and then that representation is used to emit code. This decouples the code generation logic from how the test is generated (e.g. via a test generator or a visual editor).

Right now it uses the `BrowserEvent` type, but as we start capturing actual events this type will become more specific.

src/views/RecordingPreviewer/RecordingPreviewer.tsx
```diff
@@ -118,6 +150,17 @@ export function RecordingPreviewer() {
               </IconButton>
             </DropdownMenu.Trigger>
             <DropdownMenu.Content>
+              {settings?.recorder.enableBrowserRecorder && (
```
allansson, line 153: Having it in the dropdown is not the prettiest, but it doesn't really matter right now. I will be collecting feedback on the UI once we can record something of value.

e-fisher (approved):
LGTM

Llandy3d (approved):
🚀

Issue #332, opened 2024-11-11
Search functionality for requests inspector


Llandy3d:
We are able to filter requests in the requests inspector but it would be useful to be able to search for a specific string and filter the list based on which requests contain the desired value, wether it's in the body, headers,...

Ideally we highlight where the value was found in the request when the detail view is open.

PR #406, opened 2025-01-14
feat: Search all request data

e-fisher:
## Description

Add ability to search through all request data.

- Toggle full search button in search input
- List results with highlighted matches
- Open corresponding details tab upon clicking a result
- Sync search term to code editor when clicking a body result to highlight matches

## Current limitations
The performance is not great with big recording, but that's because of the number of rendered rows, search itself is quick. I'll open a new issue to explore virtualized tables to improve performance.

## How to Test

Try different with different recordings, verify expected results can be found. Test in recorder, recorder previewer and validator.

## Checklist

https://github.com/user-attachments/assets/efd8d334-195c-47c0-a96c-399286fb11e6

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/332


src/components/WebLogView/SearchResults.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function ResultContent({ segment }: { segment: Segment }) {
+  return (
+    <>
+      {/* Show ellipsis on the left side: wrapped ltr inside rtl is needed */}
+      {/* to prevent swapping sides of punctuation marks */}
+      {segment.beforeMatch.length > 0 && (
+        <Text css={{ flexShrink: 1, whiteSpace: 'pre' }} dir="rtl" truncate>
+          <Text dir="ltr" css={{ whiteSpace: 'pre' }}>
```
going-confetti, line 146: This solution will break RTL content. Can we find an alternative?
Not sure if that's the reason, but if you go to `https://ar.wikipedia.org/wiki/%D8%A8%D8%B7%D8%A7%D8%B7%D8%A7` while recording, and then tried to search for `البَطَاطا`, you won't see any matches, even though the page contains plenty:

image

In fact, if you check the HTML content in the inspector, you'll see that the text looks like it was parsed using a wrong encoding (which is probably not an issue introduced here - I'll check)

going-confetti (changes requested):
Looks good apart from the performance issues you described.
I noticed an issue with RTL (and probably BiDi) text in recordings.

Also, there's a bug where if you select a request and append characters to your search query to the point when this requests shouldn't be matched anymore, then you'll still see the request details with the old highlighting:
image

cristianoventura (approved):
I've tested with a few recordings including the preview and validator and, other than what was already mentioned (including performance), this looks amazing to me! 🤩

src/components/WebLogView/SearchResults.tsx
e-fisher: > This solution will break RTL content. Can we find an alternative? Not sure if that's the reason, but if you go to `https://ar.wikipedia.org/wiki/%D8%A8%D8%B7%D8%A7%D8%B7%D8%A7` while recording, and then tried to search for `البَطَاطا`, you won't see any matches, even though the page contains plenty:

> In fact, if you check the HTML content in the inspector, you'll see that the text looks like it was parsed using a wrong encoding (which is probably not an issue introduced here - I'll check)

The reason for messed arabic characters was [atob](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) - it supports ASCII character but not UTF-8, I've added a helper method to fix that. 

As for this solution breaking RTL content - I don't have a lot of experience with RTL languages, but I would expect that source code HTML (which we show in search preview) is still in LTR format. Now that encoding is fixed it's easier to debug, let me know if you still find RTL related issues.

e-fisher:
Ready for another review! 
Created [issue](https://github.com/grafana/k6-studio/issues/423) to address performance issues

src/components/WebLogView/SearchResults.tsx
going-confetti: > The reason for messed arabic characters was [atob](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) - it supports ASCII character but not UTF-8, I've added a helper method to fix that.

Nice, really cool fix!

> As for this solution breaking RTL content - I don't have a lot of experience with RTL languages, but I would expect that source code HTML (which we show in search preview) is still in LTR format.

The HTML tags are in LTR, but the content is in RTL. This is called [BiDi/Bidirectional text](https://en.wikipedia.org/wiki/Bidirectional_text) - you can see how it works in `monaco` by placing the cursor in the middle of an opening html tag name with an RTL content and pressing the right arrow - once it reaches the content, the cursor will jump to the right side of it and start moving in the left direction.


https://github.com/user-attachments/assets/6888a210-a48f-4d90-8376-2504146e78b5



Unfortunately, forcing direction does break stuff in RTL/BiDi texts - if you try searching for the following string in a recording with the page I linked in my previous comment, it won't find any matches:
> كلمة «بطاطا» مأخوذة من الكلمة الإسبانية patata. وأصل الكلمة الإسبانية يعود إلى اللغات المحلية في البحر

I have some experience working on RTL/BiDi support, so I'll try to think of an alternative

going-confetti (approved):
LGTM 🚀

PR #412, opened 2025-01-15
feat: Record, display and generate code for page navigation events 

allansson:
## Description

This PR adds recording of page navigation events, i.e. entering an address in the address bar, using the browser history and reloading a page. The events are displayed in a simple manner in the browser event log of the recording and exporting a browser script will emit the appropriate function calls.

> ### Note: for some reason running the script using the validator doesn't work. Copying the script and running it via the CLI works though. I'm wondering maybe we're doing some trickery to the script before running the validator? I will take a look at it in another PR.

## How to Test

1. Start a recording.
2. Navigate to different places by entering an address in the address bar (also works when selecting a suggestion).
3. Click links and use history to navigate back and forward.
4. Reload the page.
5. Stop the recording.
6. In the top-right dotted menu of the recording preview, click "Export browser script".

You should have a list of page navigation and reload events and the generated script should reflect your recorded session.

## Checklist

Recorded events:

image

After generating a script:

image

## Related PR(s)/Issue(s)

https://github.com/grafana/k6-cloud/issues/3088


allansson (commented):
Self-review

src/codegen/browser/code/context.ts
```diff
@@ -0,0 +1,77 @@
+export interface Import {
```
going-confetti, line 1: What do you think about `src/schemas/imports.ts` and the respective types? It was meant to be a foundation for the custom imports feature, but since that one is postponed, we can easily introduce changes to it, since it's only used in a few places.

going-confetti (approved):
LGTM! Added a non-blocking question

src/codegen/browser/code/context.ts
allansson: Those types do seem identical, but I'm going to keep it as-is for the time being. I think there's a value in keeping the codegen logic completely separate, e.g. a change in the schema won't ripple out into a bunch of changes in the codegen.

Issue #317, opened 2024-11-10
Move `recording selector` position in the generator


Llandy3d:
We should move the position of the "recording selector" living under the `recording` tab of the generator to the top of view near the `save generator` button due to the importance of that item and the fact that it affects everything in the generator.

![Image](https://github.com/user-attachments/assets/5acdf4e1-e40b-4b68-a21c-07151dfba43d)

Llandy3d:
Also consider feedback from maja for this changes if we need more UI changes

PR #413, opened 2025-01-15
feat: move recording selector position

cristianoventura:
## Description

This PR moves the Recording dropdown from the "Requests" tab in the Preview section to the top bar on the Generator page.

## Checklist

image

Smaller window size:

image

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/317

### DIFF
--- a/src/views/Generator/GeneratorControls/GeneratorControls.tsx
+++ b/src/views/Generator/GeneratorControls/GeneratorControls.tsx
@@ -40,7 +40,7 @@ export function GeneratorControls({ onSave, isDirty }: GeneratorControlsProps) {
   return (
     <>
       <RecordingSelector />
-      <Flex align="center" justify="between" gap="2">
+      <Flex align="center" justify="between" gap="2" ml="2">
         <ButtonWithTooltip
           onClick={onSave}
           disabled={!isDirty}
--- a/src/views/Generator/GeneratorControls/GeneratorControls.tsx
+++ b/src/views/Generator/GeneratorControls/GeneratorControls.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react'
-import { DropdownMenu, IconButton } from '@radix-ui/themes'
+import { DropdownMenu, Flex, IconButton } from '@radix-ui/themes'
 
 import { useScriptPreview } from '@/hooks/useScriptPreview'
 import { exportScript } from '../Generator.utils'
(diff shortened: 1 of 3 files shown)


src/views/Generator/GeneratorControls/GeneratorControls.tsx
```diff
@@ -38,52 +39,55 @@ export function GeneratorControls({ onSave, isDirty }: GeneratorControlsProps) {
...
-          />
-          <ExportScriptDialog
-            onExport={exportScript}
-            open={isExportScriptDialogOpen}
-            onOpenChange={setIsExportScriptDialogOpen}
-          />
-        </>
-      )}
+      <RecordingSelector />
+      <Flex align="center" justify="between" gap="2">
```
cristianoventura, line 43: The only change here was wrapping the content in this `Flex` component

src/views/Generator/RecordingSelector.tsx
```diff
@@ -65,7 +65,10 @@ export function RecordingSelector() {
             id="recording-selector"
             placeholder="Select recording"
             css={css`
-              width: 250px;
+              width: 200px;
```
going-confetti, line 68: Now that we have more horizontal space, I'd expect it to be wider rather than narrower than before, because with the addition of the starting URL in recording names, the names became even longer.

src/views/Generator/GeneratorControls/GeneratorControls.tsx
```diff
@@ -38,52 +39,55 @@ export function GeneratorControls({ onSave, isDirty }: GeneratorControlsProps) {
...
-            onOpenChange={setIsValidatorDialogOpen}
-          />
-          <ExportScriptDialog
-            onExport={exportScript}
-            open={isExportScriptDialogOpen}
-            onOpenChange={setIsExportScriptDialogOpen}
-          />
-        </>
-      )}
+      <RecordingSelector />
```
going-confetti, line 42: I think the `+` button might be a bit confusing with this new placement: it's not immediately clear if it's a part of the recording selected, the save generator button, or something else entirely. Perhaps we can add a gap/margin so that the hierarchy is clearer?
image

going-confetti (commented):
Code looks good, added a few questions/suggestions that I'd like to hear your thoughts on

src/views/Generator/RecordingSelector.tsx
cristianoventura: I was mainly trying to keep things concise on the right side of the top bar, but given that names now can be longer, it makes sense to make it wider.

src/views/Generator/GeneratorControls/GeneratorControls.tsx
cristianoventura: I considered using a vertical divider as an alternative, but I kept the extra margin so it's easier to manage in smaller resolutions.

e-fisher (approved):
LGTM! 🚀

PR #417, opened 2025-01-16
chore: Remove yarn usage from pre-commit hook

going-confetti:
## Description
We use `npm` in k6 studio, but the pre-commit hook still references `yarn`, which could lead to not being able to create commits if you don't have `yarn` installed

## How to Test
* Checkout the branch and make sure you can add commits to it and the pre-commit hook runs (adding an obvious eslint issue should be an easy way to do it)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-yarn lint-staged
+npx lint-staged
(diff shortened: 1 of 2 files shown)


cristianoventura (approved):
LGTM!

PR #422, opened 2025-01-17
chore: Add basic feature flags support

going-confetti:
## How to Test

* Enter the "konami code" (`up-up-down-down-left-right-left-right-b-a`)
* Toggle the `data-files` feature
* Check that the `Data files` section is displayed in the sidebar
image

## Checklist

### DIFF
--- a/src/components/DevToolsDialog.tsx
+++ b/src/components/DevToolsDialog.tsx
@@ -5,7 +5,7 @@ import { useState } from 'react'
 import { useFeaturesStore } from '@/store/features'
 import { Feature } from '@/types/features'
 
-export function DevToolsDialog() {
+export function DevModeDialog() {
   const [isOpen, setIsOpen] = useState(false)
 
   useKonami({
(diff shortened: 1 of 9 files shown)


src/components/DevToolsDialog.tsx
```diff
@@ -0,0 +1,46 @@
+import { Button, Dialog, Flex, Switch, Text } from '@radix-ui/themes'
+import useKonami from 'use-konami'
+import { useState } from 'react'
+
+import { useFeaturesStore } from '@/store/features'
+import { Feature } from '@/types/features'
+
+export function DevToolsDialog() {
```
allansson, line 8: If we want to leave this out in production builds, then we could do this:

```tsx
function DevModeDialog() {
  // Full implementation
}

function ProdModeDialog() {
  return null
}

export const DevToolsDialog = 
  import.meta.env.MODE === "development" 
    ? DevModeDialog 
    : ProdModeDialog
```

Vite will replace `import.meta.env.MODE` with the static value so that e.g. a production build will have `"production" === "development"` . Then tree shaking recognizes that `"production" === "development"` will always be false, removes the truthy branch and, since it's no longer referenced anywhere, the `DevModeDialog` as well.

allansson (approved):
Tried it out and it works like a charm. Left a comment on the code that might be worth considering.

src/components/DevToolsDialog.tsx
going-confetti: Good suggestion, thanks!

e-fisher (approved):
LGTM 🚀

cristianoventura (approved):
Looks great! 🤩

PR #427, opened 2025-01-22
internal(Data files): UI for importing data files

going-confetti:
## How to Test
* Use konami code to enable the data-files feature
* Click on `+` in the data files tree in the sidebar
* Select a JSON/CSV file (ideally, try both)
* Click on the file in the sidebar
* Check its preview

I used is [this JSON file](https://github.com/fanzeyi/pokemon.json/blob/master/pokedex.json), it has some nesting so it's good to test how nested properties are displayed 

image


src/constants/files.ts
```diff
@@ -1,2 +1,5 @@
 // eslint-disable-next-line no-control-regex
 export const INVALID_FILENAME_CHARS = /[<>:"/\\|?*\x00-\x1F]/
+
+// Limit the file size for preview to avoid performance issues
+export const FILE_SIZE_PREVIEW_THRESHOLD = 1024 * 1024 * 10
```
going-confetti, line 5: Could probably be higher, I'll revisit this before opening the PR for review.
We don't want to load large files into the main process memory, so similarly to how GitHub doesn't show previews of large files, we'll display a message if the file size is above this threshold

src/components/Layout/Sidebar/Sidebar.tsx
```diff
@@ -106,6 +110,19 @@ export function Sidebar({ isExpanded, onCollapseSidebar }: SidebarProps) {
...
+                actions={
+                  <Tooltip content="Import data file" side="right">
+                    <IconButton
+                      asChild
+                      aria-label="Import data file"
+                      variant="ghost"
+                      size="1"
+                      onClick={handleImportDataFile}
+                    >
+                      <PlusIcon />
```
going-confetti, line 122: I don't really like the plus icon here, because it's used to create resources in other places (recorder/generator file trees). If you a radix icon that works better, please suggest it

src/main.ts
```diff
@@ -632,6 +637,51 @@ ipcMain.handle('data-file:import', async (event) => {
...
+      invariant(
+        fileType === 'csv' || fileType === 'json',
+        'Unsupported file type'
+      )
+
+      const { size } = await stat(filePath)
+
+      if (size > FILE_SIZE_PREVIEW_THRESHOLD) {
+        return null
+      }
```
e-fisher, line 656: Show an error in the UI indicating file is too large to preview, instead of just "no preview available"?

src/components/Layout/Sidebar/Sidebar.tsx
e-fisher: Maybe `download` icon? 
screencapture 2025-01-24 at 11 03 02

If you do a google search for "import icon" most of them consist of an arrow pointed into a rectangle

src/constants/files.ts
e-fisher: I'd say github case is a bit different, it doesn't show preview to prevent overloading dom and slowing down the browser, while in our case we only show the first 10 rows. We should probably do some testing with large files and see how it affects electron's memory consumption, verify there's no memory leak, and that it gets cleaned up properly. Not part of this PR though 👍

e-fisher (commented):
I think this works great! Had a few minor comments:

1. Empty lines in CSV are parsed as empty rows:
screencapture 2025-01-24 at 11 06 49

2. If you import a CSV and JSON file with the same name, there's no indication which is which in the sidebar. Perhaps we could include file extension or add a badge, like the one you have in the preview title.
screencapture 2025-01-24 at 11 07 04

3. Not a `chore` 😅

going-confetti:
> I think this works great! Had a few minor comments:
> 
>     1. Empty lines in CSV are parsed as empty rows:
Good catch, I'll need to verify how it works in the k6 csv module and match the behavior


>     2. If you import a CSV and JSON file with the same name, there's no indication which is which in the sidebar. Perhaps we could include file extension or add a badge, like the one you have in the preview title.
Good point, I'll check what works/looks better 👍 


>     3. Not a `chore` 😅

Fair enough, it's just that it's currently also not a user-facing feature. Should we add the `wip:` prefix to the list? 🤔

src/components/Layout/Sidebar/Sidebar.tsx
going-confetti: Maybe the `upload` one then?
image

e-fisher:
> > ```
> > 3. Not a `chore` 😅
> > ```
> 
> Fair enough, it's just that it's currently also not a user-facing feature. Should we add the `wip:` prefix to the list? 🤔

Either that or `internal` 🤷

src/components/Layout/Sidebar/Sidebar.tsx
e-fisher: This feels more like "share" action to me, but I don't have strong opinion 👍

Llandy3d: what about the `file plus` ?
image

src/main.ts
Llandy3d: or warning 🤔

src/components/Layout/Sidebar/Sidebar.tsx
going-confetti: I think I like this suggestion the most

src/main.ts
going-confetti: Added a warning/info message. Once the layout update PR is merged, I'll probably also add a grot illustration :D

src/utils/dataFile.ts
```diff
@@ -0,0 +1,29 @@
...
+        return parsedContent as DataRecord[]
+      }
+
+      return [parsedContent] as DataRecord[]
+    }
+    case 'csv': {
+      return parseCSV<DataRecord>(content, {
+        header: true,
+        delimiter: ',',
+        skipEmptyLines: true,
```
going-confetti, line 23: I checked `k6/experimental/csv` and it does skip empty lines 👍

going-confetti:
> 2. If you import a CSV and JSON file with the same name, there's no indication which is which in the sidebar. Perhaps we could include file extension or add a badge, like the one you have in the preview title.

I will address this in a separate PR

Llandy3d (approved):
🚀

PR #428, opened 2025-01-22
feat: Update generator layout and improve empty state screens

e-fisher:
## Description

This PR introduces the following changes to the Generator view to improve user experience:

1. Give request list more space, move rule list to the bottom third of the view. The rule list doesn't need much space and most of the time spent will be in request log - with this layout change it's easier to inspect requests and search data required for rules. 

screencapture 2025-01-22 at 16 35 17

2. Show empty state messages on newly created recordings. Currently, if you create a new generator and skip the selection of allowed hosts, you'll end up with this view:

screencapture 2025-01-22 at 16 43 12
- Empty states messages are not displayed
- There's guidance on next action user should take

With changes introduced in this PR:
screencapture 2025-01-22 at 16 45 55

Now the user is guided to 1) select hosts, 2) add a rule

3. Add new Grot illustrations for empty request list and rule state. Previously used 404 illustration gave an impression something was wrong when creating a new generator.

With this being a rather impactful change, feedback and comments are always welcome! 

## Future improvements

Next iteration would be to improve on rule results preview. Editing a rule switches request tab to result preview, which feels counterintuitive to me. We might explore combining request list and preview into one view.

## How to Test

Verify everything still works as expected, try creating generators, editing rules and validating scripts. 

## Checklist


going-confetti (commented):
Looks good for the most part. My biggest concern is that request inspector becomes less usable:
image
The screenshot above is from my 4k monitor with 2x UI scaling (so, there's "only" 1080px worth of vertical space, but then you need to subtract the heights of the dock and the status bar)

cristianoventura (commented):
The addition of the empty state messages looks great! 🤩

I noticed that the visualization of the "Content" tab is a bit cramped when compared to before.

image

I also noticed that the vertical scroll in the Request section doesn't reach the bottom. It happens for all three tabs (Headers, Content and, Cookies)

https://github.com/user-attachments/assets/2d260041-acfc-4ca4-a4b4-2f74a24c7080

Perhaps the "Test rules" section could be contained in the middle of the section and leave the right side for the Request/Response inspector only?

going-confetti:
> I also noticed that the vertical scroll in the Request section doesn't reach the bottom. It happens for all three tabs (Headers, Content and, Cookies)

This should be fixed in my PR from yesterday

going-confetti:
> Perhaps the "Test rules" section could be contained in the middle of the screen and leave the right side for the Request/Response inspector only?

I think this is a good suggestion, we'd just need to decide where to put the rule editor. Perhaps, that's a good time consider opening it in place of the rule list? 🤔

cristianoventura (dismissed):
> I think there's room for improvement, but probably better than what we have now 🤔 The automatic switch to rule preview feels intrusive, but that should be solved when/if we merge request list and preview into a single view.

I agree - For now it looks a lot better even on a smaller screen! 🎸

going-confetti (changes requested):
I updated the branch and wanted to extract the updated EmptyMessage component into its own PR when I noticed that the some of the illustrations aren't consistent (e.g. the stroke width and color of the visor/glasses):
image

src/views/Generator/GeneratorTabs/RequestList.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <ScrollArea scrollbars="vertical">
+        {isRecordingMissing && (
+          <Flex justify="center" align="center" height="100%">
+            <EmptyMessage
+              message="The selected recording is missing, select another one from the top menu"
+              pt="0"
+            />
+          </Flex>
+        )}
+        {!areHostsSelected && (
```
going-confetti, line 66: There are at least 5 different cases when we might want to show a message like this, maybe we can address some of them with the same message if it's generic enough: 
1. No recording is selected (if the user creates an empty generator)
2. Selected recording missing
3. Selected recording is empty (an edge case, but still)
4. Allowlist is empty
5. Allowlist is not empty, but the selected hosts are only used for static assets (see example below)

image

e-fisher:
We'll need custom illustrations created for us to match the existing style, I've reverted to default Grot and will add new illustrations once we have them. 

I've also addressed cases you've mentioned:

screencapture 2025-02-10 at 12 13 28

screencapture 2025-02-10 at 12 13 45

screencapture 2025-02-10 at 12 13 59

screencapture 2025-02-10 at 12 14 45

The recording related messages now also include the recording dropdown instead of directing the user to the top menu, similarly like hosts related messages, let me know what you think!

going-confetti (approved):
LGTM!

going-confetti (approved):

PR #430, opened 2025-01-23
fix: Content preview layout is broken

going-confetti:
## Description

Apart from the scrollarea fix and padding reductions, it also fixes an issue described in the comment

## Screenshots
Before
image

After
image

### DIFF
--- a/src/components/WebLogView/ResponseDetails/Content.tsx
+++ b/src/components/WebLogView/ResponseDetails/Content.tsx
@@ -24,6 +24,10 @@ export function Content({ data }: { data: ProxyData }) {
     return reset
   }, [reset])
 
+  useEffect(() => {
+    setSelectedTab('content')
+  }, [format])
+
   if (!contentType || !content || !format) {
     return (
       <Flex height="200px" justify="center" align="center">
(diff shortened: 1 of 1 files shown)


src/components/WebLogView/ResponseDetails/Content.tsx
```diff
@@ -24,6 +24,10 @@ export function Content({ data }: { data: ProxyData }) {
     return reset
   }, [reset])
 
+  useEffect(() => {
+    setSelectedTab('content')
```
going-confetti, line 28: There's a bug in `main`:
* Open a non-media response
* Switch to `Raw`
* Open a media response - you'll see its raw representation with no way to switch to a normal preview

One can argue that keeping the selected tab state is user friendly - in that case, we can remove that useEffect, it should still work with the changes on line 45

cristianoventura: I think it's okay for now and we can change if we receive feedback about the usability.

cristianoventura (approved):
LGTM!

Llandy3d (approved):
🚀

PR #434, opened 2025-01-24
feat: Execute browser scripts in the validator.

allansson:
## Description

This PR modifies the way script are run in the validator so that browser tests can be run.

The previous approach didn't work because using `--iterations` and `--vus` would overwrite the scenario configs, which also removed the required browser options.

Instead of string manipulations, the new approach parses the script and modifies its AST in various ways. To solve the problem of running browsers tests, we now insert a custom options export with a configuration equivalent to the command line args:

```ts
export const options = {
  scenarios: {
    default: {
      executor: 'shared-iterations',
      iterations: 1,
      vus: 1,
      // Browser options, if 'k6/browser' is imported
      options: {
        browser: {
          type: "chromium"
        }
     }
    }
  }
}
```

## How to Test

1. Record a browser test
3. Export a script
4. Run the script in the validator
5. Generate a protocol test
6. Run the validator

You should be able to run a browser test, i.e. there shouldn't be a message in the console saying that the required scenario options are missing. Everything should work like it did before when running a protocol test.

## Checklist

What happened before this PR:

image

After this PR:

image

### DIFF
--- a/src/script.ts
+++ b/src/script.ts
@@ -267,12 +267,13 @@ export const enhanceScript = async ({
   )
 
   return format(script, {
-    parser: 'ts',
+    parser: 'k6',
     plugins: [
       estree,
+      // This is a custom parser plugin that simply returns our modified AST.
       {
         parsers: {
-          ts: {
+          k6: {
             astFormat: 'estree',
             parse: () => {
               return scriptAst


src/__snapshots__/script/browser-options/with-browser-import.js
allansson, line 1: Not sure if we want these snapshots stored here or if we want to gather them somewhere in the project root (not even sure if that's possible).

src/codegen/browser/tstree.ts
allansson, line 1: I previously had copied the type definitions because I had some problems with declaration merging. Turns out I was just stupid and now it works, so back to the official types it is.

src/codegen/estree/declarations.ts
allansson, line 1: Since some of these helpers are used to "enhance scripts" I've moved them into a more generic location.

src/codegen/estree/traverse.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        p: 7
+      }
+
+      export {
+        q,
+        r,
+        s
+      }
+
+      export * from 'module'
```
allansson, line 37: Yup, they are all valid ways of exporting stuff.

src/codegen/estree/traverse.ts
```diff
@@ -0,0 +1,1524 @@
...
+import { exhaustive } from '@/utils/typescript'
+
+type Visitor = {
+  [P in ts.Node['type']]?: (node: Extract<ts.Node, { type: P }>) => void
+} & {
+  enter?: (node: ts.Node) => void
+  leave?: (node: ts.Node) => void
+}
+
+export function traverse(node: ts.Node, visitor: Visitor) {
```
allansson, line 12: This was a pain to write. It visits every possible type of node in typescript. All 168 of them.

src/script.ts
```diff
@@ -110,45 +122,190 @@ export const runScript = async (
...
+    throw new Error('Failed to parse group snippet')
+  }
+
+  if (checksAst === undefined) {
+    throw new Error('Failed to parse checks snippet')
+  }
+
+  if (scriptAst === undefined) {
+    throw new Error('Failed to parse script content')
+  }
```
allansson, line 178: Not sure what we should do if the user imports an invalid script file. Right now I just throw, but maybe I should return the unaltered "script" and let k6 fail when trying to parse it?

src/script.ts
```diff
@@ -110,45 +122,190 @@ export const runScript = async (
...
+      declaration: declareConst({
+        declarations: [
+          constDeclarator({
+            id: identifier('options'),
+            init: options,
+          }),
+        ],
+      }),
+    })
+  )
```
allansson, line 267: A more sophisticated version would try to merge the injected options with existing options, but I think this a good enough start.

allansson (commented):
Self-review

going-confetti (approved):
LGTM 👏

PR #439, opened 2025-01-28
feat: Record, display and generate code for click events

allansson:
## Description

This PR adds recording of click events to the browser recorder. It currently supports left-click (although it has the plumbing to do right and middle clicks) with any modifier key.

## How to Test

1. Start a recording.
2. Click around on elements with and without modifier keys (option, shift, cmd, etc.)
3. View the browser event log
4. Export a script
5. Run the script using the validator

The browser event log should contain the click events and the exported script should contain the clicks and be runnable.

## Checklist


src/codegen/estree/nodes.ts
```diff
@@ -3,8 +3,11 @@ import { TSESTree as ts } from '@typescript-eslint/types'
 // Since we're generating our own AST we don't have any positional information, but
 // the types require it. To fix it, we spread this dummy object into the nodes we create.
 export const baseProps = {
-  loc: null as unknown as ts.SourceLocation,
-  range: null as unknown as ts.Range,
+  loc: {
+    start: { line: 0, column: 0 },
+    end: { line: 0, column: 0 },
+  },
+  range: [0, 1] as [number, number],
```
allansson, line 10: Prettier would go into an infinite loop when emitting arrays without some kind of location information. Luckily, it doesn't care that everything is seemingly contained in a single character in the doc.

extension/src/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { finder } from '@medv/finder'
```
allansson, line 1: Not sure I'm happy with the selectors this generates. They seem very brittle. I might try some alternative in the future.

Llandy3d (approved):
Reviewing from the functionality standpoint, I correctly managed to record click events and run it 🙌 

I would wait for another review on the code as well 🙇

src/views/Recorder/BrowserEventLog.tsx
```diff
@@ -16,31 +43,91 @@ function EventIcon({ event }: EventIconProps) {
...
+
+interface ClickDescriptionProps {
+  event: ClickEvent
+}
+
+function ClickDescription({ event }: ClickDescriptionProps) {
+  const modifiers = getModifierKeys(event.modifiers)
+  const button = getButtonDescription(event)
+
+  const clickedText = modifiers.concat(button).join(' + ')
```
going-confetti, line 107: Might be a good place for the [Kbd](https://www.radix-ui.com/themes/docs/components/kbd) component (font size consistency should probably be improved tho):

image

extension/src/recorder.ts
```diff
@@ -0,0 +1,61 @@
+import { BrowserEvent } from '@/schemas/recording'
+import { runtime } from 'webextension-polyfill'
+import { generateSelector } from './selectors'
+
+// Technically we can't rely on the button index because a user could re-assign
+// the meaning of the button but it's a fringe case.
```
going-confetti, line 6: It's probably pretty common among left-handed users, so I'm not sure it's a fringe case. According to MDN it doesn't change the meaning on the button though, so we should be safe:
> Users may change the configuration of buttons on their pointing device so that if an event's button property is zero, it may not have been caused by the button that is physically left–most on the pointing device; _however, it should behave as if the left button was clicked in the standard button layout._

going-confetti (approved):
LGTM 🚀 ! Really excited about using ASTs for protocol-level tests in future

src/views/Recorder/BrowserEventLog.tsx
allansson: Awesome! I did not know that was a thing. Thank you! 🙏

PR #441, opened 2025-01-29
internal(Data files): Select data files in Test options

going-confetti:
## Description
WIP UI for selecting data files in Generator's test options
image

## How to Test
* Enable the `data-files` feature flag
* Open existing generator and try editing variables and files - make sure the changes are saved
* ⚠️ Verify that Variables editor works as expected even without the feature flag enabled
* ⚠️ Verify that existing generators can be opened
* ⚠️ Verify that new generators can be created


src/views/Generator/TestOptions/DataFiles.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <div>
+      <Text size="2" as="p" mb="2">
+        Make your test more realistic and prevent server-side caching from
+        affecting your results by using data files.
+      </Text>
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <FieldGroup name="files" label="Files" errors={errors}>
+          <CheckboxGroup.Root
```
going-confetti, line 68: This needs some work before the feature is made public. Considering the user will be able to configure access method, maybe using a data table similar to the variables editor will be better 🤔

src/views/Generator/TestOptions/DataFiles.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const handleChange = (value: string[]) => {
+    setValue(
+      'files',
+      value.map((fileName) => ({
+        name: fileName,
+      }))
+    )
+  }
+
+  const filesField = register('files')
```
going-confetti, line 58: I probably don't know how to cook `@hookform`, but this feels excessive for what I'm trying to achieve here 😞

cristianoventura: A controlled `CheckboxGroup` could also work here and we may be able to remove the `watch` and `register` calls.

```tsx
<Controller
   name="files"
   control={control}
   render={({ field }) => (
       <CheckboxGroup.Root
         value={field.value}
         onValueChange={field.onChange}
        >
           {options.map((option) => .... }
      </CheckboxGroup.Root>
   )}
/>
```

cristianoventura (commented):
I wanted to confirm if this is the expected behaviour or a bug

- choose a data file from the checkbox group
- save the generator
- open a different generator
- the data file for the previous generator was still checked

This caused the `isDirty` state of all generators to be `true`, which triggered the prompt to save it even when no changes were made.

https://github.com/user-attachments/assets/071a489e-3331-487b-89a6-3c07d1895c65

going-confetti:
> I wanted to confirm if this is the expected behaviour or a bug
> 
>     * choose a data file from the checkbox group
> 
>     * save the generator
> 
>     * open a different generator
> 
>     * the data file for the previous generator was still checked
> 

That's definitely a bug, thanks. Turns out I forgot to update `setGeneratorFile` and I'm actually thinking whether we should also reset the state when the user navigates away from a generator

src/views/Generator/TestOptions/DataFiles.tsx
going-confetti: So, I tried something like this, but the problem I ran into was that that field has type `TestData`, but `CheckboxGroup` value is `string[]`. Maybe I can resolve that on submit?

Llandy3d (approved):
Reviewing from the functionality standpoint, everything seems to still be working correctly 🙌

src/views/Generator/TestOptions/DataFiles.tsx
cristianoventura: Ah I see, yes in that case we may have to transform them `onSubmit` before saving it to file.

cristianoventura (approved):
LGTM! I'll leave it up to you to decide whether any changes are needed for the `Checkbox.Root` component.

going-confetti:
> LGTM! I'll leave it up to you to decide whether any changes are needed for the Checkbox.Root component.

Thanks! I will probably change the UI to be similar to Thresholds and Variables: there will be a table with a file select in the first column and access method configuration in the second column 🤔 
I should definitely spend some time learning how to use hookform :D

PR #445, opened 2025-01-31
internal(Data files): Generate code for opening data files

going-confetti:
## Description
Few things to keep in mind:
* It uses the global `open()` instead of the new asynchronous one from `k6/experimental/fs`
* It uses `papaparse` from k6 jslibs instead, but the plan is to switch to `k6/experimental/csv` soon

image
image

## How to Test
* Enable the `data-files` toggle
* Import a json file and a csv file
* In Test options, select both files
* Create a custom code rule/custom code rules similar to the example above
* Run validate from Generator
* Export the script
* Run it in Validator


src/codegen/codegen.ts
```diff
@@ -22,30 +23,72 @@ export function generateScript({
...
       ${generateVUCode(recording, generator.rules, generator.options.thinkTime)}
     }
   `
 }
 
+export function generateImports(generator: GeneratorFileData): string {
+  const hasDataFiles = generator.testData.files.length > 1
+  const hasCsvDataFiles = generator.testData.files.some(({ name }) =>
+    name.toLowerCase().endsWith('csv')
+  )
```
going-confetti, line 44: I'm not sure if  inferring the file type from the filename is a good solution. We can probably store the type along the name. Although it won't change the fact the type is inferred in a similar way elsewhere

src/script.ts
```diff
@@ -66,17 +69,16 @@ export const runScript = async (
...
     '--log-format=json',
     '--quiet',
+    ...(!enableUsageReport ? ['--no-usage-report'] : []),
   ]
 
-  if (!enableUsageReport) {
-    k6Args.push('--no-usage-report')
-  }
+  // Create symlink to data folder so it can be accessed by the script
+  await symlink(DATA_FILES_PATH, path.join(app.getPath('temp'), 'Data'), 'dir')
```
going-confetti: Alternatively, we can create it at startup and remove at shutdown 🤔

src/script.ts
```diff
@@ -66,17 +69,16 @@ export const runScript = async (
...
     '--log-format=json',
     '--quiet',
+    ...(!enableUsageReport ? ['--no-usage-report'] : []),
   ]
 
-  if (!enableUsageReport) {
-    k6Args.push('--no-usage-report')
-  }
+  // Create symlink to data folder so it can be accessed by the script
+  await symlink(DATA_FILES_PATH, DATA_FILES_TEMP_PATH, 'dir')
```
going-confetti, line 76: Alternatively, we can create it at startup and remove at shutdown 🤔

Llandy3d (dismissed):
🚀 

One only nit is that I'm not sure how safe is to expect to have `Scripts` and `Data` folders just for us in `/tmp`.
Maybe we should add another folder to avoid collision with other apps, but for now is fine 🙇

going-confetti:
> One only nit is that I'm not sure how safe is to expect to have `Scripts` and `Data` folders just for us in `/tmp`. Maybe we should add another folder to avoid collision with other apps, but for now is fine 🙇

Thanks for pointing this out, that seems to be a pretty big deal actually. I'll update the code

Llandy3d (approved):
LGTM!

PR #452, opened 2025-02-05
fix: Generated path is incorrect when saving script on Windows

going-confetti:


src/main.ts
```diff
@@ -406,11 +406,9 @@ ipcMain.handle('script:run-from-generator', async (event, script: string) => {
 ipcMain.handle(
   'script:save',
   async (event, script: string, fileName: string = 'script.js') => {
-    console.info('script:save event received')
-
     const browserWindow = browserWindowFromEvent(event)
     try {
-      const filePath = `${SCRIPTS_PATH}/${fileName}`
```
going-confetti, line 413: 😨

Llandy3d (approved):

PR #453, opened 2025-02-05
feat: Simpler default file names

going-confetti:
## Description

- Make auto-generated file names shorter and reduce noise, by removing the timestamp and using optional increment at the end.
- Remove `GeneratorFile` as it's become redundant
- Stringify HAR/Generator data in the main process
image

## How to Test
Verify that the following works as before (apart from the new filenames):
- Create an new generator, then do it again on the same date
- Create a few new recordings with the same starting URL
- Create a new recording without a starting URL
- Try re-configuring a generator and saving changes

### DIFF
--- a/src/hooks/useCreateGenerator.ts
+++ b/src/hooks/useCreateGenerator.ts
@@ -1,8 +1,6 @@
 import { useCallback } from 'react'
 import { useNavigate } from 'react-router-dom'
 
-import { generateFileNameWithTimestamp } from '@/utils/file'
-import { createNewGeneratorFile } from '@/utils/generator'
 import { getRoutePath } from '@/routeMap'
 import { useToast } from '@/store/ui/useToast'
 import log from 'electron-log/renderer'
@@ -13,11 +11,7 @@ export function useCreateGenerator() {
 
   const createTestGenerator = useCallback(async () => {
     try {
-      const newGenerator = createNewGeneratorFile()
-      const fileName = await window.studio.generator.saveGenerator(
-        JSON.stringify(newGenerator, null, 2),
-        generateFileNameWithTimestamp('json', 'Generator')
-      )
+      const fileName = await window.studio.generator.createGenerator()
 
       navigate(
         getRoutePath('generator', { fileName: encodeURIComponent(fileName) })
--- a/src/hooks/useCreateGenerator.ts
+++ b/src/hooks/useCreateGenerator.ts
@@ -9,21 +9,25 @@ export function useCreateGenerator() {
   const navigate = useNavigate()
   const showToast = useToast()
 
-  const createTestGenerator = useCallback(async () => {
-    try {
-      const fileName = await window.studio.generator.createGenerator()
+  const createTestGenerator = useCallback(
+    async (recordingPath = '') => {
+      try {
+        const fileName =
+          await window.studio.generator.createGenerator(recordingPath)
 
-      navigate(
-        getRoutePath('generator', { fileName: encodeURIComponent(fileName) })
-      )
-    } catch (error) {
-      showToast({
-        status: 'error',
-        title: 'Failed to create generator',
-      })
-      log.error(error)
-    }
-  }, [navigate, showToast])
+        navigate(
+          getRoutePath('generator', { fileName: encodeURIComponent(fileName) })
+        )
+      } catch (error) {
+        showToast({
+          status: 'error',
+          title: 'Failed to create generator',
+        })
+        log.error(error)
+      }
+    },
+    [navigate, showToast]
+  )
 
   return createTestGenerator
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -66,7 +66,7 @@ import {
 import { ProxyStatus, StudioFile } from './types'
 import { configureApplicationMenu } from './menu'
 import * as Sentry from '@sentry/electron/main'
-import { exhaustive } from './utils/typescript'
+import { exhaustive, isNodeJsErrnoException } from './utils/typescript'
 import { DataFilePreview } from './types/testData'
 import { parseDataFile } from './utils/dataFile'
 import { createNewGeneratorFile } from './utils/generator'
@@ -430,16 +430,13 @@ ipcMain.handle(
 ipcMain.handle(
   'har:save',
   async (_, data: HarWithOptionalResponse, prefix: string) => {
-    const fileName = await generateUniqueFileName({
+    const fileName = await createFileWithUniqueName({
+      data: JSON.stringify(data, null, 2),
       directory: RECORDINGS_PATH,
       ext: '.har',
       prefix,
     })
 
-    await writeFile(
-      path.join(RECORDINGS_PATH, fileName),
-      JSON.stringify(data, null, 4)
-    )
     return fileName
   }
 )
@@ -488,16 +485,12 @@ ipcMain.handle('har:import', async (event) => {
 // Generator
 ipcMain.handle('generator:create', async (_, recordingPath: string) => {
   const generator = createNewGeneratorFile(recordingPath)
-  const fileName = await generateUniqueFileName({
+  const fileName = await createFileWithUniqueName({
+    data: JSON.stringify(generator, null, 2),
     directory: GENERATORS_PATH,
     ext: '.json',
     prefix: 'Generator',
   })
-
-  await writeFile(
-    path.join(GENERATORS_PATH, fileName),
-    JSON.stringify(generator, null, 2)
-  )
 
   return fileName
 })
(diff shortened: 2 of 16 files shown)


src/hooks/useCreateGenerator.ts
```diff
@@ -13,11 +11,7 @@ export function useCreateGenerator() {
 
   const createTestGenerator = useCallback(async () => {
     try {
-      const newGenerator = createNewGeneratorFile()
```
going-confetti, line 16: No need to do it in the render process

src/main.ts
```diff
@@ -427,11 +427,22 @@ ipcMain.handle(
...
+  async (_, data: HarWithOptionalResponse, prefix: string) => {
+    const fileName = await generateUniqueFileName({
+      directory: RECORDINGS_PATH,
+      ext: '.har',
+      prefix,
+    })
+
+    await writeFile(
+      path.join(RECORDINGS_PATH, fileName),
+      JSON.stringify(data, null, 4)
```
going-confetti, line 441: HARs and generators are now stringified just before writing to disk. I believe it makes more sense to do it this way (among other things it improves the DX since the param type is not `string` anymore).

Sidenote/question: why is it `4` for HARs and `2` for generators? 🤔

src/main.ts
```diff
@@ -962,3 +985,35 @@ const cleanUpProxies = async () => {
     kill(proc.pid)
   })
 }
+
+const generateUniqueFileName = async ({
```
going-confetti, line 989: This should be in a util file, but I'd need to create a new one, since `utils/file.ts` has some exports that are used in `render` :/

src/utils/rules.ts
```diff
@@ -5,7 +5,7 @@ export function createEmptyRule(type: TestRule['type']): TestRule {
     case 'correlation':
       return {
         type: 'correlation',
-        id: self.crypto.randomUUID(),
+        id: crypto.randomUUID(),
```
going-confetti, line 8: Question: although it's not a big deal, should we stop storing rule ids in the generator file?

src/views/Generator/Generator.utils.ts
```diff
@@ -31,19 +31,9 @@ export async function exportScript(fileName: string) {
...
-    JSON.stringify(generatorData, null, 2),
-    fileName
-  )
-}
-
 export const loadGeneratorFile = async (fileName: string) => {
-  const generatorFile = await window.studio.generator.loadGenerator(fileName)
-  return GeneratorFileDataSchema.parse(generatorFile.content)
+  const generator = await window.studio.generator.loadGenerator(fileName)
+  return GeneratorFileDataSchema.parse(generator)
```
going-confetti, line 36: IMO this should be happening in `main`

src/views/Recorder/Recorder.tsx
```diff
@@ -105,11 +105,8 @@ export function Recorder() {
       })
 
       const har = proxyDataToHar(grouped)
-      const prefix = startUrl && getHostNameFromURL(startUrl)
-      const fileName = await window.studio.har.saveFile(
-        JSON.stringify(har, null, 4),
-        prefix
-      )
+      const prefix = getHostNameFromURL(startUrl) ?? 'Recording'
```
going-confetti, line 108: Here's another change: if no starting URL is provided, the name will follow the `Recording - YYYY-MM-DD` pattern

going-confetti (commented):
Self-review

src/main.ts
cristianoventura: I don’t have full context on the decision, but my guess is that 4 spaces were chosen for improved readability, especially since HAR files can become quite complex and deeply nested. 🤔

src/main.ts
```diff
@@ -962,3 +985,35 @@ const cleanUpProxies = async () => {
...
+  prefix: string
+  ext: string
+}): Promise<string> => {
+  const timestamp = new Date().toISOString().split('T')[0] ?? ''
+  const template = `${prefix ? `${prefix} - ` : ''}${timestamp}${ext}`
+
+  let uniqueFileName = template
+  let fileExists = await access(path.join(directory, uniqueFileName))
+    .then(() => true)
+    .catch(() => false)
```
cristianoventura, line 1004: Is there a reason why we shouldn't use `fs.existsSync` here instead? It'd look semantically more appropriate as it returns a boolean value.

src/utils/rules.ts
cristianoventura: Right now, I believe it's only used by the state manager (so we know each one is selected/being edited etc). It might be easier to have an id on file than setting than during runtime 🤔

cristianoventura (commented):
It works as expected! 🚀 

I left a comment for clarification before approving.

src/main.ts
allansson: Prettier defaults to `2`, so I'd say that's a good thing to go by.

src/main.ts
allansson: This is where having 

```
shared/
  - utils/
    - file.ts
main/
  - index.ts
render/
  - utils/
    - file.ts
  - index.ts
```

This would allows us to setup a `no-restricted-imports` rule in ESLint so that things in `render/` can't import from `main/` and vice versa.

src/main.ts
allansson: I think what you want to do is use `await fs.writeFile(path, data, { flag: 'ax+' })` (see [flags](https://nodejs.org/api/fs.html#file-system-flags)) and repeat that with a new suffix until the write succeeds.

src/utils/rules.ts
allansson: I think we should store them because:

1. They are convenient when writing the frontend, e.g. for `key` props.
2. If you diff two files, you can tell whether it was one thing being changed or one thing being removed and another added.
3. JavaScript in general has no concept of value equality when it comes to complex types, so being able to easily check for equality via an `id` is very handy.

Like @cristianoventura said, setting them during runtime is harder. 

You'd have to traverse the entire hierarchy of objects and slap and id on the things you care about, and if you load the thing from disk again you'll get an entirely new object with new id:s.

For instance, consider checking if a file has unsaved changes. If the id:s aren't stored, we would have to strip all the ids from everything before doing any comparison.

That said, could we switch to `nanoid`? No need to use crypto-grade randomness and `nanoid`s are a bit terser than UUIDs. 😅

src/views/Generator/Generator.hooks.ts
```diff
@@ -41,7 +37,10 @@ export function useUpdateValueInGeneratorFile(fileName: string) {
   return useMutation({
     mutationFn: async ({ key, value }: { key: string; value: unknown }) => {
```
allansson, line 38: Nothing to do with this PR really, but just an idea: instead of passing a key and value to update (which doesn't seem to be type safe?), why not use a `map` function:

```ts
useUpdateInFile(fileName: (generator: Generator) => {
  return {
    ...generator,
    myValue: "hello"
  }
})
```

src/views/Generator/Generator.utils.ts
allansson: Agreed.

going-confetti: I'll open a separate PR for this change

src/main.ts
cristianoventura: IMO it's okay to use these methods in certain places (like a one-off operation, during initialization) that don't compromise the event loop.

@allansson's suggestion seems to handle both the check and the write at the same time so we can consider that approach as well.

If using `fs.access`, we could condering cleaning up the then/catch block if we move it to a helper function since it's been used in a few places.

```ts
const fileExists = await checkFileExsits(path.join(directory, uniqueFileName)
```

cristianoventura (dismissed):

src/utils/rules.ts
going-confetti: >That said, could we switch to nanoid? No need to use crypto-grade randomness and nanoids are a bit terser than UUIDs. 😅

We could, but what would be the benefits of this?

src/main.ts
going-confetti: I rewrote the function to use `writeFile`

src/utils/rules.ts
allansson: > We could, but what would be the benefits of this?

None, really. A bit more efficient (crypto is computationally expensive), saves a few bytes of space and IMO they are easier to read than UUIDs.

going-confetti: I think the last argument is the strongest one, but these ids aren't user-facing so I wouldn't make it a priority.
As for performance, in the benchmarks I found `crypto.randomUUID()` had much higher number of operations per second than `nanoid()`, but it probably doesn't make any measurable difference in the context of k6 studio

src/main.ts
Llandy3d: I don't think we were setting the padding so I'm not sure why they would be different 🤔

Llandy3d (approved):
LGTM! 🙌

Llandy3d (approved):

allansson (approved):
Really like this change. Remove much of the clutter in the sidebar and it overall just feel more organized now.

Great job! 💯

PR #463, opened 2025-02-10
internal: Use tanstack query to handle generator and recording loading

going-confetti:
## Description
* Use tanstack query for loading generator and recordings in all places

## How to Test
⚠️ Please be thorough ⚠️ 
Test that generators and recordings are loading as expected. Verify that the allowlist dialog is opened automatically on recording change when needed.

## Checklist

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,7 +12,6 @@ import {
   writeFile,
   unlink,
   readdir,
-  FileHandle,
   rename,
   access,
   readFile,
@@ -47,7 +46,7 @@ import {
   INVALID_FILENAME_CHARS,
   TEMP_GENERATOR_SCRIPT_FILENAME,
 } from './constants/files'
-import { HarFile, HarWithOptionalResponse } from './types/har'
+import { HarWithOptionalResponse } from './types/har'
 import { GeneratorFileData } from './types/generator'
 import kill from 'tree-kill'
 import find from 'find-process'
@@ -70,6 +69,7 @@ import { exhaustive, isNodeJsErrnoException } from './utils/typescript'
 import { DataFilePreview } from './types/testData'
 import { parseDataFile } from './utils/dataFile'
 import { createNewGeneratorFile } from './utils/generator'
+import { GeneratorFileDataSchema } from './schemas/generator'
 
 if (process.env.NODE_ENV !== 'development') {
   // handle auto updates
@@ -439,23 +439,18 @@ ipcMain.handle(
   }
 )
 
-ipcMain.handle('har:open', async (_, fileName: string): Promise<HarFile> => {
-  console.info('har:open event received')
-  let fileHandle: FileHandle | undefined
-  try {
-    fileHandle = await open(path.join(RECORDINGS_PATH, fileName), 'r')
-    const data = await fileHandle?.readFile({ encoding: 'utf-8' })
+ipcMain.handle(
+  'har:open',
+  async (_, fileName: string): Promise<HarWithOptionalResponse> => {
+    const data = await readFile(path.join(RECORDINGS_PATH, fileName), {
+      encoding: 'utf-8',
+      flag: 'r',
+    })
     // TODO: https://github.com/grafana/k6-studio/issues/277
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    const har = await JSON.parse(data)
-
-    // TODO: https://github.com/grafana/k6-studio/issues/277
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    return { name: fileName, content: har }
-  } finally {
-    await fileHandle?.close()
+    return JSON.parse(data)
   }
-})
+)
 
 ipcMain.handle('har:import', async (event) => {
   console.info('har:import event received')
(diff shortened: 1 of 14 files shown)


e-fisher (commented):
Did some initial testing, there seem to be a weird case when switching between generators that contain different sets of hosts, the allowlist would popup when returning to previously opened generator. I'll dome some more testing to try to find exact reproduction steps (because it doesn't happen with every generator), but adding this recording in the meantime in case you have some ideas:


https://github.com/user-attachments/assets/53038c88-6b64-4387-bb19-ddd0d7ac7865

src/main.ts
```diff
@@ -439,23 +439,18 @@ ipcMain.handle(
   }
 )
 
-ipcMain.handle('har:open', async (_, fileName: string): Promise<HarFile> => {
-  console.info('har:open event received')
```
Llandy3d, line 443: I noticed we have some events have this log and others do not, any specific reasons why we might not want them anymore ? 🤔

src/main.ts
```diff
@@ -439,23 +439,18 @@ ipcMain.handle(
...
+    const data = await readFile(path.join(RECORDINGS_PATH, fileName), {
+      encoding: 'utf-8',
+      flag: 'r',
+    })
     // TODO: https://github.com/grafana/k6-studio/issues/277
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    return { name: fileName, content: har }
-  } finally {
-    await fileHandle?.close()
+    return JSON.parse(data)
```
Llandy3d, line 451: This makes me think about the scenario where the file is too big and electron might just drop the data when trying to send it, silently 😕

src/main.ts
```diff
@@ -508,20 +503,14 @@ ipcMain.handle(
 ipcMain.handle(
   'generator:open',
   async (_, fileName: string): Promise<GeneratorFileData> => {
-    let fileHandle: FileHandle | undefined
-
-    try {
-      fileHandle = await open(path.join(GENERATORS_PATH, fileName), 'r')
+    const data = await readFile(path.join(GENERATORS_PATH, fileName), {
```
Llandy3d, line 506: if we pass a file that doesn't exist it will fail in the schema parsing section right ?
Is the error correct in that case ? 🤔

going-confetti: No, `readFile` should throw an error in this case.
> Is the error correct in that case ? 🤔

Do you mean the notification we show to the user? Currently, it says "Failed to load generator", which would still be more or less correct. It might be a good idea to have a distinction between fs errors/schema errors/anything else

src/main.ts
going-confetti: I should've discussed this with the team first, but I don't think these bring a lot of value, while making the code a bit harder to maintain (it's a bit of s stretch, I know):
- You need to remember to add it when adding a new handler
- You need to remember to update it if you rename the channel

Also these logs aren't written to the app logs, so they are only useful when debugging something locally, in which case wouldn't it be easier to add temporary console.logs in the local branch?

Llandy3d: I agree they are more for debugging rather than info
Depending on what you are doing might be useful to see all the events your operations are doing to track what's happening, if you have to add them back manually might be a slog 🤔 

> You need to remember to update it if you rename the channel
hopefully that's not something that happens often, specially post GA 😅
I also see how they can be error-prone, annoying, the ideal scenario would be for electron to an event when it arrives automatically 🤔

src/main.ts
Llandy3d: I replied to the wrong thing 🤦 
yeah will be worth to explore streaming data in the future as we move more things 👍

src/main.ts
going-confetti: Yes, I wish there was a simple way to add a middleware for logging all this automatically

going-confetti: There's also [this doc](https://www.electronjs.org/docs/latest/tutorial/debugging-main-process) on debugging in the main process

e-fisher:
Attaching 2 generators and their corresponding recordings to help with debugging - switching between those 2 generators (not recordings) opens up the hosts dialog.
[generator-debug.zip](https://github.com/user-attachments/files/18765447/generator-debug.zip)

src/views/Generator/Generator.tsx
```diff
@@ -14,44 +14,38 @@ import { TestRuleContainer } from './TestRuleContainer'
...
+    fileName,
+    onSuccess: setGenerator,
+    onError: () => {
+      showToast({
+        title: 'Failed to load generator',
+        status: 'error',
+      })
+      navigate(getRoutePath('home'), { replace: true })
+    },
+  })
```
e-fisher, line 45: Not that I'm against this approach, I actually liked `onSuccess` and `onError` callbacks, but if we decide to follow this pattern with other hooks it would add duplication of setting up those callbacks or require custom wrappers for tanstack hooks. 

Have you considered using tanstack's proposed [migration guide](https://github.com/TanStack/query/discussions/5279), to use `useEffect` hooks instead?

```suggestion
  const { isLoading, data, isSuccess, isError } = useGeneratorFile({ fileName })

  useEffect(() => {
    if (isSuccess && data) {
      setGenerator(data)
    }
  }, [isSuccess, data, setGenerator])

  useEffect(() => {
    if (isError) {
      showToast({
        title: 'Failed to load generator',
        status: 'error',
      })
      navigate(getRoutePath('home'), { replace: true })
    }
  }, [isError, navigate, showToast])
```

going-confetti:
@CodiumAI-Agent /review

PR #465, opened 2025-02-11
internal: Parse generator files in main

going-confetti:
## How to Test
Verify that the following features work as expected:
* Opening recordings and generators
* Opening and running an external script

### DIFF
--- a/src/codegen/codegen.test.ts
+++ b/src/codegen/codegen.test.ts
@@ -120,6 +120,7 @@ describe('Code generation', () => {
             value: 1,
           },
         },
+        thresholds: [],
       },
       testData: {
         variables: [],
@@ -129,7 +130,6 @@ describe('Code generation', () => {
       allowlist: [],
       includeStaticAssets: false,
       scriptName: 'my-script.js',
-      thresholds: [],
     }
 
     it('should generate imports', async () => {
--- a/src/main.ts
+++ b/src/main.ts
@@ -433,11 +433,8 @@ ipcMain.handle(
       encoding: 'utf-8',
       flag: 'r',
     })
-    // TODO: https://github.com/grafana/k6-studio/issues/277
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    const har = JSON.parse(data)
 
-    return har
+    return JSON.parse(data)
   }
 )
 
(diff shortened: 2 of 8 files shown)


src/codegen/codegen.test.ts
```diff
@@ -129,7 +130,6 @@ describe('Code generation', () => {
       allowlist: [],
       includeStaticAssets: false,
       scriptName: 'my-script.js',
-      thresholds: [],
```
going-confetti, line 132: Sneaky fix for a failing typecheck that prevented me from making a commit. I wonder why it the CI checks don't fail :thinking:

src/main.ts
```diff
@@ -327,31 +326,18 @@ ipcMain.on('browser:stop', async () => {
...
-
-  const fileHandle = await open(scriptPath, 'r')
-  try {
-    const script = await fileHandle?.readFile({ encoding: 'utf-8' })
 
-    return { path: scriptPath, content: script }
-  } finally {
-    await fileHandle?.close()
-  }
+  return scriptPath
```
going-confetti, line 331: We only need the path anyway, because there's a route redirect on selection and the actual loading is handled after the redirection

src/main.ts
```diff
@@ -439,23 +425,21 @@ ipcMain.handle(
...
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    const har = await JSON.parse(data)
+    const har = JSON.parse(data)
 
-    // TODO: https://github.com/grafana/k6-studio/issues/277
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    return { name: fileName, content: har }
-  } finally {
-    await fileHandle?.close()
+    return har
```
Llandy3d, line 440: small nit: can't we return directly the parse without the extra variable ?

src/main.ts
```diff
@@ -508,20 +492,15 @@ ipcMain.handle(
...
+    })
+    // TODO: https://github.com/grafana/k6-studio/issues/277
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const generator = GeneratorFileDataSchema.parse(JSON.parse(data))
 
-      return generator
-    } finally {
-      await fileHandle?.close()
-    }
+    return generator
```
Llandy3d, line 503: same nit as above

Llandy3d (dismissed):
Everything works 🙌

src/main.ts
going-confetti: Done, managed to remove 2 todos along the way 😎

Llandy3d (approved):

PR #466, opened 2025-02-11
feat: Show actions menu button when hover over file item

going-confetti:
## Description
Make it easier to discover file actions by adding a menu, that can be opened with keyboard or mouse click

## How to Test
Verify that both context menu (on right click) and this new menu work and look identical.

image


cristianoventura (approved):
👍

Issue #460, opened 2025-02-06
Error on startup in macOS: Javascript error has occured in the browser process


federicotdn:
### Brief summary

When starting up k6 Studio on a mac, I get:

![Image](https://github.com/user-attachments/assets/6fc44b27-2a0b-46c8-9c9f-00d8458dacf3)

The logs say:

```
[2025-02-06 17:00:02.054] [error] Unhandled Error: spawn /Applications/k6 Studio.app/Contents/Resources/x86_64/k6-studio-proxy ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

The proxy icon on the bottom-left shows as "offline".
This is running on an Intel-based mac from 2019.

### k6 Studio version

v0.12.0

### OS

macOS 14.3 Sonoma

### Steps to reproduce the problem

1. Start up k6 Studio.

### Expected behaviour

Error should not be present!

### Actual behaviour

Error is present.

federicotdn:
The app was run from the Applications folder, not Downloads.

PR #467, opened 2025-02-11
fix: intel mac proxy

Llandy3d:
## Description

Fix https://github.com/grafana/k6-studio/issues/460
Fix https://github.com/grafana/k6-studio/issues/486

The build for both mac architectures is done on an arm machine and that would make the packager packager only the `arm64` folder even on an `x86_64` architecture, this PR fixes that by including both folders on mac with a small cost of `71~MB` increase of application size

## Checklist

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -9,6 +9,17 @@ import { FusesPlugin } from '@electron-forge/plugin-fuses'
 import { FuseV1Options, FuseVersion } from '@electron/fuses'
 import { getPlatform, getArch } from './src/utils/electron'
 
+let resources = [
+  './resources/' + getPlatform() + '/' + getArch(),
+]
+
+if (getPlatform() === 'mac') {
+  resources = [
+    './resources/mac/arm64',
+    './resources/mac/x86_64',
+  ]
+}
+
 const config: ForgeConfig = {
   packagerConfig: {
     icon: './resources/icons/logo',
@@ -20,7 +31,7 @@ const config: ForgeConfig = {
       './resources/splashscreen.html',
       './resources/logo-splashscreen-dark.svg',
       './resources/logo-splashscreen.svg',
-      './resources/' + getPlatform() + '/' + getArch(),
+      ...resources,
     ],
     osxSign: {
       optionsForFile: () => {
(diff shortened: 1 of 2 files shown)


forge.config.ts
```diff
@@ -9,6 +9,17 @@ import { FusesPlugin } from '@electron-forge/plugin-fuses'
...
+let resources = [
+  './resources/' + getPlatform() + '/' + getArch(),
+]
+
+if (getPlatform() === 'mac') {
+  resources = [
+    './resources/mac/arm64',
+    './resources/mac/x86_64',
+  ]
+}
```
going-confetti, line 21: Nit: I'm not a big fan of reassigning variables, so wanted to suggest an alternative:
```ts
function getPlatformSpecificResources() {
  if (getPlatform() === 'mac') {
    return ['./resources/mac/arm64', './resources/mac/x86_64']
  }

  return ['./resources/' + getPlatform() + '/' + getArch()]
}

const config: ForgeConfig = {
  // ...
  extraResource: [
    // ...
    ...getPlatformSpecificResources()
  ]
}
```

forge.config.ts
```diff
@@ -9,6 +9,17 @@ import { FusesPlugin } from '@electron-forge/plugin-fuses'
 import { FuseV1Options, FuseVersion } from '@electron/fuses'
 import { getPlatform, getArch } from './src/utils/electron'
 
+let resources = [
+  './resources/' + getPlatform() + '/' + getArch(),
```
going-confetti, line 13: Can we use `path.join('./resources', getPlatform(), getArch())`?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "k6-studio",
   "productName": "k6 Studio",
-  "version": "0.12.0",
+  "version": "0.13.0",
```
going-confetti, line 4: This doesn't seem right 👀

forge.config.ts
Llandy3d: makes absolutely more sense :D

package.json
Llandy3d: this is just for testing and will be reverted before the PR gets merged 🙇

cristianoventura (approved):
🍎

Issue #472, opened 2025-02-12
Stale state after renaming a file


going-confetti:
* The tooltip is showing the previous name
* When clicking on rename again the previous name is showing

PR #487, opened 2025-02-12
fix: Old name is shown when trying to rename file again

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/472

## Description
* Fix https://github.com/grafana/k6-studio/issues/472
* Fix a bug with overwriting existing files when renaming

## How to Test
* Try renaming files - it should work as expected
* Try using a filename that already exists - the rename operation should fail with a toast notification shown to the user


cristianoventura (approved):
LGTM!

Issue #435, opened 2025-01-25
Validate existing test script which refer other files


ram-gupta:
Currently existing script can be validated. Couldn't find a way to validate test script which refer other files in a folder as there is no option to open a folder. Organising test code in multiple files is a common use-case.

Is support for working with folders planned in upcoming release or if its already there can you help understand how to use it?

Llandy3d:
Hi, work on a better workspace structure is planned for the future, currently the main support are for tests generated by the application and included in the current structure.

We currently support importing script files separately so for tests separated into multiple files I don't think we have proper way to handle them right now since we are prioritizing tests generated by the application 🤔

going-confetti:
Thank you for the issue report @ram-gupta,

As far as I understand, you tried to open the test script using the `Open external script` button and k6 Studio wasn't able to run it - is that correct?
<img width="359" alt="Image" src="https://github.com/user-attachments/assets/1f779493-40d5-45c8-b5c8-6ed7426736b0" />

I'm aware of this problem and will try to fix it.
Without including the script itself, can you elaborate a little bit on how the test code is organized in this case? Do you use CSV/JSON files for data parameterization, import local JavaScript modules, or have some test logic in these other files?

ram-gupta:
Hi @going-confetti, all of that and the test code is organised something like:
```
.
├── config/
│   ├── test_options.json
│   ├── test_vars.json
│   ├── test_data.csv
├── auth-sdk/
│   ├── auth.js
├── utils/
│   ├── helper.js
└── k6-test.js (test script)

```

going-confetti:
@ram-gupta this should be resolved in version `0.14.0`. Please re-open this issue if it isn't.

ram-gupta:
Hi @going-confetti, I checked 0.14.0 and the issue seems to be still there, I can't even see the script errors just a alert.

<img width="609" alt="Image" src="https://github.com/user-attachments/assets/39dd10b7-9b1b-42b8-a7b6-da1216261562" />

Llandy3d:
👋  An improved error with more details is coming in the next release, if you don't see other errors I assume this due to the new archive process we are running 🤔 

Could you try to run `k6 archive k6-test.js` on your script ?
It should report the error when k6 tries to create an archive of your script

ram-gupta:
Hi @Llandy3d, test script needs an env variable to run, so `k6 archive -e K6_ENV=dev test.js` runs fine but without env variable fails as expected.

[...]

PR #492, opened 2025-02-13
fix: Validator requires admin role on Windows

going-confetti:
## Description
* Fix the issue of not being able to validate tests on Windows in some cases
    * Recently, we changed how test runs were initiated to support external dependencies in future. We used `symlink` to create a temporary symlink, which didn't work on Windows under some conditions (we suspect it had something to do with the user role/permissions).
    * We switched to a different approach, one that should also address a recently reported issue with validating external scripts with dependencies
* Fix the issue of not being able to see the external script content
* Fix the issue of not being able to validate external scripts with dependencies (data files, external modules, etc)

## How to Test

- Test both `Validator` and `Validator dialog`
- Make sure external scripts work as well
- Test on Windows

## Related PR(s)/Issue(s)

Hopefully, it will also fix this bug: https://github.com/grafana/k6-studio/issues/435 - I created a very simple script with external dependencies and it worked, so 🤞

### DIFF
--- a/src/script.ts
+++ b/src/script.ts
@@ -1,18 +1,57 @@
 import { app, dialog, BrowserWindow } from 'electron'
 import { spawn, ChildProcessWithoutNullStreams } from 'node:child_process'
-import { readFile, writeFile, symlink, unlink } from 'fs/promises'
+import { readFile, writeFile, unlink } from 'fs/promises'
 import path from 'path'
 import readline from 'readline/promises'
 import { K6Check, K6Log } from './types'
 import { getArch, getPlatform } from './utils/electron'
-import {
-  DATA_FILES_PATH,
-  DATA_FILES_TEMP_PATH,
-  SCRIPTS_TEMP_PATH,
-} from './constants/workspace'
-import { TEMP_SCRIPT_FILENAME } from './constants/files'
+import { TEMP_K6_ARCHIVE_PATH, TEMP_SCRIPT_SUFFIX } from './constants/workspace'
 
 export type K6Process = ChildProcessWithoutNullStreams
+
+const spawnK6 = ({
+  args,
+  env = {},
+  onStdOut = () => {},
+  onStdErr = () => {},
+  onClose = () => {},
+}: {
+  args: string[]
+  env?: NodeJS.ProcessEnv
+  onStdOut?: (data: string) => void
+  onStdErr?: (data: string) => void
+  onClose?: (code: number) => void
+}) => {
+  let k6Path: string
+
+  if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
+    k6Path = path.join(
+      app.getAppPath(),
+      'resources',
+      getPlatform(),
+      getArch(),
+      'k6'
+    )
+  } else {
+    k6Path = path.join(process.resourcesPath, getArch(), 'k6')
+  }
+
+  // add .exe on windows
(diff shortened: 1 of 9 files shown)


src/script.ts
```diff
@@ -54,72 +33,143 @@ export const runScript = async (
...
-    '--vus=1',
-    '--iterations=1',
-    '--insecure-skip-tls-verify',
-    '--log-format=json',
-    '--quiet',
-    ...(!enableUsageReport ? ['--no-usage-report'] : []),
-  ]
-
-  // Create symlink to data folder so it can be accessed by the script
-  await symlink(DATA_FILES_PATH, DATA_FILES_TEMP_PATH, 'dir')
```
going-confetti, line 76: According to [this thread](https://github.com/nodejs/node/issues/18518#issuecomment-1509547477), it's not possible to create symlinks with `fs.symlink` unless the user is an admin

src/script.ts
```diff
@@ -54,72 +33,143 @@ export const runScript = async (
...
-
-  const k6 = spawn(k6Path, k6Args, {
-    env: { ...process.env, ...proxyEnv },
+  const k6 = spawn(k6Path, args, {
+    env: { ...process.env, ...env },
```
going-confetti, line 43: @Llandy3d do we need the whole `process.env` here?

Llandy3d: I think we do because we are expanding the default env and want to keep the other values as well 🤔

cristianoventura (changes requested):
Works as expected on Windows 🚀 🪟 

I noticed a few issues, one of them possibly a regression:

- When using the validator dialog, the spinner doesn't clear when the script finishes or errors out

https://github.com/user-attachments/assets/42f08fb6-ad6d-4337-bdc4-b94045e27c8b

- On macOS, I wasn't able to run one or two random external scripts I already had in disk but couldn't find a pattern as to why some of these files fail. I'll check them more closely and will come back once I find out what's going on.

> Unhandled rejection Error: Error invoking remote method 'script:run': Error: Failed to create archive: k6 process exited with code 104

src/script.ts
```diff
@@ -54,72 +33,143 @@ export const runScript = async (
...
+  browserWindow.webContents.send('script:log', logData)
+}
+
+export const archiveScript = (scriptPath: string): Promise<string> => {
+  return new Promise((resolve, reject) => {
+    const k6Args = ['archive', scriptPath, '-O', TEMP_K6_ARCHIVE_PATH]
+
+    spawnK6({
+      args: k6Args,
+      onClose: (code) => {
```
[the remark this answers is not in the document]
going-confetti: Yes, good point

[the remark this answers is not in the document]
going-confetti: Good point, added 👍

going-confetti:
> On macOS, I wasn't able to run one or two random external scripts I already had in disk but couldn't find a pattern as to why some of these files fail. I'll check them more closely and will come back once I find out what's going on.

Can you share these scripts? `104` [indicates](https://github.com/grafana/k6/blob/12d6b9e1b25e41b8de001ed1465affe3ed1fd375/errext/exitcodes/codes.go#L36) invalid configuration. I tried reproducing this with a script created in Test Builder and saw a different error related to scenarios/absence of the default export, though that's a separate thing

cristianoventura (approved):
LGTM!

e-fisher (approved):
Validating from generator doesn't show toast on success / failure, otherwise LGTM!

PR #497, opened 2025-02-18
internal(Data files): Add data files support in Parameterization rules

going-confetti:
## Description
Add a way to select value from file in Parameterization rule editor - this PR and the finishing UI touches are the biggest missing piece stopping us from making the feature public.

## How to Test
1. Enable the `data-files` feature
2. Import a JSON/CSV file
3. Open a generator
4. Add the file in `Test options` <- the UI will be updated
5. Add a parameterization rule and select `Data file` in selector
6. Verify that the generated script works

image

### DIFF
--- a/src/views/Generator/RuleEditor/ParameterizationEditor/FileSelect.tsx
+++ b/src/views/Generator/RuleEditor/ParameterizationEditor/FileSelect.tsx
@@ -1,0 +1,80 @@
+import { Code } from '@radix-ui/themes'
+import { ControlledSelect, FieldGroup } from '@/components/Form'
+import { ParameterizationRule } from '@/types/rules'
+import { useFormContext } from 'react-hook-form'
+import { useGeneratorStore } from '@/store/generator'
+import { useMemo } from 'react'
+import { useDataFilePreview } from '@/views/DataFile/DataFile.hooks'
+
+export function FileSelect() {
+  const {
+    control,
+    watch,
+    formState: { errors },
+  } = useFormContext<ParameterizationRule>()
+
+  const fileName = watch('value.fileName')
+  const propertyName = watch('value.propertyName')
+  const files = useGeneratorStore((store) => store.files)
+  const { data: preview, isLoading } = useDataFilePreview(fileName)
+
+  const fileOptions = useMemo(() => {
+    return files.map((file) => ({
+      value: file.name,
+      label: (
+        <Code size="2" truncate variant="ghost">
+          {file.name}
+        </Code>
+      ),
+    }))
+  }, [files])
+
+  const propsOptions = useMemo(() => {
+    return preview?.props.map((prop) => ({
+      value: prop,
+      label: (
+        <Code size="2" truncate variant="ghost">
+          {prop}
+        </Code>
+      ),
+    }))
+  }, [preview])
+
+  return (
+    <>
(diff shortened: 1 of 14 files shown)


src/views/Generator/RuleEditor/ParameterizationEditor/FileSelect.tsx
```diff
@@ -0,0 +1,85 @@
...
+            css: { maxWidth: 'var(--radix-select-trigger-width)' },
+            position: 'popper',
+          }}
+        />
+      </FieldGroup>
+      <FieldGroup
+        name="value.fileName"
+        errors={errors}
+        label="Key"
+        hint="Used to retrieve a value from a randomly selected row in the data file"
```
going-confetti, line 65: Does this make sense? Alternatives I had in mind were `Property` or even `Property`/`Column` depending on the file type (although I'd prefer to treat all data files types the same)

src/views/Generator/RuleEditor/ParameterizationEditor/VariableSelect.tsx
```diff
@@ -13,12 +13,12 @@ export function VariableSelect() {
       value: variable.name,
       label: (
         <Flex gap="1" align="center">
-          <Code size="2" truncate>
```
going-confetti, line 16: The contrast was really low (orange on orange) on hover

src/views/Generator/RuleEditor/ParameterizationEditor/FileSelect.tsx
cristianoventura: "Column" would be clearer IMO, because that's how the data is typically organized in CSVs, which makes it a more intuitive term for users to understand. "Key" might be a bit ambiguous 🤔

cristianoventura (dismissed):
LGTM! I found one validation issue that could either be fixed now or on a follow-up PR. 

This happens in two instances:

- When you de-select a data file that is being used by a parameterization rule, the "key" field is not cleared, throwing an error when you validate the script
- When you de-select all data files while having a parameterization rule, the "Replace with" dropdown still has the "Data file" option selected

https://github.com/user-attachments/assets/3de52c55-2b0e-484f-907b-ff3debe62980

Other than that, it would be good to introduce tests for `parameterization.ts` and `codegen.ts`, but these can also be done on a separate PR.

src/views/Generator/RuleEditor/ParameterizationEditor/FileSelect.tsx
going-confetti: I agree that it's rather ambiguous. But then "Column" might not be the right fit for JSON files

going-confetti:
> LGTM! I found one validation issue that could either be fixed now or on a follow-up PR.
> 
> This happens in two instances:
> 
>     * When you de-select a data file that is being used by a parameterization rule, the "key" field is not cleared, throwing an error when you validate the script
> 
>     * When you de-select all data files while having a parameterization rule, the "Replace with" dropdown still has the "Data file" option selected
> 

I partially addressed the first instance by doing what we do in Variables Editor (not letting the user remove a variable/file that is referenced in a rule). I'll improve it further in future PRs, though we'll need to make it even less error-prone in future by warning the user about removing recordings and data files that are use by a generator

cristianoventura (approved):

PR #499, opened 2025-02-18
feat: Apply rules in request list in generators

e-fisher:
## Description

This is the first PR towards improved rule preview - apply rules in request list in generators. Previously, we would show unmodified requests, with exception when opening a rule editor and viewing rule preview tab, which will be removed in future PRs. Also, it wasn't possible to see request list with all rules applied, just the one you have selected.

Per suggestion during our latest sync I've also added ability to switch between original request list and the one with rules applied. It affects request list, details inspector, script, and validator. This may come handy when debugging rules and user won't have to disable rules one-by-one.

## How to Test

Verify modifications added by rules are reflected in request list, inspector, script, and validator. Confirm with both "apply rules" toggle enabled and disabled.

## Checklist

https://github.com/user-attachments/assets/025d8e1b-efcd-4693-9674-5802c1e7009f


src/views/Generator/GeneratorTabs/GeneratorTabs.tsx
going-confetti, line 1: Is it expected to still have the Rule preview tab? Personally, I find it more confusing to have preview in two places and to be able to turn off rule application while being in the rule editor

src/views/Generator/TestRuleContainer/RulesDisabledOverlay.tsx
```diff
@@ -0,0 +1,29 @@
...
+export function RulesDisabledOverlay() {
+  const theme = useTheme()
+  const rulesEnabled = useGeneratorStore((store) => store.rulesEnabled)
+
+  if (rulesEnabled) {
+    return null
+  }
+
+  return (
+    <Flex
```
going-confetti, line 14: This doesn't prevent the user from using keyboard to edit the rules, but I'm not convinced we should disable/hide the UI if the switch is off - maybe keeping the rule list/rule editor accessible, but showing a very noticeable `Callout` component at the top of that pane will work better? Here's what it could look like (the copy is a quick draft)

image
image

Might be worth discussing this with @majavojnoska

src/store/generator/slices/rules.ts
```diff
@@ -4,6 +4,7 @@ import { ImmerStateCreator } from '@/utils/typescript'
 interface State {
   rules: TestRule[]
   selectedRuleId: string | null
+  rulesEnabled: boolean
```
going-confetti, line 7: I think it should have a different name, because we already have a concept of disabled rules and it means something else.

src/views/Generator/GeneratorTabs/GeneratorTabs.tsx
```diff
@@ -77,6 +80,20 @@ export function GeneratorTabs({
...
+                  size="1"
+                  color={!rulesEnabled ? 'gray' : undefined}
+                >
+                  <Flex gap="2">
+                    <Switch
+                      size="1"
+                      checked={rulesEnabled}
+                      onCheckedChange={setEnableRules}
+                    />
+                    Apply rules
```
going-confetti, line 94: To me, this part of the new UI is rather confusing - why would I not want to apply rules to my recording?
I also think that when it comes to script generation rules should be applied regardless of this setting.

I don't have a good copy suggestion, but maybe something focused on the recording rather than rules? The reason I think it might be better is because the goal of having this switch (as far as I understand) is for the user to be able to check the original recording without any test rules applied.

going-confetti (changes requested):

src/views/Generator/TestRuleContainer/RulesDisabledOverlay.tsx
e-fisher: Agree, this looks better 🙌 Replaced overlay with callout:

screencapture 2025-02-19 at 14 46 38

src/views/Generator/GeneratorTabs/GeneratorTabs.tsx
e-fisher: I've used "View original requests" and will revisit after we collect feedback

e-fisher:
> Is it expected to still have the Rule preview tab? Personally, I find it more confusing to have preview in two places and to be able to turn off rule application while being in the rule editor

Keeping the preview tab in this iteration to preserve the ability to see which requests are affected by the selected rule. Will be removed in the next PR.

Other comments have been addressed, and it's ready for another review 🙌

going-confetti (approved):
LGTM!

PR #500, opened 2025-02-18
internal(Cloud auth): Cloud authentication via Grafana Cloud

allansson:
## Description

This PR adds cloud authentication and profile management via Grafana Cloud SSO. It is implemented behind the "cloud-auth" feature flag, since we are currently missing the k6 APIs to exchange a Grafana access token for a user's personal API token.

I've completely made the UI up, but it's coded in a way that makes it easy to re-purpose (e.g. to put everything in a dialog).

## How to Test

First, there's a bit of configuration that's needed.

1. Log in to `grafana-dev.com`.
2. Go to the ksix org.
3. Find OAuth Clients in the left sidebar.
4. Create a new client and copy it's client ID.
5. Create the following `.env` file at the root of the repo:
    ```
    K6_API_URL=https://api.staging.k6.io/cloud/v6
    GRAFANA_CLIENT_ID=<CLIENT ID FROM ABOVE>
    GRAFANA_API_URL=https://grafana-dev.com/api
    GRAFANA_COM_URL=https://grafana-dev.com/
    ```

Now you can start the app:
1. Enable "cloud-auth" in the devtools dialog.
2. Click the new Profile icon in the bottom-left corner.
3. Sign in using your Grafana account.
4. Try signing out, denying authorization and anything you can think of.

## Checklist

https://github.com/user-attachments/assets/91fdd9d8-1d45-432d-aaac-ee2b001796c7


src/auth.ts
```diff
@@ -0,0 +1,283 @@
...
+    if (profile === null) {
+      return null
+    }
+
+    return profile.user
+  })
+
+  let pending: AbortController | null = null
+
+  ipcMain.handle('auth:sign-in', async (): Promise<SignInResult> => {
```
allansson, line 114: This function basically acts like a co-routine, where it sends the latest state to the UI and waits for the user doing something, e.g. selecting a stack. Once the entire process has completed, the result of the sign in is returned.

src/components/Layout/ActivityBar/ActivityBar.tsx
```diff
@@ -31,22 +36,34 @@ export function ActivityBar() {
...
         </Grid>
 
-        <Flex direction="column" align="center" gap="3" mt="auto">
+        <Flex direction="column" align="center" gap="3">
           <ThemeSwitcher />
           <ProxyStatusIndicator />
           <SettingsButton />
           <HelpButton />
+        </Flex>
+        <Separator orientation="horizontal" size="2" />
```
allansson, line 59: I'm letting the `Separator` be, otherwise I'd need to implement some kind of `<NotFeature />` component. 😅

src/main.ts
```diff
@@ -180,6 +182,8 @@ const createWindow = async () => {
     },
   })
 
+  initAuth(mainWindow)
+
```
allansson, line 186: Trying not to add more functionality to `main.ts`.

src/services/k6/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  signal: AbortSignal
+) {
+  await timeout(5_000)
+
+  signal.throwIfAborted()
+
+  return {
+    api_token: MOCK_PERSONAL_API_TOKEN,
+  }
+}
```
allansson, line 17: Ignore everything here... this is mocked until we get the new API endpoint.

src/views/Profile/Avatar.tsx
```diff
@@ -0,0 +1,21 @@
+import { css } from '@emotion/react'
+import { PersonIcon } from '@radix-ui/react-icons'
+import { Flex } from '@radix-ui/themes'
+
+export function Avatar() {
```
allansson, line 5: I know there's a `Avatar` component in radix-ui, but it only supports `solid` and `soft` variants which looked to dense IMO.

vite.main.config.ts
```diff
@@ -7,12 +7,33 @@ import {
...
 // https://vitejs.dev/config
 export default defineConfig((env) => {
   const forgeEnv = env as ConfigEnv<'build'>
   const { forgeConfigSelf } = forgeEnv
-  const define = getBuildDefine(forgeEnv)
+
+  const define = {
+    ...getBuildDefine(forgeEnv),
+    ...getDotEnv({
+      GRAFANA_CLIENT_ID: '<tbd>',
```
allansson, line 31: We don't have a client id in prod. Not sure if we should commit it either (although the client id is not meant to be a secret).

vite.main.config.ts
```diff
@@ -21,7 +42,7 @@ export default defineConfig((env) => {
         formats: ['cjs'],
       },
       rollupOptions: {
-        external,
+        external: external.filter((id) => id !== 'openid-client'),
```
allansson, line 45: Electron Forge wants to mark every package in `package.json` as external, but the `openid-client` needs to be bundled because it uses ES modules and electron doesn't support it (in node you can import CJS files from an ES module, but you can't `require` an ES module).

allansson (commented):
Self-review

allansson:
> Love the granular splitting of components and added comments! Code was easy to follow and understand 💯
> 
> Few comments from testing:

>     1. Could we use react-select for stack list to make it searchable? It's already added to the repo and has _a somewhat_ working wrapper for radix.

I considered using it, but opted not to because in 99.99% of cases it's not really needed. The raintank/ksix/ephemeral orgs aren't really representative of real world data. I actually think there are limitations to the number of stacks allowed based on your subscription.

But then again, it couldn't hurt. I'll see how it looks.
 
>     2. I've tried selecting an archived stack and followed the displayed instructions, but after logging in, I was stuck on that screen, and had to choose a different stack. Would it be possible to add "Retry" button?

The tricky part here is that the `archived` status comes from us fetching the list of stacks, so we'd need to refetch the stacks but the `access_token` never leaves the `sign-in` handler.

I have an idea on how to do it, though. 🤔

src/components/LinkButton.tsx
```diff
@@ -0,0 +1,12 @@
+import { Link, Reset } from '@radix-ui/themes'
+import { ComponentProps } from 'react'
+
+export function LinkButton(props: ComponentProps<'button'>) {
```
e-fisher, line 4: Looks like it does the same as `TextButton` https://github.com/grafana/k6-studio/blob/d14fe686a7c3bc5c8a4377969c1b876079eee2b4/src/components/TextButton.tsx#L2

Though your implementation is cleaner, can we remove the old one and use yours in both places instead?

e-fisher (commented):
The refresh doesn't seem to work for me, my stack is ready, but the message in studio say it's still archived. Is there an additional delay?


https://github.com/user-attachments/assets/e7f918b2-f49d-48d9-8b30-74bf316c53a9

vite.main.config.ts
Llandy3d: We could use github variables to insert it so that it can be changed ahead of build time 🤔

Llandy3d (commented):
I think a modal similar to the settings would work well for auth 👀 

 I also wonder if it would make sense to have a searchable table to also display more stack information 🤔

src/components/LinkButton.tsx
allansson: I will keep the `TextButton` name but replace the implementation. In retrospect, `LinkButton` sounds like a button that is a link.

vite.main.config.ts
allansson: Yeah, but that's one more step of setup... 

Thinking on it, I think we should just commit it to code. It's not secret. Once the binary is built anybody with enough motivation and access to Google can read it. 

Jumping through hoops to hide it adds complexity and all we achieve is security by obscurity.

allansson:
> I think a modal similar to the settings would work well for auth 👀

I like having it as a full size page (it just looks better imo) but I've changed it to a dialog because I'm sure you won't be the only one to ask if it shouldn't be a dialog.

> I also wonder if it would make sense to have a searchable table to also display more stack information 🤔

Until we know what that stack information would be and given the urgency of the feature, I'd rather not add scope-creep to this PR. 😅 

_PS. I looked in to having a list box instead of a dropdown, but Radix UI doesn't have one._ 😞

Llandy3d (dismissed):
🚀 🚀 🚀

src/components/Profile/GrafanaCloudSignIn/SelectingStack.tsx
```diff
@@ -0,0 +1,113 @@
...
+      )}
+      <Button
+        disabled={
+          selectedStack === null ||
+          selectedStack.status === 'archived' ||
+          selectedStack.status === 'restoring'
+        }
+        onClick={handleSelect}
+      >
+        Login
```
cinaglia, line 109: idea: It looks like we're using "logging in" and "signing in" interchangeably when referring to signing in using Grafana Cloud and selecting a Grafana stack. Perhaps we should stick to "signing in" for the process of authenticating and "selecting" (or maybe "connecting") for picking the current stack.

```suggestion
        Connect
```

allansson: I was actually thinking that to myself at some point, but I forgot about it and never got back to it. Thanks for bringing it up!

e-fisher (commented):
Few comments from testing latest changes:

1. The "stack is archived" message is no longer displayed, it allows selecting stack regardless of archived status. While I think that's an improvement to UX, just wanted to mention to make sure it's expected, and we don't need to wait for stack to be active?

2. Clicking logout doesn't log you out from Grafana account, but only from selected stack, you have to click it multiple times to logout from all stacks. Is this expected? 

https://github.com/user-attachments/assets/8d7c5299-3811-48d8-bd5b-6c54420cfa24


3. The dialog has a lot of unused space, perhaps we could lower the height to it's content size? 

screencapture 2025-02-26 at 09 47 19

allansson:
> The "stack is archived" message is no longer displayed, it allows selecting stack regardless of archived status. While I think that's an improvement to UX, just wanted to mention to make sure it's expected, and we don't need to wait for stack to be active?

Bug introduced by switching from a `boolean` to string literal enum. Will fix.

> Clicking logout doesn't log you out from Grafana account, but only from selected stack, you have to click it multiple times to logout from all stacks. Is this expected?

Yes. I changed this so that you're signed in to a specific stack. This allows users to sign in from different Grafana Cloud accounts. Before I had to tell the user to sign-out first, losing all the stacks they painstakingly signed in to before.

Suppose I could change it so that we group the stacks by the Grafana account as well.

> The dialog has a lot of unused space, perhaps we could lower the height to it's content size?

Definitely.

allansson:
>The "stack is archived" message is no longer displayed, it allows selecting stack regardless of archived status. While I think that's an improvement to UX, just wanted to mention to make sure it's expected, and we don't need to wait for stack to be active?

@e-fisher This is working for me. Are you sure that the stack you are testing is actually archived?

> While I think that's an improvement to UX, just wanted to mention to make sure it's expected, and we don't need to wait for stack to be active

The reason we're showing the message is that restoring a stack can take a really long time (and the API is even slower to update its status). We'd have to do a lot of polling and handling timeouts, etc.

And in the end, I don't think it's that common for users to want to access an archived stack. This is another case of Raintank/ksxi being horrible examples of real world usage.

src/components/Layout/ActivityBar/ActivityBar.tsx
going-confetti:
```suggestion
        <Feature feature="cloud-auth">
          <Separator orientation="horizontal" size="2" />
        </Feature>
```

src/components/Layout/ActivityBar/Profile.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+function ProfileDialog({
+  open,
+  onOpenChange,
+}: {
+  open: boolean
+  onOpenChange: (open: boolean) => void
+}) {
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
```
going-confetti, line 14: image

I was a bit puzzled at first, because I expected to see a close button somewhere. One can argue that our users should know that clicking outside/hitting `esc` closes a dialog, but I think it should be more explicit (I'm actually surprised radix dialogs don't have the `x` button :/)

e-fisher (dismissed):
LGTM! 🙌 Found a few minor issues, but don't think they should be blockers.

1. I'm still able to reproduce this, had to re-authenticate after "waking up" my stack to be able to select it.

>e-fisher: The refresh doesn't seem to work for me, my stack is ready, but the message in studio say it's still archived. Is there an additional delay?

2. When I'm already logged into a stack, clicking "sign in into another stack" and then cancelling results in "logged out" modal state. Closing and reopening modal shows correct state.

https://github.com/user-attachments/assets/a97cdf0d-6d15-404c-a1ba-001d2a67b8b6

Llandy3d (approved):
Looks great 🙌 

Same with edgar point n.2, cancelling while signed it returns the wrong state

cristianoventura (dismissed):
LGTM! 👍

Llandy3d (approved):
🚀 🚀 🚀

PR #502, opened 2025-02-20
feat: In-browser element inspector for browser recording

allansson:
## Checklist


src/views/Recorder/BrowserEventLog.tsx
allansson, line 1: This file was split up into individual files. Pretty much every new file in the `BrowserEventLog` is just a component extracted as-is from this file. The notable exceptions are the `Selector` and `PageNavigationDescription` components.

src/views/Recorder/BrowserEventLog/EventDescription/Selector.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (isRecording) {
+      window.studio.browserRemote.highlightElement(selector)
+    }
+  }
+
+  const handleMouseLeave = () => {
+    if (isRecording) {
+      window.studio.browserRemote.highlightElement(null)
+    }
+  }
```
allansson, line 22: This logic has been added since the component was split from `BrowserEventLog`.

package.json
```diff
@@ -85,6 +86,8 @@
     "@radix-ui/react-icons": "^1.3.0",
     "@radix-ui/react-label": "^2.1.0",
     "@radix-ui/react-toast": "^1.2.1",
+    "@radix-ui/react-toggle-group": "^1.1.2",
+    "@radix-ui/react-tooltip": "^1.1.8",
```
allansson, line 90: I cannot use `@radix-ui/theme` in the browser extension. We are using a shadow DOM to isolate our styling from the recorded page, but the `<Theme />` component mounts its styles in `document.head`. This means that the theme styles are applied to the recorded page, and not to our shadow DOM. As far as I can tell, there's no way of specifying where the styles should be mounted.

package.json
```diff
@@ -73,6 +73,7 @@
     "@dnd-kit/core": "^6.1.0",
     "@dnd-kit/modifiers": "^7.0.0",
     "@dnd-kit/sortable": "^8.0.0",
+    "@emotion/cache": "^11.14.0",
```
allansson, line 76: We need to configure emotion to inject the style elements inside our shadow root, and that requires us to configure a custom cache.

e-fisher (approved):
LGTM! 🚀 
Should we drop the dot after event description, looks a bit weird with div class names? 
screencapture 2025-02-26 at 10 49 52

cristianoventura (approved):
Looks great and works as expected! 🚀 

I noticed the `ToolBoxButton` component is not visible in some pages. Increasing the `z-index` to a higher value should solve this problem.

image

image

PR #505, opened 2025-02-21
feat: multiple correlation extraction support

### LINKED ISSUE #309: Allow for multiple extraction support from a single correlation rule
When using a `correlation rule` after defining the `Extractor`, it currently extracts the value only from the first request that matches. If multiple requests are matched for extraction, they are ignored.

This is a good default but there could be the possibility that the user wants to create a single correlation rule and have it extract all the requests matching the selector. This allows for setting up a single rule and the value would be replaced on each extraction at runtime.
We should add support for toggling between these two behaviours for the correlation rule.

---

As a user I should be a able to toggle between the current way the correlation rule works and a new behaviour where the same rule extracts the value from each request it matches at runtime.

Llandy3d:
## Description

Allows for the configuration of the `Correlation Rule` between a `single` and `multiple` extraction mode.

UX needs to be improved or actually just the ui 🤔 
There is a lint error on the `v0` of the generator schema and I would need feedback to better understand how to handle this scenario 🙇 

We also need to decide if `multiple` should be the default behaviour, I remember it being recommended but it would be great to get a second opinion.

Closes #309 

- [x] Confirm default behaviour before merging `single` vs `multiple`

## How to Test

Create a new recording making multiple requests with POST data to `https://test-api.k6.io/my/crocodiles/`
Create a new Generator with a correlation rule matching on the csrftoken in the form
Checkout the branch and see that the rules loads correctly and it's in `multiple` mode
Toggle between single and multiple, checking that code is correlating correctly

## Checklist


src/views/Generator/RuleEditor/CorrelationEditor.tsx
```diff
@@ -34,6 +35,29 @@ export function CorrelationEditor() {
         <Text size="2" as="p" mb="2" color="gray">
           Extraction value for correlation.
         </Text>
+        <Box mb="3">
+          <Label>Extraction Mode</Label>
+          <Select.Root
```
going-confetti, line 40: This should be a switch or two radio buttons

src/views/Generator/RuleEditor/CorrelationEditor.tsx
```diff
@@ -34,6 +35,29 @@ export function CorrelationEditor() {
...
+            }
+          >
+            <Select.Trigger />
+            <Select.Content>
+              <Select.Item value="single">
+                Single (extract once)
+              </Select.Item>
+              <Select.Item value="multiple">
+                Multiple (extract every time)
+              </Select.Item>
```
going-confetti, line 54: If Single/Multiple is not descriptive enough, why not call the options `Extract once` and `Extract every time`/`Extract every occurrence`? Maybe even `Extract first occurrence`/`Extract every occurrence` because it gives a better idea of which occurrence/value will be extracted than `once`.

src/views/Generator/RuleEditor/CorrelationEditor.tsx
```diff
@@ -34,6 +35,29 @@ export function CorrelationEditor() {
...
+            </Select.Content>
+          </Select.Root>
+          <Text size="1" as="p" mt="1" color="gray">
+            Choose whether to extract the value once or on every matching response.
+          </Text>
```
going-confetti, line 59: This is redundant because it repeats what's written in the options

going-confetti (changes requested):
Added some suggestions re: UI

src/schemas/generator/v1/rules.ts
```diff
@@ -123,6 +123,7 @@ export const ParameterizationRuleSchema = RuleBaseSchema.extend({
 
 export const CorrelationRuleSchema = RuleBaseSchema.extend({
   type: z.literal('correlation'),
+  extractionMode: z.enum(['single', 'multiple']).default('multiple'),
```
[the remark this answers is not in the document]
Llandy3d: Good suggestion 🤔

cristianoventura (approved):
LGTM! 🙌 

> There is a lint error on the v0 of the generator schema and I would need feedback to better under

Was this already fixed?

Llandy3d:
image

we might need more space between elements 🤔

src/views/Generator/RuleEditor/CorrelationEditor.tsx
```diff
@@ -5,9 +5,21 @@ import { FilterField } from './FilterField'
 import { SelectorField } from './SelectorField'
 import { Label } from '@/components/Label'
 import { useFormContext } from 'react-hook-form'
+import { FieldGroup } from '@/components/Form'
+import { ControlledRadioGroup } from '@/components/Form/ControllerRadioGroup'
+
+const EXTRACTION_MODE_OPTIONS = [
+  { value: 'single', label: 'First occurence' },
+  { value: 'multiple', label: 'All occurences' },
+]
```
going-confetti, line 14: I've been thinking about what @allansson said the other day and I think this could indeed be improved.
Maybe a switch with two labels: `Use first value` and `Use latest value`? It'd also be a better description of how the rule actually works (`all` is too ambiguous IMO).

What do you think?

Llandy3d: Not sure if it's easy to understand with `first` and `latest` value 🤔 
Same for `all`, although it seems to make it clearer on a conceptual level (more than one) I agree it might not be the best.
I would say, let's gather more feedback 🤔

src/schemas/generator/index.ts
```diff
@@ -13,12 +13,29 @@ export function migrate(generator: z.infer<typeof AnyGeneratorSchema>) {
     case '0':
       return migrate(v0.migrate(generator))
     case '1.0':
+      migrateEmptyExtractionModeInCorrelation(generator)
```
Llandy3d, line 16: after much thought I've opted for a migration to align the old `single` mode, migration system is really cool 🙌

going-confetti: `migrate` is essentially a recursive function: once the file has been migrated to the latest version (currently 1.0`, it should just return the generator object.
If it requires extra migration, then it simply needs a new schema version. Otherwise once we add v2, it will look like this:
```ts
  switch (generator.version) {
    case '0':
      return v1.migrate(v0.migrate(generator))
    case '1.0':
      return migrate(migrateEmptyExtractionModeInCorrelation(generator))
    case '2':
      return generator
    default:
      return exhaustive(generator)
  }
```

src/schemas/generator/index.ts
```diff
@@ -13,12 +13,29 @@ export function migrate(generator: z.infer<typeof AnyGeneratorSchema>) {
...
       return generator
     default:
       return exhaustive(generator)
   }
 }
 
+// We added a new field `extractionMode` with a default value of `multiple` for the Correlation Rule
+// This rule was working in a `single` extraction mode previously so this migration checks that the value
+// is not defined and in that case we manually set the previous way of working
+function migrateEmptyExtractionModeInCorrelation(
```
going-confetti, line 26: If this function is required, then:
1. It should be declared in one of the `v1` files, as it won't be needed in versions after that
2. it shouldn't mutate the generator object

src/schemas/generator/v1/rules.ts
```diff
@@ -104,6 +104,7 @@ export const CorrelationExtractorSchema = z.object({
   filter: FilterSchema,
   selector: ExtractorSelectorSchema,
   variableName: z.string().optional(),
+  extractionMode: z.enum(['single', 'multiple']).default('multiple').optional(),
```
going-confetti, line 107: If you decide to create a new schema version, it should be a required property

going-confetti (changes requested):
Not sure if a new schema version is needed, but I think the new migration function doesn't follow our schema migration approach

going-confetti (approved):

PR #508, opened 2025-02-24
fix: Selected rule is not reset when opening another generator

going-confetti:
## Checklist

### DIFF
--- a/src/store/generator/useGeneratorStore.ts
+++ b/src/store/generator/useGeneratorStore.ts
@@ -48,6 +48,7 @@ export const useGeneratorStore = create<GeneratorStore>()(
       recording = []
     ) =>
       set((state) => {
+        state.selectedRuleId = null
         // options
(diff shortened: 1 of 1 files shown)


src/store/generator/useGeneratorStore.ts
```diff
@@ -48,6 +48,7 @@ export const useGeneratorStore = create<GeneratorStore>()(
       recording = []
     ) =>
       set((state) => {
+        state.selectedRuleId = null
```
going-confetti, line 51: It's not a big issue, because it only occurs when you have rules with the same id (so pretty much only when you have copies of the same generator file).

going-confetti: We should probably use router for this, but it would be a bigger change

cristianoventura (approved):
LGTM!

PR #509, opened 2025-02-24
fix: Code editor scrollbars aren't consistent with app UI

going-confetti:
## Description

image

### DIFF
--- a/src/components/Monaco/ReactMonacoEditor.tsx
+++ b/src/components/Monaco/ReactMonacoEditor.tsx
@@ -24,6 +24,10 @@ const defaultOptions: monaco.editor.IStandaloneEditorConstructionOptions = {
   scrollBeyondLastLine: false,
   scrollbar: {
     alwaysConsumeMouseWheel: true,
+    verticalSliderSize: 4,
+    horizontalSliderSize: 4,
+    verticalScrollbarSize: 12,
+    horizontalScrollbarSize: 12,
   },
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Monaco/ReactMonacoEditor.tsx
```diff
@@ -24,6 +24,10 @@ const defaultOptions: monaco.editor.IStandaloneEditorConstructionOptions = {
   scrollBeyondLastLine: false,
   scrollbar: {
     alwaysConsumeMouseWheel: true,
+    verticalSliderSize: 4,
+    horizontalSliderSize: 4,
+    verticalScrollbarSize: 12,
```
going-confetti, line 29: Adds "buffer" of 4px each side to match the Radix ScrollArea style and make it easier to click on when operated with mouse

src/globalStyles.ts
```diff
@@ -25,4 +25,14 @@ export const globalStyles = css`
   .rt-SelectItem > span:not(.rt-SelectItemIndicator) {
     width: 100%;
   }
+
+  /* Make Monaco scroll thumbs rounder */
```
going-confetti, line 29: I tried matching the track styles as well, but the vertical and horizontal scrollbars would overlap each other in the bottom right corner and it looked off. Maybe we'll find a solution in future

cristianoventura (approved):
I tested on both Windows and Mac and it looks great!

Issue #363, opened 2024-11-22
Support for Test data files


Llandy3d:
We need to support Test Data files for use with the `Parameterization` rule.
This could be `csv` or `json` files that contain multiple entries of values to be used in tests.

We are supporting test variables but for real world usage we need the ability to import json/csv files with test data that the users can use in their script.
This is going to need a proper design doc and unanswered questions we have are like:

- Where does test data live ?
- Is test data specific to a single Generator or is it shared between all of them ?
- Which modes of VU sharing do we allow ?

A simple implementation could be something in the lines of this, by using the new csv module:

![Image](https://github.com/user-attachments/assets/501bb878-4aaf-4d8f-97a5-2df7f17e4c56)

But this is a really open problem and we need to gather more detailed information on specific use cases where it makes sense to use data files.

---

This ticket could be divided in two steps:

- [ ] Explore how to tackle and where to put test data files
- [ ] Implement the logic for handling test data

Issue #507, opened 2025-02-24
Move Test Data options into its own section


Llandy3d:
Currently Test Data dialogs (`variables` and `data files`) are found inside of the `Test Options` popover.
We should move these configurations outside the nested Test Options and into a new popover specific to test data.

<img width="803" alt="Image" src="https://github.com/user-attachments/assets/d58cda44-4752-4594-9ca6-40a8e6878aa9" />

PR #510, opened 2025-02-25
feat: Add data files support

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/363 https://github.com/grafana/k6-studio/issues/507

## Description
Add support for data files in parameterization (and custom code) rules. This is the first iteration of the feature.

- Supported file formats: JSON, CSV
- Max file size: 10mb
- CSV files must be comma-separated
- CSV files should have a header, otherwise the first row will be treated as such
- nesting is not supported in JSON

## Known issues
1. It's possible to import two files files with the same name, but different file type - I believe it's an edge case and I would like to tackle it later with the addition of file type icons (at the moment, we don't have them). Let me know if you think it's a dealbreaker.
2. I noticed a regression in the PageHeader component that makes it impossible to view long file names (see screenshot 1). Since it wasn't introduced in this PR, I will fix it separately.

## How to Test

* Import a data-file
* Use its value in a parameterization rule

image
image
image


src/views/Generator/TestData/DataFiles.tsx
```diff
@@ -0,0 +1,129 @@
...
+  )
+
+  return (
+    <Table.Row
+      css={css`
+        vertical-align: middle;
+      `}
+    >
+      <Table.Cell>{file.name}</Table.Cell>
+      <Table.Cell>Unique item per iteration</Table.Cell>
```
cristianoventura, line 112: Maybe `Random item per iteration` is a bit more clear?

src/views/Generator/TestData/DataFiles.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <Table.Cell>
+        <Tooltip
+          content="Data file is referenced in a rule"
+          hidden={!isFileInUse}
+        >
```
cristianoventura, line 117: 🤩

cristianoventura (dismissed):
LGTM!

src/views/Generator/TestData/DataFiles.tsx
going-confetti: It's not random 😬 it's access array values based on `scenario.iterationInTest` as described here. I'll try to come up with better copy

src/views/DataFile/DataFileTable.tsx
```diff
@@ -29,11 +29,15 @@ export function DataFileTable({ preview }: DataFileTableProps) {
...
-      <Table.Root size="1" variant="surface">
+      <Table.Root size="1" variant="surface" layout="auto">
         <Table.Header>
           <Table.Row>
             {preview.props.map((prop) => (
-              <Table.ColumnHeaderCell key={prop} uppercase={false}>
+              <Table.ColumnHeaderCell
+                key={prop}
+                uppercase={false}
+                maxWidth="300px"
```
going-confetti, line 39: Added `maxWidth` to table columns to make sure it will look decent even with longer string values

Llandy3d (approved):
🚀 🚀 🚀 

When importing a data file we don't jump into that view nor show a notification, I would possibly consider the notification of successful data file import 🤔

PR #512, opened 2025-02-28
feat: Highlight requests affected by rules

e-fisher:
## Description

This PR removes the rule preview tab and highlights the affected request in the request list.

1. Reduce opacity of requests that didn't match rule's query when editing a rule, creating a "highlight" effect of matched rules.
2. Remove rule preview tab since all changes injected by rules are now visible in request list.
3. Show value extracted by correlation in rule editor.
4. Fix a bug when correlated values in request host were not reflected in request list.

I'm also working on making extracted value clickable, which would select the source request in the list, but it required additional refactoring and will be added in a separate follow-up PR.

## How to Test

Verify that only affected requests are highlighted, test with all rule types. Try inspecting requests in other views (recorder, validator) and verify it works as expected. 

## Checklist

screencapture 2025-02-28 at 17 26 13


Llandy3d (commented):
This is just initial feedback and not complete, two things I've noticed:

1. On the light theme, seems like there is not enough differentiation between a request affected or not by the rule 🤔 
2. The `Extracted value` line disappears when there is no extraction, we might want to keep it mentioning something in the lines of nothing matched ? (The disappearing could be confusing 🤔 )

src/views/Generator/RuleEditor/CorrelationEditor.tsx
```diff
@@ -25,6 +28,14 @@ export function CorrelationEditor() {
     }
   }
 
+  const extractedValue = useMemo(() => {
+    if (selectedRuleInstance?.type !== 'correlation') {
+      return
+    }
+
+    return selectedRuleInstance.state.extractedValue
+  }, [selectedRuleInstance])
```
going-confetti, line 37: `selectedRuleInstance` is already `useMemo`'ed, so this can probably be removed in favor of
```ts
const extractedValue = selectedRuleInstance?.type !== 'correlation' ? selectedRuleInstance.state.extractedValue : undefined
```

src/components/WebLogView/Row.tsx
```diff
@@ -8,24 +8,44 @@ import { TableCellWithTooltip } from '../TableCellWithTooltip'
...
 import { getRequestType } from './WebLogView.utils'
 import { SearchResults } from './SearchResults'
+import { useMemo } from 'react'
 
 interface RowProps {
   data: ProxyDataWithMatches
   isSelected?: boolean
   onSelectRequest: (data: ProxyDataWithMatches) => void
   filter?: string
+  highlightedRequestIds?: string[]
```
going-confetti, line 18: Note: I think we're approaching a moment when we'll need to make WebLogView more modular and to essentially have 2 separate components for Generator (where we need highlighted requests) and Recorder (when we add some recording editing functionality perhaps?)

src/components/WebLogView/Row.tsx
```diff
@@ -8,24 +8,44 @@ import { TableCellWithTooltip } from '../TableCellWithTooltip'
...
+  filter,
+  highlightedRequestIds,
+}: RowProps) {
+  const opacity = useMemo(() => {
+    if (!highlightedRequestIds) {
+      return 1
+    }
+
+    return highlightedRequestIds.includes(data.id) ? 1 : 0.4
+  }, [highlightedRequestIds, data.id])
```
going-confetti, line 34: I saw @Llandy3d's comment about having more noticeable difference, but I'd like to challenge the idea of using opacity here mostly for accessibility/usability reasons:
For example, when I select a rule, I may want to edit its filter based on the request list. Having low opacity makes this pretty hard. Do you have alternative ideas?

going-confetti (commented):
Looks and feels great! Added some thoughts in the comments

src/components/WebLogView/Row.tsx
Llandy3d: Colors could be a good approach also to differentiate between requests matched for the extraction and requests matched for the replacing 🤔 
I think this is a good followup item we've been discussing but for a different PR

src/views/Generator/RuleEditor/CorrelationEditor.tsx
e-fisher: Good point, I've also added invariant to narrow down selected rule instance type

src/components/WebLogView/Row.tsx
e-fisher: Agree, I've added ability to inject Row component through props and removed `highlightedRequestIds` dependency from WebLogView

src/components/WebLogView/Row.tsx
e-fisher: As discussed, I'll work on improved highlighting and present in a separate PR

src/views/Generator/GeneratorTabs/RequestRow.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <Row
+      {...props}
+      css={{
+        opacity,
+        td: {
+          // Match border color with hihglighted rows
+          boxShadow: `inset 0 -1px var(--gray-${opacity === 1 ? 'a5' : 'a8'})`,
+        },
+      }}
```
e-fisher, line 23: For this iteration, I was able to get away with just style injection to achieve the same functionality. But this will also allow further customization.

src/components/WebLogView/WebLogView.tsx
```diff
@@ -41,43 +43,36 @@ export const WebLogView = memo(function WebLogView({
     onSelectRequest(selectedRequest)
   }, [selectedRequest, onSelectRequest])
 
-  if (groups !== undefined) {
```
e-fisher, line 44: This condition was added to support both grouped and ungrouped recordings, but we always default to at least one group (even for imported hars) so this condition became redundant.

cristianoventura (approved):
Approving since the highlighting will be addressed separately. I tested with all rules types and it looks good! This approach is a lot easier to manage without having to look at the "Rule preview" tab. 🥇

PR #513, opened 2025-03-02
feat: linux support

### LINKED ISSUE #336: Linux support
Development has happened primarily on Windows and macOS machines and slowed down for linux due to some browser launching bugs.

We should explore again the underlying issue with that feature on linux and start making releases also on linux.

Llandy3d:
## Description

Linux support.
It also includes some other fixes that have been commented.

- [x] Fix icon not appearing in dock
- [x] Add amd64 support
- [x] Add arm64 support

Closes #336 

## How to Test

Clone the repository and run `npm start` on a Linux machine or virtual machine

## Checklist


src/main.ts
```diff
@@ -313,8 +315,16 @@ ipcMain.handle('browser:start', async (event, url?: string) => {
...
   if (currentBrowserProcess) {
-    await currentBrowserProcess.close()
+    // macOS & windows
+    if ('close' in currentBrowserProcess) {
+      await currentBrowserProcess.close()
+      // linux
+    } else {
+      currentBrowserProcess.kill()
+    }
+
```
Llandy3d, line 327: Probably not the best of the best approach here but considering I want to raise removing the puppeteer dependency it seems good enough.

Basically we just check if we have the `close` from the puppeteer object and if not we just call `kill`.

src/proxy.ts
```diff
@@ -125,7 +125,7 @@ export const getCertificatesPath = () => {
   if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
     return path.join(app.getAppPath(), 'resources', 'certificates')
   } else {
-    return path.join(process.resourcesPath, 'certificates')
+    return path.join(app.getPath('userData'), 'certificates')
```
Llandy3d, line 128: we were still using the resources path for storing the certificates, this gives a permission error on linux but regardless, it's better to use the `userData` path for this data

install-k6.js
```diff
@@ -5,6 +5,7 @@ const K6_VERSION = 'v0.55.0'
 const K6_PATH_MAC_AMD = `k6-${K6_VERSION}-macos-amd64`
 const K6_PATH_MAC_ARM = `k6-${K6_VERSION}-macos-arm64`
 const K6_PATH_WIN_AMD = `k6-${K6_VERSION}-windows-amd64`
+const K6_PATH_LINUX_AMD = `k6-${K6_VERSION}-linux-amd64`
```
cristianoventura, line 8: If we're only supporting amd64 for now, maybe we should abort the installation when `process.arch` isn't supported.

cristianoventura (commented):
This is looking pretty good! Great job getting it running on Linux 🐧

I was able to get it running on `arm64` after some tweaks. However, this only works with Chromium by changing the executable path in Settings. If we are not considering arm64 for the next release, maybe we can add some notes to the README either by saying it isn't supported or, it can run with Chromium.

I found a few things that we missed for Linux. Let me know if you need any help addressing any of these.

- There is no way to open the "Logs" folder through the UI (Help > Application Logs) as we are not setting the appropriate executable for Linux here:

https://github.com/grafana/k6-studio/blob/e58252d2440fa4cace2a73e12d5d81ac4ffa2481/src/logger.ts#L37-L39

For GNOME base systems such as UIbuntu, we should add one more condition here to use `nautilus`.

- The allowed file extension when selecting the browser path is `exe` as we have a missing condition for Linux here:

https://github.com/grafana/k6-studio/blob/e58252d2440fa4cace2a73e12d5d81ac4ffa2481/src/settings.ts#L125-L127

- The k6 studio logo is broken

image

install-k6.js
Llandy3d: I would say, let's add it :)

Llandy3d:
For the logs I kept `open` also for linux, seems like it's the better choice than `nautilus` 🤔 
I also changed the `process.platform` into the `getPlatform()` helper to have a consistent platform name for various possible values.

Regarding the icon I'm not sure why it's not working for you, I didn't notice any issues on `amd64` 🤔 🤔  (beside the icon not appearing in the dock that today is automagically appearing, so I guess some form of icon cache 🤷 )

README.md
```diff
@@ -43,10 +43,12 @@ https://github.com/user-attachments/assets/46ccd340-804f-4aa5-b3a4-acec9f6a2dcd
...
 ## Installation
 
-Grafana k6 Studio is available for Mac and Windows (Linux support coming soon). You can download the latest version from the [Releases page](https://github.com/grafana/k6-studio/releases).
+Grafana k6 Studio is available for Mac and Windows and Linux. You can download the latest version from the [Releases page](https://github.com/grafana/k6-studio/releases).
 
 > [!IMPORTANT]  
-> [Google Chrome](https://www.google.com/chrome/browser-tools/) needs to be installed on your machine for the recording functionality to work.
+> For the recording functionality to work on Linux arm64, [Chromium](https://www.chromium.org/Home/) needs to be installed.
+>
+> Other platforms (including Linux x86_64), [Google Chrome](https://www.google.com/chrome/browser-tools/) needs to be installed.
```
cristianoventura, line 51: I've updated the readme to include some instructions for Linux arm64.

install-k6.js
```diff
@@ -52,6 +54,29 @@ Remove-Item -Path "${K6_PATH_WIN_AMD}" -Recurse
   execSync(command, { shell: 'powershell.exe' })
 }
 
+const getLinuxK6Binary = () => {
+  const arch = process.arch === 'arm64' ? 'arm64' : 'x86_64'
+  const binaryName = arch === 'arm64' ? K6_PATH_LINUX_ARM : K6_PATH_LINUX_AMD
```
cristianoventura, line 59: I've included the k6 binary to work on Linux arm64

src/browser.ts
```diff
@@ -10,15 +10,23 @@ import { mkdtemp } from 'fs/promises'
...
 
-function getBrowserPath() {
+async function getBrowserPath() {
   const { recorder } = appSettings
 
   if (recorder.detectBrowserPath) {
+    if (getPlatform() === 'linux' && process.arch === 'arm64') {
+      // Chrome is not available for arm64, use Chromium instead
+      return await getChromiumPath()
+    }
```
cristianoventura, line 28: Chrome isn't available on Linux arm64 so we're defaulting to Chromium instead.

cristianoventura (commented):
@Llandy3d I've added changes needed to get Linux working on arm64, including the `k6-studio-proxy` and `k6` binaries!

[Screencast from 2025-03-04 11-50-30.webm](https://github.com/user-attachments/assets/15ce20e8-66f8-41fa-8f03-06cce2ebd657)

src/browser.ts
Llandy3d: We could actually attempt both in all platforms, starting with Chrome and having Chromium as a fallback 🤔

README.md
```diff
@@ -43,10 +43,12 @@ https://github.com/user-attachments/assets/46ccd340-804f-4aa5-b3a4-acec9f6a2dcd
 
 ## Installation
 
-Grafana k6 Studio is available for Mac and Windows (Linux support coming soon). You can download the latest version from the [Releases page](https://github.com/grafana/k6-studio/releases).
+Grafana k6 Studio is available for Mac and Windows and Linux. You can download the latest version from the [Releases page](https://github.com/grafana/k6-studio/releases).
```
going-confetti, line 46:
```suggestion
Grafana k6 Studio is available for Windows, Mac, and Linux. You can download the latest version from the [Releases page](https://github.com/grafana/k6-studio/releases).
```

install-k6.js
going-confetti: I presume this is no longer accurate as we should be able to support both now?

src/browser.ts
going-confetti: I think this is the way to go: If Chromium is supported, one doesn't have to have Chrome installed on Windows/Mac to use Recorder.

src/browser.ts
```diff
@@ -54,24 +56,34 @@ export const launchBrowser = async (
...
+    `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
+    disableChromeOptimizations,
+    url?.trim() || 'about:blank',
+  ]
+
+  // if we are on linux we spawn the browser directly and attach the on exit callback
+  if (getPlatform() === 'linux') {
+    const browserProc = spawn(path, args)
+    browserProc.once('exit', sendBrowserClosedEvent)
+    return browserProc
```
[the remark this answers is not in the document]
going-confetti: Are there any reasons not to use this approach for all platforms?

src/logger.ts
```diff
@@ -35,7 +36,9 @@ export function openLogFolder() {
   const logPath = path.dirname(logFile)
 
   // supports only Mac and Windows at this time
```
going-confetti, line 38: 👀

going-confetti (commented):
Can you add testing instructions for both archs?

install-k6.js
Llandy3d: Correct we support both

src/browser.ts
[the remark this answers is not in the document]
Llandy3d: just the time to properly test everything!
I would define the plan as:
1. Release support for linux
2. Remove `puppeteer` launching logic on Windows/Mac
3. Adjust chromium launching/finding logic also on Windows/Mac

I do not want the linux changes to affect the other platforms at this stage and I would have the last 2 items as separate tickets and PRs 🤔

Llandy3d:
I have removed the updates to the `README.md` file because those would appear as soon as we merge this PR but we still don't have a `Linux` release out there. 
The readme will be updated on a separate occasion 🙇

cristianoventura (dismissed):
I tested on Linux arm64 and works great with Chromium! I'll be working on a separate branch to support Chromium by default on all platforms.

going-confetti (dismissed):

cristianoventura (approved):

PR #518, opened 2025-03-03
internal(Load Zones): Ability to add load zones

cristianoventura:
## Description

This PR is a continuation of the Load Zones feature and includes:

- Ability to add load zones in the generator
- Validations
- Saving changes to the generator file

The remaining work to complete this feature includes:

🛠️ Codegen
🛠️ (possibly) fetch list of load zones via API
🛠️ Conditionally show this feature (and codegen) only when the user is authenticated to Cloud

## How to Test

- enable the `load-zones` feature
- add and remove load zones in Test Options
- check that the "even" distribution works as expected
- check that validation messages work as expected
- check that the changes are saved to the generator file

## Checklist

https://github.com/user-attachments/assets/e3d446bd-f501-4b1a-b8b8-2431bed6f4fd


src/views/Generator/TestOptions/LoadZones/LoadZoneRow.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  // Disable load zone options that are already in use
+  const getLoadZoneOptions = () => {
+    return LOAD_ZONES_REGIONS_OPTIONS.map((option) => ({
+      ...option,
+      disabled: loadZones.some(
+        (zone, i) => zone.loadZone === option.value && i !== index
+      ),
+    }))
+  }
```
cristianoventura, line 37: This ensures that options can't be selected more than once

image

src/views/Generator/TestOptions/LoadZones/LoadZones.tsx
```diff
@@ -48,6 +78,20 @@ export function LoadZones() {
...
+    if (distribution !== 'even') return
+
+    const basePercent = Math.floor(100 / fields.length)
+    const remainder = 100 % fields.length
+
+    fields.forEach((_, index) => {
+      // ensure only integers are used
+      const percent = index < remainder ? basePercent + 1 : basePercent
+      setValue(`loadZones.${index}.percent`, percent)
+    })
```
cristianoventura, line 92: I'm following the same approach used in the cloud where only integers are being used and when "evenly" is selected, floating points are distributed to some items in the array, so not all of them will have the same percentage. For example:

image


instead of al of them having `33.33%`. This makes it easier for users to match and manage the required sum of `100%`.

Llandy3d (approved):
This is really neat, great work 🙌 

Regarding fetching from cloud I think we can create a different ticket for it to tackle after we have that 🤔 

Regarding the error when % are not 100, I wonder if it would be useful to show the current total as well, might make it easier to have the user do less math on their numbers to see how much they are off 🤔

src/views/Generator/TestOptions/LoadZones/LoadZoneRow.tsx
```diff
@@ -0,0 +1,69 @@
...
+            control={control}
+            name={`loadZones.${index}.loadZone`}
+            options={getLoadZoneOptions()}
+          />
+        </FieldGroup>
+      </Table.Cell>
+      <Table.Cell>
+        <FieldGroup errors={errors} name={`loadZones.${index}.percent`} mb="0">
+          <TextField.Root
+            type="number"
```
going-confetti, line 53: adding `min="0"` (or should it be "1"?) should make it a bit harder for the user to enter negative number by mistake (I accidentally did that with by scrolling on my trackpad 😮)

src/views/Generator/TestOptions/LoadZones/LoadZones.tsx
```diff
@@ -59,6 +108,25 @@ export function LoadZones() {
           </RadixLink>
           .
         </Text>
+
+        <FieldGroup name="distribution" label="Distribution" errors={errors}>
```
going-confetti, line 112: This is not a blocking comment, but IMO the test builder UI/copy could be improved. Some thoughts:
- I think the even/manual labels aren't clear enough
- Not sure how I feel about saving both the distribution type and the actual values when distribution is set to `even` - it almost looks like a purely UI setting 🤔. One idea I had was to make the fields editable, but have a "distribute evenly" button, but that would make the UX of adding new zones worse.

going-confetti (approved):
LGTM!

cristianoventura:
> I wonder if it would be useful to show the current total as well
@Llandy3d I think that's a great idea! I'll implement that on the next branch.

> have a "distribute evenly" button
@going-confetti I like this idea actually! I'll ask Maja and team for feedback.

Issue #515, opened 2025-03-03
Improve UX when Chrome is not detected


Llandy3d:
As of now when you try start a new recording if Chrome is missing your are allowed to press the start recording button and the user is greeted with an error message

<img width="698" alt="Image" src="https://github.com/user-attachments/assets/678f8457-43bd-4b64-bf71-66245acbe7a5" />

We should improve the experience by checking ahead of time if Chrome is detectable and if not we should guide the user to get it installed or specify the path in the settings.
We can also consider disabling the start recording button 🤔

PR #523, opened 2025-03-05
fix: It's possible to start a recording with no supported browser installed

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/515

## How to Test
For the browser stuff, mock the handler in main.
To test the proxy warning, kill the proxy process.

Screenshot 2025-03-05 at 14 57 32
Screenshot 2025-03-05 at 14 53 40
Screenshot 2025-03-05 at 14 53 19

### DIFF
--- a/src/components/Settings/RecorderSettings.tsx
+++ b/src/components/Settings/RecorderSettings.tsx
@@ -57,7 +57,7 @@ export const RecorderSettings = () => {
         name="recorder.browserPath"
         onSelectFile={handleSelectFile}
         buttonText="Select executable"
-        hint="The location of the browser executable (Grafana k6 Studio currently supports Chrome)"
+        hint="The location of the browser executable (Google Chrome needs to be installed on your machine for the recording functionality to work)"
         disabled={recorder.detectBrowserPath}
       />
     </SettingsSection>
--- a/src/components/Settings/RecorderSettings.tsx
+++ b/src/components/Settings/RecorderSettings.tsx
@@ -57,7 +57,7 @@ export const RecorderSettings = () => {
         name="recorder.browserPath"
         onSelectFile={handleSelectFile}
         buttonText="Select executable"
-        hint="The location of the browser executable (Google Chrome needs to be installed on your machine for the recording functionality to work)"
+        hint="Google Chrome needs to be installed on your machine for the recording functionality to work"
         disabled={recorder.detectBrowserPath}
       />
     </SettingsSection>
--- a/src/hooks/useSettings.ts
+++ b/src/hooks/useSettings.ts
@@ -28,6 +28,6 @@ export function useSaveSettings(onSuccess?: () => void) {
 export function useBrowserCheck() {
   return useQuery({
     queryKey: ['settings', 'browserCheck'],
-    queryFn: window.studio.ui.checkBrowserPath,
+    queryFn: window.studio.ui.detectBrowser,
   })
 }
--- a/src/hooks/useSettings.ts
+++ b/src/hooks/useSettings.ts
@@ -24,3 +24,10 @@ export function useSaveSettings(onSuccess?: () => void) {
     },
   })
 }
+
+export function useBrowserCheck() {
+  return useQuery({
+    queryKey: ['settings', 'browserCheck'],
+    queryFn: window.studio.ui.checkBrowserPath,
+  })
+}
--- a/src/views/Recorder/EmptyState.tsx
+++ b/src/views/Recorder/EmptyState.tsx
@@ -18,6 +18,7 @@ import { useProxyStatus } from '@/hooks/useProxyStatus'
 import { useStudioUIStore } from '@/store/ui'
 import { TextButton } from '@/components/TextButton'
 import { useBrowserCheck } from '@/hooks/useSettings'
+import { ProxyStatus } from '@/types'
 
 interface EmptyStateProps {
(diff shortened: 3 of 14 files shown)


src/views/Recorder/EmptyState.tsx
```diff
@@ -60,23 +89,91 @@ export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
...
+          width="100%"
         >
-          <TextField.Root
-            {...register('url')}
-            placeholder="e.g. quickpizza.grafana.com"
-            autoFocus
-            css={css`
-              flex-grow: 1;
-              border-bottom-right-radius: 0;
-              border-top-right-radius: 0;
```
going-confetti, line 72: 🤦‍♂️ it had square corners all this time, because the button was supposed to be attached to its right side. It's fixed now

src/views/Recorder/EmptyState.tsx
```diff
@@ -60,23 +89,91 @@ export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
...
+            >
+              {isLoading ? <Spinner /> : <DiscIcon />} Start recording
+            </Button>
+          </Flex>
         </FieldGroup>
-        <Button disabled={isLoading} type="submit">
-          <DiscIcon /> Start recording
-        </Button>
+        <WarningMessage
+          isProxyOnline={proxyStatus === 'online'}
```
going-confetti, line 120: There's an interesting side-effect of this change: When the app starts and the proxy is not yet running, this warning will be displayed. I guess a better UX would be showing some like "Proxy is starting" in an info box - what do you think?

going-confetti: @Llandy3d here's an idea:
1. Add another callback to `launchProxy`, something like `onStart`
2. Call it [here](https://github.com/grafana/k6-studio/blob/198ac6dd07fe21e073693ecdfcc77e0c6b07a410/src/proxy.ts#L72)
3. Change `restarting` to `starting`

src/hooks/useSettings.ts
```diff
@@ -24,3 +24,10 @@ export function useSaveSettings(onSuccess?: () => void) {
     },
   })
 }
+
+export function useBrowserCheck() {
+  return useQuery({
+    queryKey: ['settings', 'browserCheck'],
+    queryFn: window.studio.ui.checkBrowserPath,
+  })
+}
```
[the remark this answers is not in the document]
Llandy3d: `detectBrowser` or `detectBrowserPath` ?
We could make this function return directly the browser path 🤔

src/views/Recorder/EmptyState.tsx
Llandy3d: I would vote as well for showcasing the proxy status in an info box, similar on how we do on the sidebar on the lower left.
Might be enough to show the same value 🤔

Llandy3d (commented):
Is it possible the settings directly in the `Recorder` tab when pressing from the error ?

src/hooks/useSettings.ts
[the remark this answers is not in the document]
going-confetti: I thought about this, but then I realized there's no need for the client to know the path. I like `detectBrowser`

src/views/Recorder/EmptyState.tsx
going-confetti: Here's the fix for the initial proxy status being `offline`: https://github.com/grafana/k6-studio/pull/524

src/hooks/useSettings.ts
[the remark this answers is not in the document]
going-confetti: > A possible use case would be to show in the settings pane which browser we have auto-detected for the user

Ah, I actually thought about this some time ago. You're right

src/components/Settings/RecorderSettings.tsx
```diff
@@ -57,7 +57,7 @@ export const RecorderSettings = () => {
         name="recorder.browserPath"
         onSelectFile={handleSelectFile}
         buttonText="Select executable"
-        hint="The location of the browser executable (Grafana k6 Studio currently supports Chrome)"
+        hint="The location of the browser executable (Google Chrome needs to be installed on your machine for the recording functionality to work)"
```
cristianoventura, line 60: We're testing Linux with arm64, which doesn't support Chrome but uses Chromium instead. At some point we'll have to conditionally change the browser name here.

https://github.com/grafana/k6-studio/pull/513/files#diff-e84e938a5120632cd9bfda48f378416139afdb46dbb35444370bae86a6f1c248R25-R27

src/views/Recorder/EmptyState.tsx
cristianoventura: +1 for having the proxy status in an info box

cristianoventura (commented):
I noticed the alignment of the "Select executable" button on the Settings dialog is off in this branch. But other than that all the changes look good and are a great addition to the UX.

image

src/components/Settings/RecorderSettings.tsx
going-confetti: IMO we either officially support Chromium or we don't and looks like we need to officially support it, in which case it should say `Chromium or Google Chrome` as one shouldn't have to have Chrome installed on Mac either

Llandy3d (approved):
🚀 🚀

PR #528, opened 2025-03-06
feat: Add a way to import data files directly from Test Data popover

going-confetti:
## Description
* Add a way to import data files directly from the Test Data popover
* Show a toast on success

## How to Test
Try importing a data file from the test data popover

image


Llandy3d (approved):
LGTM! 🚀

Issue #533, opened 2025-03-07
Add support for Chromium


cristianoventura:
Currently, our application only supports Chrome as the default browser. However, we need to extend this support to also work with Chromium if Chrome is not found on the system. In case of Linux arm64 where Chrome isn't supported, Chromium should be the default.

Issue #525, opened 2025-03-05
Make it possible to open Settings with specific tab selected


going-confetti:
Currently, it's not possible to open the settings dialog from code outside of the settings button component.
This will change in https://github.com/grafana/k6-studio/pull/523

The next improvement would be having the option to specify a specific tab so that the user wouldn't have to make extra clicks.

PR #535, opened 2025-03-07
feat: add support for Chromium on all platforms

cristianoventura:
## Description

This PR extends the Linux PR https://github.com/grafana/k6-studio/pull/513 and makes Chromium the second default browser for all platforms.

## How to Test

- Install Chromium
- Change the order of browser detection in `browser.ts` to make sure Chromium is detected
- On Windows, Chromium needs to be available in the system's `$PATH` for it to be detected. Once it's added, restart your terminal and relaunch the application `npm start`

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/533

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -13,6 +13,7 @@ import { appSettings } from './main'
 import { exec, spawn } from 'child_process'
 import { getPlatform } from './utils/electron'
 import log from 'electron-log/main'
+import { promisify } from 'util'
 
 const createUserDataDir = async () => {
   return mkdtemp(path.join(os.tmpdir(), 'k6-studio-'))
@@ -21,19 +22,21 @@ const createUserDataDir = async () => {
 export async function getBrowserPath() {
   const { recorder } = appSettings
 
-  if (recorder.detectBrowserPath) {
-    if (getPlatform() === 'linux' && process.arch === 'arm64') {
-      // Chrome is not available for arm64, use Chromium instead
-      return await getChromiumPath()
-    }
-
-    return computeSystemExecutablePath({
-      browser: Browser.CHROME,
-      channel: ChromeReleaseChannel.STABLE,
-    })
+  if (!recorder.detectBrowserPath) {
+    return recorder.browserPath || ''
   }
 
-  return recorder.browserPath as string
+  const chromePath = getChromePath()
+  if (chromePath) {
+    return chromePath
+  }
+
+  const chromiumPath = await getChromiumPath()
+  if (chromiumPath) {
+    return chromiumPath
+  }
+
+  throw new Error('Could not detect Chrome or Chromium browser')
 }
 
 export const launchBrowser = async (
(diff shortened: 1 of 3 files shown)


going-confetti (commented):
I installed Chromium with homebrew and got this error:
image

image

I can't launch the app either, so it's not a studio problem, but I believe we should handle it gracefully
image

src/browser.ts
```diff
@@ -82,31 +90,49 @@ export const launchBrowser = async (
...
   }
 
   // macOS & windows
-  return launch({
+  const browserProc = launch({
     executablePath: path,
     args: args,
     onExit: sendBrowserClosedEvent,
   })
+  browserProc.nodeProcess.on('error', sendBrowserLaunchFailedEvent)
```
cristianoventura, line 104: `launch` uses `spawn` under the hood, which handles errors asynchronously, so try/catch doesn't work here and we need to listen on `error` instead.

cristianoventura (commented):
> I installed Chromium with homebrew and got this error: image

@going-confetti I've made some changes in the browser process to catch the exception and used a listener to show a toast in the frontend. I'll pick up https://github.com/grafana/k6-studio/issues/525 next so we can make the settings dialog more accessible.

image

going-confetti (dismissed):
LGTM!

e-fisher (dismissed):
Tested with chromium on Windows - had to specify browser executable manually, but otherwise worked as expected 👌

Llandy3d:
> Tested with chromium on Windows - had to specify browser executable manually, but otherwise worked as expected 👌

Would be interesting to understand why it fails since we are trying to autodetect it there as well 🤔

cristianoventura:
> > Tested with chromium on Windows - had to specify browser executable manually, but otherwise worked as expected 👌
> 
> Would be interesting to understand why it fails since we are trying to autodetect it there as well 🤔

@Llandy3d On Windows, we use the command `where chrome` to locate the executable. However, this command may not work if the path to the Chromium installation isn't included in the system's `$PATH` environment variable. This can happen depending on how Chromium was installed.

We could add instructions to README or the docs to make this more clear. How does that sound?

Llandy3d:
> > > Tested with chromium on Windows - had to specify browser executable manually, but otherwise worked as expected 👌
> > 
> > 
> > Would be interesting to understand why it fails since we are trying to autodetect it there as well 🤔
> 
> @Llandy3d On Windows, we use the command `where chrome` to locate the executable. However, this command may not work if the path to the Chromium installation isn't included in the system's `$PATH` environment variable. This can happen depending on how Chromium was installed.
> 
> We could add instructions to README or the docs to make this more clear. How does that sound?

thanks for explaining 🙌 
I think we are good with docs since we already mention how to set a browser path manually, we detecting that or not as of now is an implementation detail that might change 🤔

Llandy3d (approved):
🚀 🚀 🚀

PR #537, opened 2025-03-10
feat: migrate generator extension json -> k6g

Llandy3d:
## Description

The extension for generator files has been changed from `.json` to `.k6g`.

To facilitate migrations, json files in the Generator folder are automatically transformed into k6g

## How to Test

Open the app, check that your old Generators now have the `.k6g` extension.
**note:** you might want to backup your project folder until this gets merged

## Checklist

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -120,16 +120,22 @@ initializeLogger()
 // Used to convert `.json` files into the appropriate file extension for the Generator
 async function migrateJsonGenerator() {
   const items = await readdir(GENERATORS_PATH, { withFileTypes: true })
-  const files = items.filter((f) => f.isFile() && path.extname(f.name) === '.json')
+  const files = items.filter(
+    (f) => f.isFile() && path.extname(f.name) === '.json'
+  )
 
   await Promise.all(
     files.map(async (f) => {
-      const oldPath = path.join(GENERATORS_PATH, f.name)
-      const newPath = path.join(
-        GENERATORS_PATH,
-        path.parse(f.name).name + '.k6g'
-      )
-      await rename(oldPath, newPath)
+      try {
+        const oldPath = path.join(GENERATORS_PATH, f.name)
+        const newPath = path.join(
+          GENERATORS_PATH,
+          path.parse(f.name).name + '.k6g'
+        )
+        await rename(oldPath, newPath)
+      } catch (error) {
+        log.error(error)
+      }
     })
   )
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -123,13 +123,13 @@ async function migrateJsonGenerator() {
   const files = items.filter((f) => f.isFile() && path.extname(f.name) === '.json')
 
   await Promise.all(
-    files.map((f) => {
+    files.map(async (f) => {
       const oldPath = path.join(GENERATORS_PATH, f.name)
       const newPath = path.join(
         GENERATORS_PATH,
         path.parse(f.name).name + '.k6g'
       )
-      return rename(oldPath, newPath)
+      await rename(oldPath, newPath)
     })
   )
 }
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -117,6 +117,24 @@ if (require('electron-squirrel-startup')) {
 
 initializeLogger()
 
+// Used to convert `.json` files into the appropriate file extension for the Generator
+async function migrateJsonGenerator() {
+  const files = (
+    await readdir(GENERATORS_PATH, { withFileTypes: true })
+  ).filter((f) => f.isFile() && path.extname(f.name) === '.json')
```
allansson, line 124: Separating the call to `readdir` from the filtering really helps with the readability.

```suggestion
  const items = await readdir(GENERATORS_PATH, { withFileTypes: true })
  const files = items.filter((f) => f.isFile() && path.extname(f.name) === '.json')
```

Then again, I suppose we don't expect this to live long after GA.

src/main.ts
```diff
@@ -117,6 +117,24 @@ if (require('electron-squirrel-startup')) {
...
+
+  await Promise.all(
+    files.map(async (f) => {
+      const oldPath = path.join(GENERATORS_PATH, f.name)
+      const newPath = path.join(
+        GENERATORS_PATH,
+        path.parse(f.name).name + '.k6g'
+      )
+      await rename(oldPath, newPath)
```
allansson, line 133: You technically don't need to use async/await here.

```suggestion
    files.map((f) => {
      const oldPath = path.join(GENERATORS_PATH, f.name)
      const newPath = path.join(
        GENERATORS_PATH,
        path.parse(f.name).name + '.k6g'
      )
      return rename(oldPath, newPath)
```

Maybe I'm just old fashioned. 😅

src/main.ts
Llandy3d: Nothing against making things more readable :)

src/main.ts
Llandy3d: fair feedback, I actually wondered why I couldn't just await things here but If we make use of `Promise.all` then I guess returning the promise is fine.

It does seem confusing to see a normal function returning a normal value (promise in this case), it's less explicit than having pure async but I might just not be used to the javascript way 😄

e-fisher (dismissed):
LGTM 🚀

cristianoventura (approved):
LGTM!

src/main.ts
```diff
@@ -117,6 +117,23 @@ if (require('electron-squirrel-startup')) {
 
 initializeLogger()
 
+// Used to convert `.json` files into the appropriate file extension for the Generator
+async function migrateJsonGenerator() {
+  const items = await readdir(GENERATORS_PATH, { withFileTypes: true })
+  const files = items.filter((f) => f.isFile() && path.extname(f.name) === '.json')
```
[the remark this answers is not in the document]
Llandy3d: We are the ones generating the file names, if someone wants to mess with it it's on them 😄

src/main.ts
```diff
@@ -117,6 +117,23 @@ if (require('electron-squirrel-startup')) {
...
+
+  await Promise.all(
+    files.map((f) => {
+      const oldPath = path.join(GENERATORS_PATH, f.name)
+      const newPath = path.join(
+        GENERATORS_PATH,
+        path.parse(f.name).name + '.k6g'
+      )
+      return rename(oldPath, newPath)
```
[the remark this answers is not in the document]
Llandy3d: I can revert that change.
Ideally we would ask and detect files to be migrated and ask the user for confirmation but since we are going from public preview to GA this is a short-lived thing to make the transition easier.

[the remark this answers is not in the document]
allansson: Or like we did in the olden days: `return rename(oldPath, newPath).catch(() => {})`. Looks cleaner than an entire `try-catch` block, imo.

cristianoventura (approved):

going-confetti (approved):

PR #539, opened 2025-03-10
feat: Run generated scripts in Grafana Cloud k6

allansson:
## Description

This PR adds support for running scripts generated by k6 Studio in Grafana Cloud k6.

> ### ~Due to some issues with the APIs for uploading archives, the test will be Aborted by system. A fix should be released soon™.~ Fix was deployed.

### From validator

1. Sign out of any stacks
2. Export a script
3. Open the script
4. Click the run in cloud button
5. Follow through with the login
6. Verify that the test run is started.
7. While still logged in, repeat steps 3 - 6.

### From generator

1. Open a generator.
2. Go to Script Preview tab
3. Click the run in cloud button.
4. Verify that the test run is started.

## Checklist


src/utils/k6Client.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  constructor(code: number | null, stderr: string[]) {
+    super('Failed to archive script')
+
+    this.code = code
+    this.stderr = stderr
+  }
+}
+
+export class K6Client {
```
allansson, line 66: This is intended to be a typed interface to the k6 CLI. The commands and their options should reflect the options that can be passed to the CLI.

src/views/Generator/GeneratorTabs/ScriptPreview.tsx
```diff
@@ -18,27 +26,32 @@ export function ScriptPreview() {
 
   return (
     <Flex direction="column" height="100%">
-      <Flex p="2" gap="2" justify="end">
-        <Button
+      <Flex py="1" px="2" gap="2" align="center" justify="end">
+        <GhostButton
```
allansson, line 30: I changed these to ghost buttons because

1. it makes the run in cloud button more prominent
2. the soft variant just doesn't look good

src/views/Validator/Validator.tsx
```diff
@@ -155,6 +163,16 @@ export function Validator() {
           />
         }
       />
+      {scriptPath !== undefined && (
```
allansson, line 166: I'm not entirely sure why `scriptPath` would ever be `undefined`. If you're viewing a script in the validator shouldn't it always have a path?

allansson (commented):
Self-review

src/views/Validator/Validator.tsx
Llandy3d: If you open the validator from the home page it will be empty as you have to select a script

src/components/RunInCloudDialog/states/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      return <LoadingMessage>Initializing...</LoadingMessage>
+
+    case 'sign-in':
+      return <SignIn />
+
+    case 'preparing':
+    case 'uploading':
+    case 'starting':
+    case 'started':
+      return <LoadingMessage>Uploading...</LoadingMessage>
```
Llandy3d, line 24: Should we add more visual niceties to the window ?
We could consider adding a grot and/or showing some info on where we are uploading ? (project id, etc..) 🤔

allansson: Maybe this one?

image

src/views/Validator/Validator.tsx
allansson: I see. 

The UX could definitely improve there. It's not very helpful that I'm dropped in to an empty view where the only indication of what I can use it for is to import an external script (which we currently have pretty poor support for).

The view should be more guiding. Ask the user to select a script, encourage them to generate scripts with a generator and allow them to import a script. It would be the validator equivalent of the start recording page, which doesn't just show you an empty recording. It acts as a bit of a guide to start a recording.

src/components/RunInCloudDialog/states/index.tsx
allansson: I added the grot to the dialog. It's a bit more fun now.

Tried including the project name but the text just becomes too wide. Besides, since we open the run they can already see which project they're in. So I would prefer to merge as-is and maybe revisit this when we discuss having configurable projects.

Llandy3d (approved):
Really great work 🙌  🚀 🚀 🚀

e-fisher (approved):
Works as expected! 💯 

We get "Uploaded" run items additionally to the runs themself, but I'm guessing this will need to be addressed on GCk6 side

screencapture 2025-03-12 at 13 17 43

Issue #541, opened 2025-03-10
Implement keyboard shortcuts for saving generator


cristianoventura:
To enhance user experience, we'd like to add keyboard shortcuts to allow users to save their generators. This feature will support Cmd+S for macOS users and Ctrl+S for Windows and Linux users. Additionally, an indicator (e.g., an asterisk *) should appear next to the generator name to denote unsaved changes.

PR #550, opened 2025-03-11
feat: Implement keyboard shortcuts for saving generator

cristianoventura:
## Description

This PR implements logic to save the generator using the cmd+s / ctrl+s shortcut.

## How to Test

- Open a generator
- Make some changes such as enabling/disabling rules
- Save it using cmd+s or ctrl+s

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/541

### DIFF
--- a/src/hooks/useSaveShortcut.ts
+++ b/src/hooks/useSaveShortcut.ts
@@ -1,0 +1,20 @@
+import { useEffect } from 'react'
+
+const useSaveShortcut = (saveCallback: () => Promise<void>) => {
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if ((event.ctrlKey || event.metaKey) && event.key === 's') {
+        event.preventDefault()
+        return saveCallback()
+      }
+    }
+
+    window.addEventListener('keydown', handleKeyDown)
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown)
+    }
+  }, [saveCallback])
+}
+
+export default useSaveShortcut
--- a/src/views/Generator/Generator.tsx
+++ b/src/views/Generator/Generator.tsx
@@ -21,6 +21,7 @@ import { getFileNameWithoutExtension } from '@/utils/file'
 import log from 'electron-log/renderer'
 import { ProxyData } from '@/types'
 import { Details } from '@/components/WebLogView/Details'
+import useSaveShortcut from '@/hooks/useSaveShortcut'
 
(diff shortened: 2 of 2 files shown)


src/hooks/useSaveShortcut.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
+
+const useSaveShortcut = (saveCallback: () => Promise<void>) => {
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if ((event.ctrlKey || event.metaKey) && event.key === 's') {
```
cristianoventura, line 6: Ctrl+S will also work on macOS. Detecting the platform would require touching the main process, so for now, I’ve opted to leave it as is. We can revisit this when our keybinding system expands and needs more granular control.

src/views/Generator/Generator.tsx
```diff
@@ -125,7 +128,11 @@ export function Generator() {
   return (
     <View
       title="Generator"
-      subTitle={getFileNameWithoutExtension(fileName)}
+      subTitle={
+        <>
+          {getFileNameWithoutExtension(fileName)} {isDirty && '(edited)'}
```
cristianoventura, line 133: I considered using icons instead of appending `(edited)` to the title but I didn't think it was as clear as this. Open to suggestions.

image

src/hooks/useSaveShortcut.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
+
+const useSaveShortcut = (saveCallback: () => Promise<void>) => {
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
```
going-confetti, line 5: We should've probably documented it somewhere, but we have `keyboardjs` in our dependencies and it makes it much easier to manage shortcuts (here's an example: https://github.com/grafana/k6-studio/pull/177)

src/views/Generator/Generator.tsx
cristianoventura: I ended up replacing it with `*` for simplicity.

going-confetti (dismissed):
LGTM. One small thing that I noticed is that it's possible to save a generator with the shortcut even if it's not dirty. I don't think it's a big deal

cristianoventura:
@going-confetti good catch - I just addressed that bug!

going-confetti (approved):
👏

PR #552, opened 2025-03-12
style: Remove dots from short tooltips

going-confetti:


cristianoventura (dismissed):

src/components/StaticAssetsFilter.tsx
```diff
@@ -17,9 +17,7 @@ export function StaticAssetsFilter({
 
   return (
     <Label>
-      <Tooltip content="Static assets are excluded from your test by default.">
```
e-fisher, line 20: Should we have a more detailed tooltip instead? Could be something like:

Static assets are files like images, CSS, JavaScript, fonts, and videos that don't change frequently and are typically cached by browsers. Excluding them from performance tests is usually recommended because their performance is more dependent on CDN/hosting rather than your application code.

Enable this only when specifically testing static asset delivery or when your application heavily depends on asset loading behavior.

going-confetti: I believe static assets is a fairly common term that doesn't need explaining to our target audience. It might make sense to list all content types we consider static/non-static to make it clearer to the user 🤔 

As for explaining why it's important, I'd rather this stay in the docs. We might add in-app onboarding experience at some point - that could also be a good place for this information.

Llandy3d: It really depends on the "depthness" that we want to go with, I wouldn't assume users knowing the term long term, most likely even just listing something in the lines of `Static assets are files like images, CSS, JavaScript, fonts, and videos` could be already useful 🤔

Llandy3d (approved):

src/components/StaticAssetsFilter.tsx
going-confetti: I'm not fully convinced we should do it. The importance of excluding static assets is covered in the [generator doc](https://grafana.com/docs/k6-studio/components/test-generator/#allowed-hosts) and even this article doesn't explain what static assets are.

If we decide to do it, we should be very explicit and probably use `NON_STATIC_ASSET_MIME_TYPES` and `NON_STATIC_ASSET_EXTENSIONS` to generate the copy (it surprised me that we consider static html files to be non-static assets 🤔)

Llandy3d: For docs we most likely should improve on them 🤔 

As to why possible mention in-app on tooltip/info icon what are the static assets, it would be for users that don't know the terminology and could be confused by it. We don't want to exclude new users learning performance testing from the equation 🤔 
I would also assume that an expert user wouldn't even try to over of that input to see what it is and just go straight to toggle it

going-confetti: > it would be for users that don't know the terminology and could be confused by it. We don't want to exclude new users learning performance testing from the equation

We should probably discuss this again with Product and Design. I'm not sure the second sentence is correct at the current stage (and if it is - we have docs).

PR #554, opened 2025-03-12
feat: Configurable verification rule

e-fisher:
## Description

Add ability to create custom verification rules.

Supported comparison value:
- Recorded value
- Text value
- Variable

Supported operators:
- Equals
- Contains
- Does not contain

I've also refactored `RequestSnippetSchema` to allow create single `check` call when multiple verification rules match same request. 

When comparing body with recorded response I'm adding the whole body into check, this seems to work fine, although it can grow the script file and make it harder to read. In future, it would be ideal to store response bodies in separate file and import. 

## How to Test

Test with different websites and scenarios - verify script doesn't break and checks fail/pass as expected.

## Checklist

screencapture 2025-03-13 at 12 17 19

screencapture 2025-03-13 at 12 17 35


src/rules/verification.utils.ts
```diff
@@ -0,0 +1,107 @@
...
+) {
+  switch (target) {
+    case 'status':
+      return `'${response.statusCode}'`
+    case 'body': {
+      // Remove newlines when comparing the body to a recorded value
+      const singleLineContent = response.content.replace(NEWLINE_REGEX, '')
+      const escapedContent = escapeBackticksAndDollarSign(singleLineContent)
+
+      return `String.raw\`${escapedContent}\``
```
e-fisher, line 67: Ideally, we would save responses in separate file, import and reference them in checks. This would add some additional challenges managing the imports, as the first iteration I'm adding the whole response body as a single-line string. 

This might be easier to solve when/if we switch to AST 🌲

src/rules/verification.utils.ts
```diff
@@ -0,0 +1,107 @@
...
+    default:
+      return exhaustive(target)
+  }
+}
+
+function getTarget(rule: VerificationRule) {
+  switch (rule.target) {
+    case 'status':
+      // Convert to string to support .includes() for status and body
+      return 'r.status.toString()'
```
e-fisher, line 78: Converting status to string simplifies the logic - it allows doing both equal and contains comparison, as well as avoids converting variable to number (we only support string now).

src/schemas/generator/v0/index.ts
```diff
@@ -30,18 +30,26 @@ export function migrate(generator: GeneratorSchema): v1.GeneratorSchema {
...
       }
 
+      if (rule.type === 'verification') {
+        return {
+          ...rule,
+          target: 'status',
+          operator: 'equals',
+          value: { type: 'recordedValue' },
+        }
+      }
```
e-fisher, line 48: Verification rule type wasn't set correctly (optional `selector`), but UI restricted creation of invalid rules, so we don't need a new migration.

src/views/Generator/RuleEditor/VerificationEditor/ValueEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+    { value: 'variable', label: 'Variable', disabled: !hasVariables },
+  ]
+
+  const handleToggleValueType = (value: VerificationRule['value']['type']) => {
+    if (value === 'recordedValue') {
+      setValue('operator', 'equals')
+    }
+
+    setValue('value.type', value)
+  }
```
e-fisher, line 44: Only support `equals` operator when comparing to recorded values.

going-confetti (changes requested):
Added a few comments. I think statuses shouldn't only be compared to numeric values and contains/does not contain checks shouldn't be supported for them (not sure if using variables makes sense either, but that's probably on the user 🤔)

src/rules/verification.utils.ts
Llandy3d: Might also need some validation (although we get the validation for free in the editor) 🤔 

image

e-fisher:
@going-confetti @Llandy3d thank you for the feedback! I've addressed the change requests, supported comparisons:

- status
  - recorded value: equals
  - number: equals | not equals
- body
  - recorded value: equals
  - text value: equals | not equals | contains | not contains
  - variable: equals | not equals | contains | not contains

No wildcard support for status checks for now

going-confetti (approved):
LGTM 🚢

src/codegen/codegen.test.ts
```diff
@@ -423,7 +425,7 @@ describe('Code generation', () => {
         params = { headers: {}, cookies: {} }
         url = http.url\`http://test.k6.io/api/v1/foo\`
         resp = http.request('POST', url, null, params)
-        check(resp,{'statusmatches200':(r)=>r.status===200,})
+        check(resp, { 'status equals recorded value': (r) => r.status === 200 })
```
Llandy3d, line 428: we discussed in the past that the check is more useful when it showcases what it's checking against, I think we should probably leave it like before `status equals 200`

going-confetti: What if the recorded value has a different status (e.g. 201)?

Llandy3d: that part is generated so if the recorded value is `201` the check name would be as before: `status equals 201`

Llandy3d: basically: `status equals <value>: ... r.status === <value>`

Llandy3d (dismissed):
🙌

src/schemas/generator/index.ts
```diff
@@ -1,18 +1,22 @@
...
   v0.GeneratorFileDataSchema,
   v1.GeneratorFileDataSchema,
+  v2.GeneratorFileDataSchema,
 ])
 
 export function migrate(generator: z.infer<typeof AnyGeneratorSchema>) {
   switch (generator.version) {
     case '0':
-      return migrate(v0.migrate(generator))
+      return migrate(v1.migrate(v0.migrate(generator)))
```
going-confetti, line 16: `migrate` is a recursive function, so each step should only migrate the generator to the next version;
```ts
case '0':
  return migrate(v0.migrate(generator))
case '1.0':
  return migrate(v1.migrate(generator))
```

src/schemas/generator/v2/index.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from 'zod'
+import { TestRuleSchema } from './rules'
+import { TestDataSchema } from './testData'
+import { TestOptionsSchema } from './testOptions'
+
+export const GeneratorFileDataSchema = z.object({
+  version: z.literal('2.0'),
```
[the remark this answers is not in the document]
Llandy3d: good catch, if we don't have a specified reason for going with a fractional we could just set on int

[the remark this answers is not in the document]
e-fisher: Since every migration would a breaking, we can probably just do "2"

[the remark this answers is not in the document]
Llandy3d: Although there could be an argument for it having the `2.0` fractional value, for example a `2.1` could be good to document a version that added something but doesn't break the schema.
Because of that I would vote to keep the fractional part since we might have future uses for it

going-confetti (approved):

src/schemas/generator/v2/index.ts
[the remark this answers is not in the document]
cristianoventura: We had a brief discussion once about making the migrations timestamp-based instead of numeric like other major frameworks do. It'll provide a clearer and more predictable way of ordering migrations since it'll always migrate to the latest version (and prevent conflicts if multiple people are working on them) while also aligning the project with more standard practices. I'd love to continue this discussion at some point.

@Llandy3d @e-fisher @going-confetti

Issue #562, opened 2025-03-14
It's not possible to select browser executable on macOS


going-confetti:
### Steps to reproduce the problem

1. Open Settings - Recorder
2. Un-check "Automatically detect browser"
3. Select Google Chrome
4. Save changes
5. Try starting a recording

<img width="780" alt="Image" src="https://github.com/user-attachments/assets/b2d37abb-2701-474d-bdc1-334a24c32e92" />

### Expected behaviour

- Correct path should be set (`Google Chrome.app` is not executable)
- Recorder should work
- Nice to have: show a warning if the app name isn't supported or find a different way to validate the selection

<img width="786" alt="Image" src="https://github.com/user-attachments/assets/57c8dd4e-1de0-4a6a-9390-d6da05e4a92d" />

<img width="801" alt="Image" src="https://github.com/user-attachments/assets/c5af3eeb-572d-4912-aafb-3b7abe575e65" />

It's _probably_ safe to assume the executable has the same name as the app, but it might be worth getting it from the `/Contents/Info.plist` file

### Actual behaviour

Recorder doesn't work anymore and throws the following exception:

<img width="372" alt="Image" src="https://github.com/user-attachments/assets/f5970596-411e-43b3-9a8c-5f3ee53fba33" />

### Grafana k6 Studio version

0.14.0

### OS

macOS 15.3.1

Issue #501, opened 2025-02-19
Spawn error A JavaScript error occurred in the browser process


AlexeiOm:
### Brief summary

The same issue as in #467 and #486 
Appears immediately on start of a recording. Intel mac

### Grafana k6 Studio version

0.14.0 x64 mac

### OS

macOS 15.0.1

### Steps to reproduce the problem

https://github.com/user-attachments/assets/4539d0e7-257d-4643-b8c9-ac7e77ae4158

### Expected behaviour

recording started

### Actual behaviour

spawn error

AlexeiOm:
Stacktrace

Error: spawn /Applications/Yandex.app EACCES
at ChildProcess._handle.onexit (node:internal/child_process:286:19)
at onErrorNT (node:internal/child_process:484:16)
at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

Llandy3d:
Seems like it's detecting the application `Yandex.app` as the browser but cannot be launched due to permissions errors 🤔 

The application currently has official support for `Chrome`, could you try to configure it in the settings by specifying the path instead of relying to the automatic detector ?

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/a2189852-a48c-417a-857a-21cdeb270fbb" />

AlexeiOm:
Chrome is actually my default browser. It is all the same no matter what url do I choose. Probably yandex.app is a non-existing app, I don't know that one.

https://github.com/user-attachments/assets/14f8cf4e-f5f2-4795-b0f1-c604c077e13b

AlexeiOm:
I see, it does not warn within the first start whether to choose a browser as an executable. But it changes nothing. The same error with chrome app
Error: spawn /Applications/Google Chrome.app EACCES
at ChildProcess._handle.onexit (node:internal/child_process:286:19)
at onErrorNT (node:internal/child_process:484:16)
at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

Llandy3d:
> I see, it does not warn within the first start whether to choose a browser as an executable. But it changes nothing. The same error with chrome app Error: spawn /Applications/Google Chrome.app EACCES at ChildProcess._handle.onexit (node:internal/child_process:286:19) at onErrorNT (node:internal/child_process:484:16) at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

[...]

PR #576, opened 2025-03-18
fix: It's not possible to select browser executable on macOS

cristianoventura:
## Description

This PR resolves an issue that prevents selecting the correct executable browser path on macOS from the settings page.

## How to Test

- open Settings > Recorder
- select a custom browser path
- notice that the correct executable path was populated

## Checklist

https://github.com/user-attachments/assets/ec6c2cdf-a8cb-4051-9998-b8b2eadae0bc

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/562 https://github.com/grafana/k6-studio/issues/501


src/components/Settings/RecorderSettings.tsx
```diff
@@ -60,6 +74,21 @@ export const RecorderSettings = () => {
...
+        !isValidPath(recorder.browserPath?.toLocaleLowerCase()) && (
+          <Callout.Root color="amber">
+            <Callout.Icon>
+              <ExclamationTriangleIcon />
+            </Callout.Icon>
+
+            <Callout.Text>
+              The selected executable doesn&apos;t appear to be compatible.
+              Please select the correct executable for Chrome or Chromium.
+            </Callout.Text>
```
cristianoventura, line 89: This message does not really detect that the selected browser is correct. In practice you could rename the Firefox executable to `Chrome` and it would bypass it, but it handles the most common scenarios and warns the user about incompatibilities.

src/settings.ts
```diff
@@ -130,11 +131,31 @@ export async function selectBrowserExecutable() {
...
+
+  function getFilePaths() {
+    if (getPlatform() === 'mac') {
+      return filePaths.map((filePath) => {
+        const plistPath = path.join(filePath, 'Contents', 'Info.plist')
+        const executableName = getExecutableNameFromPlist(plistPath)
+        if (executableName) {
+          return path.join(filePath, 'Contents', 'MacOS', executableName)
+        }
+        return filePath
```
cristianoventura, line 148: Here we read the executable name from `Info.plist`, if available.

src/utils/plist.ts
```diff
@@ -0,0 +1,18 @@
+import { readFileSync } from 'fs'
+import { parse } from 'plist'
+import log from 'electron-log/main'
+
+export function getExecutableNameFromPlist(plistPath: string) {
+  const plistContent = readFileSync(plistPath, 'utf-8')
+  const plistData = parse(plistContent) as Record<string, unknown>
+
+  if (typeof plistData.CFBundleExecutable === 'string') {
+    return plistData.CFBundleExecutable
```
cristianoventura, line 10: [CFBundleExecutable](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleexecutable) is the default location for the executable name

src/components/Settings/RecorderSettings.tsx
```diff
@@ -32,6 +33,19 @@ export const RecorderSettings = () => {
...
+  const isValidPath = (path?: string) => {
+    if (!path) return false
+    const validPaths = [
+      'chrome.app',
+      'chromium.app',
+      'chrome.exe',
+      'chromium.exe',
+      '/chrome',
+      '/chromium',
+    ]
```
Llandy3d, line 45: We should probably include chromium derived browsers, for example Edge 🤔 
I wonder if there's a better way to detect a working browser, maybe like doing a test run in the background and checking that a request is actually recorded, might be too laborious thou 😄

going-confetti: > We should probably include chromium derived browsers, for example Edge 🤔

I wouldn't do it at this stage. If the user wants to try another Chromium-based browser, they should do it at their own risk.
As far as I understand, we don't prevent the user from selecting a different browser, so they can still try to use other browsers.

src/components/Settings/RecorderSettings.tsx
```diff
@@ -60,6 +74,21 @@ export const RecorderSettings = () => {
...
+        recorder.browserPath !== '' &&
+        !isValidPath(recorder.browserPath?.toLocaleLowerCase()) && (
+          <Callout.Root color="amber">
+            <Callout.Icon>
+              <ExclamationTriangleIcon />
+            </Callout.Icon>
+
+            <Callout.Text>
+              The selected executable doesn&apos;t appear to be compatible.
+              Please select the correct executable for Chrome or Chromium.
```
going-confetti, line 88: The hint only mentions Google Chrome. If we release this without the Chromium PR, it might be confusing

going-confetti (approved):
LGTM!

src/components/Settings/RecorderSettings.tsx
cristianoventura: I can make that change in the Chromium support PR once we fully validated it's working for all plaforms https://github.com/grafana/k6-studio/pull/535

src/components/Settings/RecorderSettings.tsx
cristianoventura: We haven't really tested in other Chrome-based browsers and potential ramifications they introduce so I'd say it's safer to just go with Chrome and Chromium for now.

Llandy3d (approved):

PR #584, opened 2025-03-20
chore: Use ExternalLink component for docs and GitHub links

going-confetti:
## Description
Use the ExternalLink component for, well, external links.
This leaves just 2 places where `openExternalLink` is used directly, but since those are in a dropdown menu, I couldn't used this component.

## How to Test
Verify that the links work

### DIFF
--- a/src/ErrorElement.tsx
+++ b/src/ErrorElement.tsx
@@ -9,11 +9,7 @@ import { css } from '@emotion/react'
 
 import GrotCrashed from '@/assets/grot-crashed.svg'
 import { ArrowLeftIcon } from '@radix-ui/react-icons'
-
-const handleCreateIssue = () =>
-  window.studio.browser.openExternalLink(
-    'https://github.com/grafana/k6-studio/issues'
-  )
+import { ExternalLink } from './components/ExternalLink'
 
 const handleOpenLogs = () => {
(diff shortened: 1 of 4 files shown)


src/ErrorElement.tsx
```diff
@@ -9,11 +9,7 @@ import { css } from '@emotion/react'
 
 import GrotCrashed from '@/assets/grot-crashed.svg'
 import { ArrowLeftIcon } from '@radix-ui/react-icons'
-
-const handleCreateIssue = () =>
-  window.studio.browser.openExternalLink(
-    'https://github.com/grafana/k6-studio/issues'
-  )
+import { ExternalLink } from './components/ExternalLink'
```
going-confetti, line 12: I think we should also add an icon to make it a bit clearer that the link will open in the browser, something like the [external link icon in Grafana UI](https://developers.grafana.com/ui/latest/index.html?path=/story/general-textlink--example)

e-fisher (approved):
LGTM

src/ErrorElement.tsx
cristianoventura: That would be a great addition!

cristianoventura (approved):
🧹✨

PR #587, opened 2025-03-21
feat: add logical operator labels

e-fisher:
## Description

Currently, we use symbols (< > ===) for threshold operators and labels for verification rules, this PR makes it consistent by using label and icon combination in both components. 

- Because of limited space in thresholds, only the icon will be shown once value is selected.
- For verification operator, we always show both label and icon. 
- I've included react-icons library to get access to more icons.

screencapture 2025-03-21 at 15 26 17@2x

screencapture 2025-03-21 at 15 26 22@2x

screencapture 2025-03-21 at 15 26 37@2x

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
         "react-hook-form": "^7.53.0",
+        "react-icons": "^5.5.0",
         "react-router-dom": "^6.24.0",
         "react-select": "^5.8.3",
         "react-use": "^17.5.1",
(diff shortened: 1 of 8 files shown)


cristianoventura (dismissed):
LGTM! 🚀

e-fisher:
I've updated the default threshold operator to be "less than" since in most cases you would "expect some value to be less than a threshold", and added 100ms as default value.

Llandy3d (dismissed):
🙌

e-fisher:
Per @going-confetti suggestion, I've replaced react-icons library with tabler icons, since it was the only set of icons used and will reduce risk of using different styled icons in the future.

Issue #318, opened 2024-11-10
Make it possible to rename a file from the view title


Llandy3d:
To rename a file in k6 Studio, you currently have to right-click on it the sidebar and select the rename option.

Consider making the filename in the view header editable, so that the user is more likely to give the resource a meaningful name. This should make the file trees more readable in comparison to the current state where they have lots of files with auto-generated names.

![Image](https://github.com/user-attachments/assets/578a2027-adb1-41bc-8ddc-ac7f36078f1f)

PR #588, opened 2025-03-21
feat: Add a way to rename files from View header

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/318

#### Description
Makes it much easier to rename a file, which should ideally nudge the users to give files names that are more meaningful than the auto generated ones.

#### Testing instructions
* Try renaming all kinds of files (recordings, generators, scripts, data files)


src/components/FileNameHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { useRenameFile } from '@/hooks/useRenameFile'
+import { StudioFile } from '@/types'
+import { getFileExtension } from '@/utils/file'
+
+import { FieldGroup } from './Form'
+
+interface FileNameHeaderProps {
+  file: StudioFile
+  isDirty?: boolean
+  showExt?: boolean
```
going-confetti, line 25: image

src/components/FileNameHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { useOverflowCheck } from '@/hooks/useOverflowCheck'
+import { useRenameFile } from '@/hooks/useRenameFile'
+import { StudioFile } from '@/types'
+import { getFileExtension } from '@/utils/file'
+
+import { FieldGroup } from './Form'
+
+interface FileNameHeaderProps {
+  file: StudioFile
+  isDirty?: boolean
```
going-confetti, line 24: image

src/components/FileNameHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+  showExt?: boolean
+}
+
+export function FileNameHeader({
```
going-confetti, line 28: Not sure if it's the best name, but we only have views that represent files, so I think it's fine

src/views/RecordingPreviewer/RecordingPreviewer.tsx
```diff
@@ -30,6 +31,11 @@ export function RecordingPreviewer() {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
   const isDiscardable = Boolean(state?.discardable)
   invariant(fileName, 'fileName is required')
+  const file: StudioFile = {
+    fileName,
+    displayName: getFileNameWithoutExtension(fileName),
```
going-confetti, line 36: Note: this is getting repetitive and is fairly error-prone - `displayName` is always (at least for now) the same as the file name without the extension, so it can be easily derived from it without the need to keep both values in sync.

src/views/Generator/Generator.tsx
```diff
@@ -143,9 +144,14 @@ export function Generator() {
     <View
       title="Generator"
       subTitle={
-        <>
-          {getFileNameWithoutExtension(fileName)} {isDirty && '*'}
-        </>
+        <FileNameHeader
```
going-confetti, line 147: This also fixes a recent regression that made it impossible to see the full file name if it overflows

src/components/FileTree/File.tsx
```diff
@@ -79,21 +79,16 @@ function EditableFile({
...
 
-  const navigate = useNavigate()
-
-  const hasEllipsis = useOverflowCheck(linkRef)
-  const fileExtension = file.fileName.split('.').pop()
+  const fileExtension = getFileExtension(file.fileName)
 
   const handleSave = async (newValue: string) => {
-    const newFileName = `${newValue}.${fileExtension}`
+    const newFileName = `${newValue.trim()}.${fileExtension}`
```
going-confetti, line 89: Without this, it's easy to have multiple files with different names that look the same (e.g. `generator.k6g` and `generator .k6g`)

src/components/FileNameHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+  )
+}
+
+interface RenameFileDialogProps {
+  file: StudioFile
+  open: boolean
+  onOpenChange: (open: boolean) => void
+}
+
+function RenameFileDialog({ file, open, onOpenChange }: RenameFileDialogProps) {
```
going-confetti, line 86: I wanted to have some kind of an inline text input, but didn't have time to implement it the right way 🫤

Llandy3d (dismissed):
LGTM! 🙌

e-fisher (commented):
Works great! 🙌  Have a few suggestions regarding UI:

1. The dialog seems a bit overkill here, maybe Popover would be a better fit?

screencapture 2025-03-27 at 14 19 49

2. Validation errors are displayed as toasts and a covered with dialog's overlay, inline validation could look better. Although, it becomes less of a problem if we switch to popover.

screencapture 2025-03-27 at 14 08 35

going-confetti:
> Works great! 🙌 Have a few suggestions regarding UI:
> 
>     1. The dialog seems a bit overkill here, maybe Popover would be a better fit?
> 

Looks better indeed 👍 
>     2. Validation errors are displayed as toasts and a covered with dialog's overlay, inline validation could look better. Although, it becomes less of a problem if we switch to popover.
> 

I'll switch to popover

src/components/FileNameHeader.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const onSubmit = async ({ fileName }: { fileName: string }) => {
+    if (!isDirty) return
+
+    await mutateAsync(`${fileName.trim()}.${fileExtension}`)
+    setIsOpen(false)
+  }
+
+  return (
+    <Popover.Root open={isOpen} onOpenChange={setIsOpen}>
```
going-confetti, line 108: image

Llandy3d (approved):

PR #591, opened 2025-03-24
chore: Use text value comparison as default option for body verification rules

e-fisher:
## Description

From internal feedback: in verification rule, comparing response body to recording could lead to massive script, we still want to provide this option, but not as a default when `Response body` is selected as `Target`. Instead, select `Text value` as default.

## How to Test

1. Create a new verification rule
2. Change `Target` to `Response body`
3. Verify `Text value` option is pre-selected.

## Checklist

screencapture 2025-03-24 at 13 59 08

### DIFF
--- a/src/schemas/generator/v2/rules.ts
+++ b/src/schemas/generator/v2/rules.ts
@@ -164,8 +164,8 @@ export const BodyVerificationRuleSchema = BaseVerificationRuleSchema.extend({
   target: z.literal(VerificationTarget.enum.body),
   operator: VerificationOperator,
   value: z.discriminatedUnion('type', [
-    RecordedValueSchema,
     StringValueSchema,
+    RecordedValueSchema,
     VariableValueSchema,
   ]),
 })


going-confetti (commented):
I have two thoughts:
1. We usually remove chores from the release notes, so maybe this should be a `feat` or a `fix`? Though I can see how none of them really fits here.
2. IMO it would make more sense to use `includes` rather than `equals` - it's hard to imagine someone copy-pasting entire body into a small text input (which is an issue worth thinking about on its own - maybe it should use textarea/monaco 🤔)

going-confetti (approved):
LGTM!

e-fisher:
1. Good point on chores being excluded from release notes, I've changed to `fix`. Although it feels like we have a need for an additional label, an "improvement" that's neither a fix nor new feature 🤔 
2. Agree, I've updated to use "contains" as default. A use case for equals would be verifying JSON body, e.g., `{"success":true}`. We'll need to revisit this when we do rule form's revamp, a monaco editor could work here, if we add support for both html and JSON highlighting.

Issue #575, opened 2025-03-18
In-app view of recorded events


allansson:
# What

To make it easier to see and edit your recording, we should add a view for showing the recorded browser events directly in the browser.

# Criteria for considering it done

Users should be able to toggle a view where a list of the recorded events is shown. The list has to be persisted between page navigations.

PR #593, opened 2025-03-24
feat(browser): Inspect events in in-browser UI

allansson:
## Description

This PR adds a view to the in-browser UI to allow users to inspect browser events without switching windows to k6 Studio. 

It shares its implementation with the event list in k6 Studio, so everything that is supported there should work in the new view:

* Hovering a selector should highlight matching elements
* Clicking a link in a navigation event should load that page.

Notes on the implementation:

* I actually wanted the drawer to be a modal so that it closes when you click outside, but I ran in to a lot of problems with it that I just couldn't get around. I will revisit it in a separate PR once I've completed some more urgent features.
* Messaging between the different parts was getting really messy and unmaintainable so this PR includes a rewrite on the communication layer.
* Since `@radix-ui/themes` can't be configured to mount its styles inside a shadow dom, I've had to style my own components. Though we have discussed using another headless UI library, I decided to stick with Radix UI Primitives because of their popularity and their API design.

Since this PR has many changed files (many which were moved _and_ modified), I've put extra effort into the self-review to try and help you understand what changed.

## How to Test

1. Start a recording
2. Record some events
3. Open the event drawer
4. In browser and studio:
    1. Check that the recorded events look the same
    2. Hover selectors and check that elements are highlighted in the browser.
    3. Click links and check that the browser navigates to the link

## Checklist

https://github.com/user-attachments/assets/9e6a89c7-d9be-4d06-bf22-88b61a0978d5

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6-studio/issues/575


extension/src/frontend/view/RemoteHighlights.hooks.ts
allansson, line 1: Moved and modified to use new messaging system.

extension/src/frontend/view/ToolBox.tsx
allansson, line 1: Moved and modified to use `@/components/primitives/Toolbar` and to add event drawer toggle.

extension/src/ui/ContainerProvider.tsx
allansson, line 1: Implemented in `@/components/primitives/ContainerProvider`.

allansson (commented):
Self-review

extension/src/messaging/index.ts
```diff
@@ -0,0 +1,137 @@
+import { z } from 'zod'
```
allansson, line 1: With the latest changes, communication between k6 Studio, background and frontend became quite complicated. Before we only had uni-directional communication (`frontend -> background -> k6 Studio`) but now each part wants to communicate with every other part:

* Background wants to notify both frontend and studio that events have been recorded.
* k6 Studio wants to tell the frontend to highlight an element.
* Frontend asks the background to get the current list of recorded events.
* etc.

I created this messaging system to make it easier to setup this communication.

The `BrowserExtensionClient` class provides a unified interface for every consumer of messages. Messages are sent and received in a type safe manner. Serialization is handled in the same way for every consumer, making it easier to update or add new messages.

Different consumers only need to configure which transport they communicate over. The background script, for example, will communicate with k6 Studio over a WebSocket client whereas k6 Studio will communicate through a WebSocket **_server_**.

We route messages using the `forward` method on the client, so we can tell e.g. the background script to forward the `highlight-element` message to the frontend.

src/views/Recorder/BrowserEventLog.tsx
```diff
@@ -0,0 +1,64 @@
+import { Flex, Button, ScrollArea } from '@radix-ui/themes'
```
allansson, line 1: Moved and modified to use `BrowserEventList`.

src/services/browser/server.ts
```diff
@@ -1,100 +1,29 @@
 import { BrowserWindow } from 'electron'
```
allansson, line 1: This is a good example of how much simpler the implementation is with the new messaging system.

extension/src/frontend/view/ElementInspector.tsx
```diff
@@ -0,0 +1,40 @@
+import { css } from '@emotion/react'
```
allansson, line 1: Moved and modified to use `@/components/primitives/Tooltip`.

extension/src/frontend/view/RemoteHighlights.hooks.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useState } from 'react'
```
allansson, line 1: Moved and modified to use new messaging system.

extension/src/frontend/view/ToolBox.tsx
```diff
@@ -0,0 +1,71 @@
+import { css } from '@emotion/react'
```
allansson, line 1: Moved, modified to use `@/components/primitives/Table` and added the event drawer toggle.

allansson (commented):
Self-review #2

extension/src/frontend/view/EventDrawer.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+                  &:hover {
+                    background-color: var(--studio-hover-color);
+                  }
+                `}
+              >
+                <Cross1Icon />
+              </Dialog.Close>
+            </div>
+            <BrowserEventList
```
going-confetti, line 160: The list is not scrollable, so if it becomes long enough, there's no way to see the most recent events
image

src/vite-env.d.ts
```diff
@@ -4,3 +4,5 @@ declare const K6_API_URL: string
 declare const GRAFANA_CLIENT_ID: string
 declare const GRAFANA_API_URL: string
 declare const GRAFANA_COM_URL: string
+
+declare const TARGET_PLATFORM: NodeJS.Platform
```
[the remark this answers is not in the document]
going-confetti: I just realized that recording tests on one platform and running on another could be tricky, because command-click is pretty much guaranteed to be different from windows-key-click. Another example is recording the `cmd+enter`-to-submit type of flow.

Is this something we'll eventually want to address in the studio? For example, by letting the user set platform-specific alternatives to these modifier keys

going-confetti (changes requested):
Thanks for the self review and in general for making sure this is so well structured 🤩 !

It looks and works great for the most part, with the exception of the bug with the drawer that prevented me from seeing the full list of events.

src/vite-env.d.ts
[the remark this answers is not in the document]
allansson: That's a good point. I hadn't thought about it either. 

This would be problematic if you're running the script in CI-pipeline using a different OS, e.g. Github Actions where using Linux is the most common I think. Same problem would apply to using keyboard shortcuts (though we don't record those yet).

We're probably getting away with it because using the mouse with modifier keys isn't that common. 

I don't even want to think about what happens if you use Option + Click to open in a new tab. 🙈

> Is this something we'll eventually want to address in the studio? For example, by letting the user set platform-specific alternatives to these modifier keys

I don't know. It feels like it's an issue in k6 browser (and browser testing in general). Best way to reliably get around it would be to do a runtime check for the OS and choose the correct mapping from that.

Maybe there's room for a UI to configure various mappings, e.g. Cmd + C <-> Ctrl + C, depending on platform and k6 Studio would emit a utility function that'd execute the right combination based on the OS.

For now though I'm happy to bury my head in the sand and pretend the problem doesn't exist.

extension/src/frontend/view/RemoteHighlights.hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    return client.on('highlight-element', ({ data }) => {
+      if (data.selector === null) {
+        setBounds([])
+
+        return
+      }
+
+      const element = document.querySelectorAll(data.selector)
```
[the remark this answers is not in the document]
allansson: Did you one better and also pluralized the message type: `highlight-element` -> `highlight-elements`.

going-confetti (approved):
LGTM :shipit:

Issue #540, opened 2025-03-10
Pre-select the domain user entered when started recording


cristianoventura:
When a generator is created after a recording, pre-select the domain that the user had entered before starting the recording in the allowlist dialog. This feature will improve user experience by reducing redundant steps.

<img width="466" alt="Image" src="https://github.com/user-attachments/assets/62e8dc63-0ded-47d2-82e8-d49b199e1185" />

Llandy3d:
As a follow up enhancement we should consider having the `host` used for starting the recording as the first element in the list in a "sticky way"

Llandy3d:
Please also consider ignoring the `googleapis` one when pressing `select all`

Llandy3d:
Also consider ⚠  icons with the `googleapis` entries

PR #616, opened 2025-04-01
feat: Pre-select first host in rercording

cristianoventura:
## Description

This PR pre-selects the first URL in `AllowlistDialog`. This is not limited to the user inputting a host when starting a recording.

In addition to this, common domain names requested by Chrome browser are reordered to the end of the list.

image

## How to Test

- Open a generator that doesn't have any hosts selected
- Check that the first item in the list is selected (which should be the first host recorded in the HAR file)
- Check that hosts that belong to `googleapis.com` and `google.com` are moved to the end of the list
- Check that the user can unselected the pre-selected domain
- The pre-selected domain feature should only work when opening another generator or changing the recording within the generator window

## Checklist

https://github.com/user-attachments/assets/43904fd4-1d90-493c-8a5a-92b9e3447de9

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/540

### DIFF
--- a/src/store/generator/slices/recording.utils.ts
+++ b/src/store/generator/slices/recording.utils.ts
@@ -20,7 +20,19 @@ export function groupHostsByParty(hosts: string[]) {
 }
 
 export function isHostThirdParty(host: string) {
-  const hostPatterns = ['.google.com', '.googleapis.com', '.gstatic.com']
+  const hostPatterns = [
+    '.google.com',
+    '.googleapis.com',
+    '.gstatic.com',
+    '.googleusercontent.com',
+    '.googleadservices.com',
+    '.doubleclick.net',
+    '.google-analytics.com',
+    '.googletagmanager.com',
+    '.googlesyndication.com',
+    '.googletagservices.com',
+    '.recaptcha.net',
+  ]
   return hostPatterns.some((pattern) => host.includes(pattern))
 }
 
--- a/src/store/generator/slices/recording.utils.ts
+++ b/src/store/generator/slices/recording.utils.ts
@@ -1,4 +1,4 @@
-import { uniq, orderBy } from 'lodash-es'
+import { uniq } from 'lodash-es'
 
 import { ProxyData } from '@/types'
 
@@ -6,13 +6,22 @@ export function extractUniqueHosts(requests: ProxyData[]) {
   return uniq(requests.map((request) => request.request.host).filter(Boolean))
 }
 
+export function groupHostsByParty(hosts: string[]) {
+  return hosts.reduce(
+    (acc, host) => {
+      const key = isHostThirdParty(host) ? 'thirdParty' : 'firstParty'
+      return {
+        ...acc,
+        [key]: [...acc[key], host],
+      }
+    },
+    { firstParty: [], thirdParty: [] }
+  )
+}
+
 export function isHostThirdParty(host: string) {
   const hostPatterns = ['.google.com', '.googleapis.com', '.gstatic.com']
   return hostPatterns.some((pattern) => host.includes(pattern))
-}
-
-export function orderThirdPartyHostsLast(hosts: string[]) {
-  return orderBy(hosts, isHostThirdParty)
 }
 
 export function shouldResetAllowList({
--- a/src/store/generator/slices/recording.utils.ts
+++ b/src/store/generator/slices/recording.utils.ts
@@ -1,4 +1,4 @@
-import { uniq, intersection, difference } from 'lodash-es'
+import { uniq, orderBy } from 'lodash-es'
 
 import { ProxyData } from '@/types'
 
@@ -11,12 +11,8 @@ export function isHostThirdParty(host: string) {
   return hostPatterns.some((pattern) => host.includes(pattern))
 }
 
-export function reorderHosts(hosts: string[]) {
-  const matchedHosts = hosts.filter((host) => isHostThirdParty(host))
-  const hostsToGoLast = intersection(hosts, matchedHosts)
-  const hostsToGoFirst = difference(hosts, hostsToGoLast)
-
-  return [...hostsToGoFirst, ...hostsToGoLast]
+export function orderThirdPartyHostsLast(hosts: string[]) {
+  return orderBy(hosts, isHostThirdParty)
 }
 
 export function shouldResetAllowList({
--- a/src/store/generator/slices/recording.utils.ts
+++ b/src/store/generator/slices/recording.utils.ts
@@ -1,9 +1,22 @@
-import { uniq } from 'lodash-es'
+import { uniq, intersection, difference } from 'lodash-es'
 
 import { ProxyData } from '@/types'
 
 export function extractUniqueHosts(requests: ProxyData[]) {
   return uniq(requests.map((request) => request.request.host).filter(Boolean))
+}
+
+export function isHostThirdParty(host: string) {
+  const hostPatterns = ['.google.com', '.googleapis.com', '.gstatic.com']
+  return hostPatterns.some((pattern) => host.includes(pattern))
+}
+
+export function reorderHosts(hosts: string[]) {
+  const matchedHosts = hosts.filter((host) => isHostThirdParty(host))
+  const hostsToGoLast = intersection(hosts, matchedHosts)
+  const hostsToGoFirst = difference(hosts, hostsToGoLast)
+
+  return [...hostsToGoFirst, ...hostsToGoLast]
 }
 
 export function shouldResetAllowList({
--- a/src/views/Generator/Allowlist/Allowlist.tsx
+++ b/src/views/Generator/Allowlist/Allowlist.tsx
@@ -4,7 +4,10 @@ import { useState } from 'react'
 
 import { PopoverDialog } from '@/components/PopoverDialogs'
 import { useGeneratorStore } from '@/store/generator'
-import { extractUniqueHosts } from '@/store/generator/slices/recording.utils'
+import {
+  extractUniqueHosts,
+  reorderHosts,
+} from '@/store/generator/slices/recording.utils'
 
 import { AllowlistDialog } from './AllowlistDialog'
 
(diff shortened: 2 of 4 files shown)


src/store/generator/slices/recording.utils.ts
```diff
@@ -1,11 +1,24 @@
...
+  const hostPatterns = ['.google.com', '.googleapis.com', '.gstatic.com']
+  return hostPatterns.some((pattern) => host.includes(pattern))
+}
+
+export function reorderHosts(hosts: string[]) {
+  const matchedHosts = hosts.filter((host) => isHostThirdParty(host))
+  const hostsToGoLast = intersection(hosts, matchedHosts)
+  const hostsToGoFirst = difference(hosts, hostsToGoLast)
+
+  return [...hostsToGoFirst, ...hostsToGoLast]
```
cristianoventura, line 19: `.googleapis.com` might be the first item in the HAR file due to Chrome automatically making requests but here we ensure that they are moved to the end of the array, leaving the first one to be a URL that the user actually requested.

src/views/Generator/Allowlist/AllowlistDialog.tsx
```diff
@@ -125,9 +131,14 @@ export function AllowlistDialog({
...
                 {filteredHosts.map((host) => (
                   <Text as="label" size="2" key={host}>
-                    <Flex gap="2">
+                    <Flex gap="2" align="center">
                       <CheckboxGroup.Item value={host} />{' '}
                       <Text truncate>{host}</Text>
+                      {isHostThirdParty(host) && (
+                        <Tooltip content="This host belongs to a third-party service">
+                          <ExclamationTriangleIcon />
+                        </Tooltip>
```
cristianoventura, line 140: I'm debating whether or not to say "third-party" service, as we won't be able to detect all possible third-party services being used on a recording.

e-fisher: That's a good point, here's an example recording of app.k6.io: we are implicitly saying hosts without the exclamation icon are not third-party, while in fact most of them are. So we either remove the icon altogether or update our third-party list to include most commonly used third party services. I'm leaning towards the latter, since it may provide additional value to users, even though we'd have to invest into researching most used services and maintaining the list. But I would love to hear others' opinion on this.

screencapture 2025-04-02 at 14 27 20

src/views/Generator/Allowlist/Allowlist.tsx
```diff
@@ -30,7 +33,19 @@ export function Allowlist() {
...
-  const hosts = extractUniqueHosts(requests)
+  const hosts = useMemo(() => {
+    const uniqueHosts = extractUniqueHosts(requests)
+    return reorderHosts(uniqueHosts)
+  }, [requests])
+
+  useEffect(() => {
+    // We only need the allowlist count when "hosts" changes
+    // This happens when the recording or the generator is changed
+    const allowlistCount = useGeneratorStore.getState().allowlist.length
```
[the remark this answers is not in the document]
cristianoventura: Using `allowlist.length` would require adding it as a dependency of `useEffect`. 

This causes an unintended behavior of automatically selecting the first item when the user unselects all checkboxes. (it's never possible to make allowlist empty).

Are there any other ways you can think of solving this?

src/views/Generator/Allowlist/AllowlistDialog.tsx
cristianoventura: Good suggestion. I'll raise this to others to get their opinions

cristianoventura (commented):
> I've noticed there was a suggestion in the issue about making "select all" not include third-party hosts, have you considered including it in this PR?

I thought the behaviour would be counter intuitive when you're clicking "select all" but not _all_ items get selected. Another option could be hiding these requests similarly to how we do with static assets 🤔

src/store/generator/slices/recording.utils.ts
```diff
@@ -1,11 +1,24 @@
...
 
+export function isHostThirdParty(host: string) {
+  const hostPatterns = ['.google.com', '.googleapis.com', '.gstatic.com']
+  return hostPatterns.some((pattern) => host.includes(pattern))
+}
+
+export function reorderHosts(hosts: string[]) {
+  const matchedHosts = hosts.filter((host) => isHostThirdParty(host))
+  const hostsToGoLast = intersection(hosts, matchedHosts)
+  const hostsToGoFirst = difference(hosts, hostsToGoLast)
```
allansson: Maybe we should look at polyfilling the [new set methods](https://github.com/zloirock/core-js?tab=readme-ov-file#new-set-methods)?

src/views/Generator/Allowlist/Allowlist.tsx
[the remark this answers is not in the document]
allansson: If we keep the code, replace your code comment with what you just wrote. 🙂 

The issue here is that we're acting on data being changed without knowing why it changed. And the why here is what's important. Like your code comment states: `This happens when the recording or the generator is changed`. That is, there's a very specific circumstance where this should happen.

So the code should move to where either of those things happened, which in our case would be the `setGeneratorFile` and `setRecording` actions on the zustand store.

_NOTE: You won't have access to `hosts` in the store, but that's because we're re-calculating the list of hosts in every place we need them. The hosts only change when the recording or generator file change, so we can calculate the hosts in those actions and keep them in the store._

allansson (commented):
Just some food for thought.

src/views/Generator/Allowlist/AllowlistDialog.tsx
Llandy3d: I think it makes sense, and we can always increase the list overtime.
I wouldn't dedicate too much time on this research because the main objective we have here is to have the user avoid googleapis requests that are going to make running the script fail due to the odd binary json thingy.

Having all third-parties detected would be great indeed but it can be a follow-up item to this 🤔

Llandy3d (commented):
should we better organize the grouping ? Maybe with a divider 🤔 
From what I see we have 3 distinct groups, the first one is with the host the user used, the second one are other requests and the final one is "third-parties"

Another question would be if we should group on the pre-selected list subdomains that make use of the same host 🤔 (here quickpizza is selected for clarity otherwise it would be easy to miss in the list)
image

Regarding the `select all` the behaviour has not changed in this PR meaning that it's less useful because we are still allowing the user to press it and add hosts that shouldn't be there. I would start with the removal of the "third-parties" detected from that functionality and we can add into it as we see more problematic domains 🤔 
With group division this could make more sense since we can inform that "third-parties" hosts are ignored by the select-all

allansson:
> Another question would be if we should group on the pre-selected list subdomains that make use of the same host 🤔 (here quickpizza is selected for clarity otherwise it would be easy to miss in the list)

Building on @Llandy3d's idea of grouping It might be helpful thinking about the domains as a tree instead:

```
[-] Hosts
  [✓] grafana.com
    [✓] grafana.com
    [✓] quickpizza.grafana.com
  [-] k6.io
    [✓] app.k6.io
    [ ] test.k6.io
[ ] 3rd parties ⓘ
  [ ] google.com
    [ ] accounts.google.com 
```

The 3rd parties option would be turned off and collapsed by default. The Select and Remove All buttons should be removed, because we want to prevent the user from accidentally including the 3rd parties.

* It makes it easy to whitelist an entire org.
* It tucks the 3rd parties away somewhere separate, so there's less clutter.
* Selecting all 3rd parties is a separate action, so it's harder to accidentally include them.

The very few (if any) users who _do_ need to test the third parties will just have to look for their domains under the 3rd parties tree instead.

src/views/Generator/Allowlist/AllowlistDialog.tsx
allansson: Any domain that the user didn't explicitly visit is a suspected 3rd party.

I'd say we are looking at 3 types of hosts:

- Those that we are pretty certain are 1st party hosts (explicitly visited second level domains)
- Those that we think might not be (those that don't share the same second level domains as those above)
- Those that are almost definitely 3rd party (well-known tracking services)

Right now we can only go on what the user typed as their starting URL (if anything), but with browser recording we can actually see which pages the user navigated to either via page interactions or by manually entering a URL. This means that we can have a better idea of what that first and second group should be.

How about this:

1. If it's a well-known tracking service, we put it in the 3rd party bin.
2. If it's not, we put it in the 1st party bin.
3. If it belongs to a second-level domain that the user visited in the browser, we pre-select it.

cristianoventura:
Thanks for all the reviews!

I've made the following changes to address some of your feedback:

- Removed the ⚠️ icon from 3rd party hosts and instead, added a separator between them in the list
- "Select all" button only affects first-party hosts
- Expanded the list of third-party hosts with some common domain names
- I made a ticket https://github.com/grafana/k6-studio/issues/661 to address the grouping of hosts that can be improved later

https://github.com/user-attachments/assets/37069b52-e3b6-4b6b-8baf-f706a986eb0f

src/views/Generator/Allowlist/AllowlistDialog.tsx
```diff
@@ -66,6 +76,10 @@ export function AllowlistDialog({
     setIncludeStaticAssets(checked && staticAssetCount > 0)
   }
 
+  function shouldDisableSelectAll() {
+    return every(firstPartyFilteredHosts, (host) => includes(allowlist, host))
+  }
```
e-fisher, line 81: NIT: This would be calculated on every render, maybe we could call this `isSelectAllDisabled` and wrap in `useMemo` instead?

e-fisher (approved):
LGTM 🚀

Llandy3d (commented):
just some questions:

* Should we also have a group for the first section ?
* Should we move the faro endpoint to the list of third parties ?
* Should we add a warning/info icon to explain what are the third parties hosts ? (In the current state seems to be confusing why you press select all and those don't get selected, also you can manually select them so we are missing out on warning the user not to 🤔 )

Llandy3d (approved):
LGTM! 🙌

PR #644, opened 2025-04-10
chore: release test version action

Llandy3d:
## Description

Adds a github action to build a test version of k6 studio without having to jump through hoops.
The action will require the user to specify a `version` to be used for that specific test release.

## Checklist

### DIFF
--- a/.github/workflows/release-test-version.yml
+++ b/.github/workflows/release-test-version.yml
@@ -1,0 +1,132 @@
+name: 'Manual Release Test Version'
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Specify a version for the test release (ex. test-browser)"
+        required: true
+        type: string
+        default: "test-version"
+
+jobs:
+  publish-studio:
+    permissions:
+      contents: write
+    strategy:
+      fail-fast: false
+      matrix:
+        # There is a bug on Ubuntu 22.04 (ubuntu-latest) regarding stripping a binary for a different
+        # architecture. Since we include both x86_64 and arm64 in linux a solution is to use ubuntu 20.04
+        # https://github.com/electron/forge/issues/3102
+        # https://github.com/electron/forge/issues/3701
+        platform: [macos-latest, windows-latest, ubuntu-20.04]
+
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set version for test release
+        with:
+          version: ${{ inputs.version }}
+        run: python update_version.py
+
+      - name: setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      # try earlier python version for appdmg
+      # https://github.com/electron/forge/issues/3371#issuecomment-2105195302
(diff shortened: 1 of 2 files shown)


.github/workflows/release-test-version.yml
```diff
@@ -0,0 +1,132 @@
...
+        platform: [macos-latest, windows-latest, ubuntu-20.04]
+
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set version for test release
+        with:
+          version: ${{ inputs.version }}
+        run: python update_version.py
```
going-confetti, line 32: Will it work if executed before `actions/setup-python@v4`?

.github/workflows/release-test-version.yml
```diff
@@ -0,0 +1,132 @@
+name: 'Manual Release Test Version'
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Specify a version for the test release (ex. test-browser)"
+        required: true
+        type: string
+        default: "test-version"
```
going-confetti, line 10: IMO it's better to ask for a suffix that will then be appended to the current version. This way it will be clearer which version we're branching off of. Not a strong opinion though.

.github/workflows/release-test-version.yml
```diff
@@ -0,0 +1,132 @@
+name: 'Manual Release Test Version'
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Specify a version for the test release (ex. test-browser)"
+        required: true
+        type: string
```
going-confetti, line 9: Do you know if the input string is trimmed by default? Just to make sure there won't be version ` `

.github/workflows/release-test-version.yml
Llandy3d: We can probably debate it in the future although it doesn't feel mandatory to have the current version there in favor of a more expressive name 🤔

.github/workflows/release-test-version.yml
Llandy3d: I don't think it's trimmed since the default value if not set like we do we would be `''`, feels like a small edge case not too important because the most common case of forgetting to set it would be that the default applies of `test-version`

.github/workflows/release-test-version.yml
going-confetti: One doesn't negate the other: `1.1.0-test` is more expressive and has more info that `test-version`

Llandy3d: I was thinking specifically as browser as an example, it would remove the possibility of having a full custom name like it was used recently if we always append to the end 🤔 
But it could be a good default (ex. if the version field is not specified, we just append `test` to the version)

going-confetti: > it would remove the possibility of having a full custom name like it was used recently if we always append to the end

Do we need this?
In my opinion. having a version for the browser builds makes a lot of sense: the browser branch has been frequently synced with main, and having a version there makes it easier to distinguish different builds

going-confetti (approved):

PR #645, opened 2025-04-10
chore: manual release action & update runner to ubuntu-latest

Llandy3d:
## Description

This finalize the new action for manually building a test version of k6 Studio from any branch with a specific version name suffix. (ex. `test-version` builds a release called `v1.2.0-test-version` where the first version is taken from the current set version)
This also updates the runner to use `ubuntu-latest` since `ubuntu-20.04` will be deprecated soon.

## Checklist

### DIFF
--- a/update_version.py
+++ b/update_version.py
@@ -10,7 +10,8 @@ with open('package.json', 'r') as f:
 
 
 with open('package.json', 'w') as f:
-    data['version'] += f'-{version}'
+    updated_version = f'{data["version"]}-{version}'
+    data['version'] = updated_version
     f.write(json.dumps(data, indent=2))
 
-print(f'Updated package.json version to {version}')
+print(f'Updated package.json version to {updated_version}')


.github/workflows/release-test-version.yml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Specify a version for the test release (ex. test-browser)"
+        description: "Specify a version suffix for the test release (limit to 20 characters or windows build will fail)"
```
going-confetti, line 7: 🤯

update_version.py
```diff
@@ -2,15 +2,15 @@
...
 with open('package.json', 'r') as f:
     data = json.loads(f.read())
 
 
 with open('package.json', 'w') as f:
-    data['version'] = version
+    data['version'] += f'-{version}'
     f.write(json.dumps(data, indent=2))
 
 print(f'Updated package.json version to {version}')
```
going-confetti, line 16: This will print just the suffix, right?

.github/workflows/release-test-version.yml
Llandy3d: welcome to the world of `.nuspec` 😢

update_version.py
Llandy3d: nice catch! pushed fix for the info log

Issue #368, opened 2024-12-02
highlight where values are replaced in Rule Preview


Llandy3d:
When inspecting requests in the Rule preview pane is not clear at a first glance where values are replaced whether by correlation or parameterization.

We should strive to highlight where this replacement happens.

PR #665, opened 2025-04-16
feat: Highlight values replaced by rules

e-fisher:
## Description

screencapture 2025-04-16 at 15 42 02

- Use same highlighting logic as search, but use different color (green) to highlight values modified by rules
- Diff between values is calculated using `diff` library, then gets converted to `fusejs` format. This will also allow implementing highlight of values changed between recording and validation run in the future.
- For payload changes preview monaco diff editor is used
- Preview x-www-form-urlencoded in monaco (instead of table) to support showing diff
- Diff editor is displayed inline on smaller screens, side-by-side on larger screens

## How to Test

Test with different payload types and all rule targets, verify values are highlighted as expected.
Verify nothing is broken in recording viewer & validator.

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/368


src/store/generator/hooks/useHighlightRequestChanges.ts
```diff
@@ -0,0 +1,119 @@
...
+      return addHighlights(originalRequest, data)
+    })
+  }, [requests, originalRequests])
+}
+
+function addHighlights(
+  originalRequest: ProxyDataWithMatches['request'],
+  data: ProxyDataWithMatches
+) {
+  // Don't overwrite search matches when present
```
going-confetti, line 34: I find this behavior confusing, because it doesn't match how other tools work.
I don't think this should be a blocking comment as long as we can create a ticket (and add it to stretch goals so it doesn't get lost) to implement more flexible text decorations.

src/components/Monaco/DiffEditor.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface Props extends ComponentProps<typeof Editor> {
+  original?: string
+}
+
+export function DiffEditor({ original, value, language, options }: Props) {
+  const theme = useTheme()
+  return (
+    <MonacoDiffEditor
```
going-confetti, line 22: Shouldn't it also have the line wrap button? Currently, it does not, which leads to layout shift when switching between content and diff
image

src/components/Monaco/DiffEditor.tsx
```diff
@@ -0,0 +1,39 @@
...
+      options={{
```
going-confetti, line 23: Is there any way to disable this popover?
image

src/utils/fuse.ts
```diff
@@ -1,12 +1,13 @@
...
-import { SearchMatch } from '@/types/fuse'
+import { Match } from '@/types/fuse'
 
 type ResultWithMatch<T> = T & {
-  matches: SearchMatch[]
+  matches: Match[]
 }
 
 export function withMatches<T>(result: FuseResult<T>): ResultWithMatch<T> {
+  // @ts-expect-error assigning readonly indices to mutable
```
going-confetti, line 10: Using `ReadonlyArray` instead of `Array` in /src/types/fuse.ts` solves this problem

going-confetti (dismissed):
Very nice addition!
I added some comments (few of them optional) with the most important one being about showing the original value in `HighlightedText`

src/components/WebLogView/ContentPreview.tsx
```diff
@@ -22,13 +25,20 @@ export function ContentPreview({
...
   searchString,
+  originalContent,
 }: ContentPreviewProps) {
   const [selectedTab, setSelectedTab] = useState('content')
 
   useEffect(() => {
     setSelectedTab('content')
   }, [format])
 
+  useEffectOnce(() => {
```
going-confetti, line 36: Some thoughts about this one:
1. What do you think about displaying the diff in the content tab? I haven't thought this through and it might not be a good idea since "raw" and "preview" will not have this info and it might be confusing.
2. If the request data is unaffected, shouldn't we hide the diff option? 
3. Regarding `useEffectOnce` specifically, is it here just to set a different default value? If so, it's worth moving this logic to line 30 IMO:
```ts
const [selectedTab, setSelectedTab] = useState(originalContent && originalContent !== content && !searchString ? 'diff' : 'content')
```

src/components/HighlightedText.tsx
```diff
@@ -57,19 +68,25 @@ export function HighlightedText({
...
-      highlightAllMatches ? indices : longestMatchOnly(indices)
+      highlightAllMatches ? filteredMatches : longestMatchOnly(filteredMatches)
     )
   }, [text, matches, highlightAllMatches])
 
+  if (!segments.length) {
+    return <span>{text}</span>
+  }
+
   return (
```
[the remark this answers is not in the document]
Llandy3d: I'm not convinced for URLs, the current way is compact and easy to digest, if you have a full diff that's going to make that just harder to read/noisy and also probably not completely correct if two rules are affecting the value of one another 🤔 

`View original requests` stays important to see the recording in it's pure form, removing all kind of noises, I wouldn't remove that

Llandy3d (approved):
Reviewing from the functionality standpoint, this works great 🙌

src/components/HighlightedText.tsx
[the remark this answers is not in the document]
Llandy3d: > What do you think about showing the full diff in headers (since your comment only covers the URLs)?
Shouldn't "noisy" and "probably not completely correct" also be true for full diffs in monaco (especially the not completely correct part)?

No strong opinions on headers but overall I wonder if it would be better to show the previous value maybe on `hover` rather than full diff, is that diff information really that useful ? we would need to figure it out 🤔 

> Can you elaborate on why we'd need it if there was a clear diff for every replacement?

Without that toggle you have no way to see the original recording, even with diffs you would be seeing the last modified values, so if the value was modified multiple times you wouldn't be able to see the original value, hence missing the possibility of checking the original raw recording making you leave the generator and going into recordings to see that information

src/store/generator/hooks/useHighlightRequestChanges.ts
e-fisher: This one is a bit tricky, since we are using same methods of highlighting for both search and diff, we can't show both highlights when the same string is matched. I've checked how diff editor handles this, and it works similar to this - search match overwrites diff highlight:

No search, `admin` highlighted in green
screencapture 2025-05-05 at 14 23 26

Search match - `admin` becomes highlighted in gray
screencapture 2025-05-05 at 14 23 34

The only difference is that monaco still highlights the whole line as an "addition", is that what you have in mind?

going-confetti: No, I was thinking about supporting multiple kinds of text nodes/decorations, that can overlap each other. For example, Google Docs doesn't overwrite search matches when they overlap with the comment highlighting:

image

src/components/HighlightedText.tsx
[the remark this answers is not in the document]
going-confetti: >No strong opinions on headers but overall I wonder if it would be better to show the previous value maybe on hover rather than full diff, is that diff information really that useful ?

I like this idea. Perhaps something toggleable with a mouse click.

>Without that toggle you have no way to see the original recording, even with diffs you would be seeing the last modified values

I think I need to brush up on how rule application works. I think I get it

src/components/Monaco/DiffEditor.tsx
e-fisher: There's a [bug](https://github.com/microsoft/monaco-editor/issues/3873) in monaco which doesn't allow disabling it with options, added CSS rule to hide it

src/components/WebLogView/ContentPreview.tsx
e-fisher: I think it's till worth keeping content and diff tabs separate, just to have an option to view content without diff highlights. Other comments addressed 👍

e-fisher:
All great points, thank you for the feedback! I've addressed most comments in this PR and created issues to be addressed in the next iteration for adding removed value tooltip, supporting both highlights at the same, and adding "modified" icon to inspector tabs. 

I've marked all 3 issues as stretch goals with priority of 9 so they don't get lost and grouped them into [epic](https://github.com/grafana/k6-studio/issues/718).

This is ready for another review, please let me know if I missed something 🙌

cristianoventura (approved):
LGTM! 🚢

Issue #654, opened 2025-04-11
`json` replacer not working for falsy values in Parameterization


Llandy3d:
### Describe the bug

Trying to replace a value with the `json` selector won't replace the value if it's falsy. (ex. `boolean` or `""` empty string)

This is most likely due to the check here: https://github.com/grafana/k6-studio/blob/main/src/rules/shared.ts#L258

This might not be limited to the `json` selector since also the other selectors make use of the same logic, to be confirmed.

### Steps to reproduce the problem

* Create a recording from `quickpizza.grafana.com` advanced pizza form
* Add a Parameterization rule with `JSON` selector and try to match on values
* Notice by inspecting the request that it won't be modified when the value is falsy

### Expected behavior

Parameterization rule with `json` selector should be always replace the matched key

### Actual behavior

Matched key not replaced when falsy

### Grafana k6 Studio version

main

### OS

all

PR #666, opened 2025-04-16
fix: Replace falsy values with json selector

e-fisher:
## Description

This fixes the issue when JSON body selectors don't work on falsy properties (empty string, boolean false).

## How to Test

Create a parameterization or correlation rule targeting a falsy property using JSON body selector.

Expected:
Rule matches and value replaced

Current:
Rule doesn't match

> [!WARNING]
> Don't rely on `quickpizza` for testing this because it sets incorrect content type (text/plain) and JSON selectors are not applied.

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/654


cristianoventura (commented):
Works well for parameterization, but I found issues when trying to extract a falsy value from a correlation.

- I tested the correlation rule using this endpoint `https://pokeapi.co/api/v2/pokemon/ditto`
- Attempt to extract `abilities[0].is_hidden` with no success

image

I did some debugging and found two missing places for it to work for correlation rules. We need to explicitly compare against `undefined`, similar to what you did in the PR.

https://github.com/grafana/k6-studio/blob/f1f035a0993deb9a9eff851d38e0f85432b065f4/src/rules/correlation.ts#L128

https://github.com/grafana/k6-studio/blob/f1f035a0993deb9a9eff851d38e0f85432b065f4/src/views/Generator/RuleEditor/CorrelationEditor.tsx#L141

e-fisher:
@cristianoventura good catch! This is an interesting case, because usually you wouldn't want to replace all "false" value, but perhaps, with the combination of custom replacer, this could be used to have dynamic boolean in payload depending on a previous response.

cristianoventura (approved):
LGTM! 🚀

PR #668, opened 2025-04-16
internal: main window cannot be restored on macOS

cristianoventura:
## Description

We introduce a change that tries to re-register IPC handlers when the main window is re-created. This PR reverts that change.

## How to Test

- Launch the app on macOS
- Close the main window
- Restore it from the dock
- Check that the window was properly recreated

## Checklist

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -102,7 +102,9 @@ if (require('electron-squirrel-startup')) {
 }
 
 initializeLogger()
-
+handlers.initialize({
+  browserServer,
+})
 mainState.initialize()
 
 // Used to convert `.json` files into the appropriate file extension for the Generator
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -197,11 +199,6 @@ const createWindow = async () => {
     },
   })
 
-  handlers.initialize({
-    browserWindow: mainWindow,
-    browserServer,
-  })
```
cristianoventura, line 203: @allansson I moved this outside of `createWindow` as calling it inside re-attempts to register the IPC handlers when the window is restored from the dock.

```
11:47:03.599 › Unhandled rejection Error: Attempted to register a second handler for 'auth:get-profiles'
    at IpcMainImpl.handle (node:electron/js2c/browser_init:2:100357)
```

I also removed `browserWindow` as a dependency since we're not using it right now. Let me know if this disrupts any work you're doing.

allansson: The `browserWindow` is a relic from a previous iteration before there was a `browserServer` so removing it is the right call.

allansson (approved):
Great find! LGTM! 👍

PR #669, opened 2025-04-16
internal(browser): Unified schema for assertion events

allansson:
## Description

While working on adding additional types of assertions I came to the conclusion that it was better to have a single browser event for all assertions and let the event contain the specifics of the assertion.

I want to get this merged before releasing so that I won't have to add a migration for it.

## How to Test

1. Start a recording
2. Add a text assertion
3. Stop recording
4. Open browser events list
5. Export script
6. Verify that generated script calls `toContainText`

## Checklist

### DIFF
--- a/src/schemas/recording/v1/browser.ts
+++ b/src/schemas/recording/v1/browser.ts
@@ -9,21 +9,21 @@ const BrowserEventBaseSchema = z.object({
   timestamp: z.number(),
 })
 
-const NavigatedToPageEventSchema = BrowserEventBaseSchema.extend({
-  type: z.literal('navigated-to-page'),
+const NavigateToPageEventSchema = BrowserEventBaseSchema.extend({
+  type: z.literal('navigate-to-page'),
   tab: z.string(),
   url: z.string(),
   source: z.union([z.literal('address-bar'), z.literal('history')]),
 })
 
-const ReloadedPageEventSchema = BrowserEventBaseSchema.extend({
-  type: z.literal('reloaded-page'),
+const ReloadPageEventSchema = BrowserEventBaseSchema.extend({
+  type: z.literal('reload-page'),
   tab: z.string(),
   url: z.string(),
 })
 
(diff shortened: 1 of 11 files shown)


src/schemas/recording/v1/browser.ts
```diff
@@ -73,16 +73,24 @@ const FormSubmittedEventSchema = BrowserEventBaseSchema.extend({
...
     type: z.literal('contains'),
     value: z.string(),
   }),
 })
 
+// Will eventually be a union of different assertion types
+const AssertionSchema = TextAssertionSchema
+
+const AssertEventSchema = BrowserEventBaseSchema.extend({
+  type: z.literal('assert'),
```
allansson, line 88: I know it breaks the past tense rule of event names, but it just sounds better as an imperative.

allansson (commented):
Self-review

cristianoventura (commented):
The schema structure looks good! However, I noticed that text assertions are not being detected after this change.

https://github.com/user-attachments/assets/1326ee78-0aaa-4a1d-b57d-0a06804aaf0a

allansson:
> The schema structure looks good! However, I noticed that text assertions are not being detected after this change.
> 
> The assertion shows up on the Events list within the browser, but not on the main window. When exporting the script, the assertions are not included.

@cristianoventura It works for me. Maybe building the extension failed? That would mean that you are running an older build sending events with the old schema, which would then be rejected by the server. Did you run `npm i`? 🤔 

We saw some issues when navigating via browser history the other day, but that was fixed. Did you use the forward or back buttons? But since you see the events in event list in the browser it doesn't seem to be quite the same issue.

If it's not that, then something must've happened with the connection between the browser and k6 Studio. I should add better logging on the k6 Studio side (and I suppose it would be nice if errors in the extension were sent to the log as well).

Either way, it _feels_ unlikely that the changes here would've cause the issue. I'm going to wait and see if @going-confetti sees the same issue.

going-confetti (approved):

PR #671, opened 2025-04-17
fix: Preview payload when correlating numeric properties

e-fisher:
## Description

Payload preview doesn't work when a numeric value is replaced by a correlation variable, for example:
```json
{ "id": ${correlation_vars['correlation_0'] }
```
In Payload preview, we parse the string as JSON, but the injected variable makes in invalid. 

To solve this I've added a function to wrap variable expressions in quotes, that's not ideal because it makes it look like the value injected would be a string, but at least it doesn't break the preview. We might want to revisit this in the future and remove JSON parse altogether, but we'd need a new way to prettify it. 

## How to Test

Use attached recording and generator.

Steps to reproduce:
1. In the generator, inspect `POST /api/ratings` request 
2. Click on payload.

**Expected result**: payload preview with injected correlation value is displayed
**Actual result**: payload preview is not available

**Verify payload preview works as expected in recording, generator and validator**

[correlate-pizza-id.zip](https://github.com/user-attachments/files/19792449/correlate-pizza-id.zip)

## Checklist

screencapture 2025-04-17 at 12 59 12

### DIFF
--- a/src/components/WebLogView/RequestDetails/utils.ts
+++ b/src/components/WebLogView/RequestDetails/utils.ts
@@ -35,7 +35,11 @@ export function parseParams(data: ProxyData) {
     return stringify(
       // TODO: https://github.com/grafana/k6-studio/issues/277
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-      JSON.parse(jsonrepair(parsePythonByteString(contentDecoded)))
+      JSON.parse(
+        jsonrepair(
+          parsePythonByteString(wrapTemplateExpressionsInQuotes(contentDecoded))
+        )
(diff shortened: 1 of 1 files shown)


Llandy3d (approved):
Works well 🙌

going-confetti (approved):
Works as expected and remedies the issue 👍 

> that's not ideal because it makes it look like the value injected would be a string, but at least it doesn't break the preview. We might want to revisit this in the future and remove JSON parse altogether, but we'd need a new way to prettify it.

I agree, it looks pretty confusing. I think we should remove JSON parse in future. Could you please create an issue for that?

Issue #605, opened 2025-03-26
Implement autocompletion for extracting JSON properties


cristianoventura:
When users are extracting JSON properties, we’d like to introduce autocompletion that suggests available properties as users type. This feature would help users quickly identify and select the correct properties, reducing the chances of errors and speeding up the process of extracting data from JSON structures.

MoAqsam:
I'd like to tackle this, i'd like a couple things clarified, as I skimmed the source code, I tend to do some up front back of the envelopes, open for feedback.

## Autocomplete/Typeahead component

Seems beside the Select being searchable, there is no, native out of the box feature for the searchable dropdown input. Can be built easily without adding new dependencies, not sure the leaning of the team here.

1. Can leverage [popovers](https://www.radix-ui.com/primitives/docs/components/popover) 
2. Existing select also kinda works, but semantically may not be the correct choice, though could potentially compose off of it.
3. Create a custom component that does input searchable drop-downs based on a set of options as a wrapper over the current inputs.

> [!note]
> Leaning towards a custom wrapper over `Input.Root`, its very simple, and the underlying component or usage can easily be replaced.

### Question 1
#### Effort: Small
- Very small effort but want to align on the heuristics of the team regarding searchable functionality leant towards customize or dependency ?

### Question 2
- How can I test the usage of the JSON inputs, i was looking at the parameterization options in the generators, but was unsure how to test it comprehensively, would like a quick reference on it.

reference used:
- https://grafana.com/docs/k6-studio/record-your-first-script/#parameterize-a-value

## Acceptance Criteria

### Overview
I'll break this feature into the following three areas to ensure parity with the existing inputs:

1. State Management
2. Autocomplete Component for JSON Paths
3. Editor Component Updates

### Structures to tackle

#### Typeahead component

- [ ] create a new component called: **Typeahead** (Picked this as its semantically relevant but open for change)
- [ ] component will render json path renderings as drop downs:
       - [ ]  Renderings of the json path, will show as the user types.
       - [ ]  User can also scroll through a lazy loaded view of the items (virtualized)
- [ ] Arrow keys to navigate dropdown options: (accessibility)
       - [ ] Enter to select an item
       - [ ] Esc to close dropdown
       - [ ] Tab and Shift+Tab 

#### State Management

- [ ] leverage the generator [store](https://github.com/grafana/k6-studio/blob/main/src/store/generator/useGeneratorStore.ts) to optimize rendering of json key paths, based on request content-types and/or json payloads.
[...]

PR #678, opened 2025-04-21
internal: [POC] Render JSON suggestions

MoAqsam:
Supports
- https://github.com/grafana/k6-studio/issues/605

This is a quick POC covering at-least a subset of the spec, as Im travelling, I wanted to make sure I could get feedback as early on as possible. I did run into areas where memory optimization and state optimization would've required some large refactoring. So for now, goal is to tackle the POC.

## Description

This PR implements a subset of the spec for rendering json paths. Its broken down into 5 commits. Please refer to the comments to gain context, here:
- https://github.com/grafana/k6-studio/issues/605

## 1. Typeahead
1b007f9e63e565752bf44846f441e51e7f7e8169

In this portion the useTypeahead.tsx and Typeahead.tsx are implemented. These two were decoupled as it allows  us to implement and test our behavioral logic in isolation, whilst allowing the Typeahead component to be interchangeable and swappable.

#### Suggestion Modes
It has three modes that can be used:

1. onDot - This will only render a suggestion when the first dot in the input is invoked _**(Default)**_
2. onFirstKey - This will start rendering on the first key
3. onThirdKey - This will start rendering on the third key, an alternative to the first as it reduces the bulk load of searching

#### Accessibility
1. Reacts to keyboard move keys (up, down)
2. Enter can be clicked on a selection

> [!note]
> The Typeahead.tsx is just a sample component, I had some hard time with Popovers, so this seem to be the simplest and most sensible solution. Open for suggestions and feedback, there is definitely some tweaks to do here.

## 2. Refactor
087c8f44c665737061e9cb73d564eb0a1f8609ae

This was a small refactor for making the editor mappings more dynamic. You can just add a new key and map the new editor component.

## 3. State Management
2c5ae31d68587ad1be0a169403cfc47e6a32a486

This portion dealt with the state management. For the sake of the POC, it computes the values when the store is invoked, and separates it into `request` and `response` json paths. 

> [!note]
> In order to prevent scope creep, and large refactors computation at invocation of the state from the store was done. Though long term, there should be some high optimization around rendering and retrieving the json paths lazily. (lazy loaded) This may require implementing business rules, depending on the size of the data for the recordings. (large data gets lazy loaded, small data gets invoked)

## 4. Feature Flag
4588ec3350d19b395341c46196cbb611cd2b3315
The POC was wrapped around a feature flag.

## 5. Integration
2a1bbdba9087e28be63b0841dcfad1313752cb07

The typeahead component was integrated and wrapped around the feature flag.

> [!note]
> There is variants to the poc with `suggestion mode` which allows you to test different behaviours, in order to prevent refactoring the feature state, this was localized to the component.

## How to Test

1. Go through https://grafana.com/docs/k6-studio/record-your-first-script/#parameterize-a-value
2. Select JSON as the parameterization key
3. Start typing

#### Video
_**I dont know whats wrong i can't upload anything**_

### Correlation
1. Go through https://grafana.com/docs/k6-studio/record-your-first-script/#correlate-dynamic-data
2. Select JSON
3. Start typing until you hit the first dot

#### Video
_**I dont know whats wrong i can't upload anything**_

## Checklist

## Related PR(s)/Issue(s)
- https://github.com/grafana/k6-studio/issues/605


src/views/Generator/RuleEditor/SelectorField.tsx
```diff
@@ -108,28 +119,49 @@ export function SelectorField({
...
   selector: Selector
   field: 'extractor.selector' | 'replacer.selector' | 'selector'
 }) {
   const {
     register,
     formState: { errors },
   } = useFormContext<TestRule>()
-
+  /**
+   * @poc - might be questionable to put the state here, but to prevent too much refactoring keeping it here for now.
```
e-fisher, line 141: You could set mode directly in JSX without the need to keep state in this hook:

```ts
  const typeAheadEnabled = useFeaturesStore(
    (state) => state.features['typeahead-json']
  )
```
...
```tsx
{typeAheadEnabled ? (
  <Typeahead
    placeholder="Search for JSON key paths"
    defaultValue={selector.path}
    mode="onDot"
    options={options || []}
    {...register(`${field}.path`)}
  />
) : (
  <TextField.Root {...register(`${field}.path`)} />
)}

```

e-fisher (changes requested):
@MoAqsam thank you for the contribution! 🙌 We've done an initial review, there's room for refactoring of the code, but the main problem is calculating suggestions on every render. 

To improve this, the suggestion processing should be done when recording is loaded, inside the main process to avoid blocking UI threads. We have an example of processing in main, parsing of generator file:

https://github.com/grafana/k6-studio/blob/560ee05dd8aedd45b79a80fb7af2530493523c25/src/handlers/generator/index.ts#L51

You should be able to do this similarly:
- Create a separate handler to load har file **and** calculate json paths. Generators would use this handler, because suggestions are only needed in generator context.
- Store json paths in generator state, use in UI

**❗ Be sure to pull main before you continue, because we've done some refactoring recently**

When this is addressed, we'll be able to do another round of review and continue from there. 

The UI/UX change requests could be addressed in a separate PR, while this is under a feature flag. 

Feel free to comment if you have questions or need help, this looks promising!

MoAqsam:
Thank you @e-fisher! Appreciate the first pass.

Im currently traveling outside for now, and will address all the points when i'm back!

src/views/Generator/RuleEditor/RuleEditor.tsx
```diff
@@ -24,22 +24,25 @@ import { CustomCodeEditor } from './CustomCodeEditor'
...
-      return <CustomCodeEditor />
-    case 'parameterization':
-      return <ParameterizationEditor />
-    case 'verification':
-      return <VerificationEditor />
-    default:
-      return exhaustive(ruleType)
+  const editor = editorMappings[ruleType]
+  if (!editor) {
+    return exhaustive(ruleType as never)
```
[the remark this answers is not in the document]
MoAqsam: Reverted here

https://github.com/grafana/k6-studio/pull/678/commits/37def1d6b6edea6a2092da59acb0f9be5586941d

src/views/Generator/RuleEditor/SelectorField.tsx
MoAqsam: Addressed here

https://github.com/grafana/k6-studio/pull/678/commits/0096e62f6702c210582603a8586372740ce64b81

src/views/Generator/RuleEditor/Typeahead/Typeahead.tsx
```diff
@@ -0,0 +1,56 @@
+import { Box, Card, TextField } from '@radix-ui/themes'
+import * as React from 'react'
```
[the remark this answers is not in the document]
MoAqsam: Addressed here

https://github.com/grafana/k6-studio/pull/678/commits/0096e62f6702c210582603a8586372740ce64b81

MoAqsam:
> @MoAqsam thank you for the contribution! 🙌 We've done an initial review, there's room for refactoring of the code, but the main problem is calculating suggestions on every render.
> 
> To improve this, the suggestion processing should be done when recording is loaded, inside the main process to avoid blocking UI threads. We have an example of processing in main, parsing of generator file:
> 
> https://github.com/grafana/k6-studio/blob/560ee05dd8aedd45b79a80fb7af2530493523c25/src/handlers/generator/index.ts#L51
> 
> You should be able to do this similarly:
> 
> * Create a separate handler to load har file **and** calculate json paths. Generators would use this handler, because suggestions are only needed in generator context.
> * Store json paths in generator state, use in UI
> 
> **❗ Be sure to pull main before you continue, because we've done some refactoring recently**
> 
> When this is addressed, we'll be able to do another round of review and continue from there.
> 
> The UI/UX change requests could be addressed in a separate PR, while this is under a feature flag.
> 
> Feel free to comment if you have questions or need help, this looks promising!


## Update

Over [here](https://github.com/grafana/k6-studio/pull/678/commits/02f8cf5dc2b16c21bf090b91b79e10b59a9c2967), i moved towards precomputing the json paths on load, and storing the unique paths in a metadata field for the UI to consume.

src/utils/harToProxyData.ts
```diff
@@ -58,19 +70,34 @@ function parseRequest(request: Entry['request']): Request {
...
+    jsonPaths,
+    content,
     contentLength: response.content?.size ?? 0,
     timestampStart: 0,
     path: '',
   }
 }
 
+function parseJsonPaths(content: string, headers: Header[]): string[] {
+  const isJsonPathsFeatureFlagTrue = defaultFeatures['typeahead-json']
```
cristianoventura: The issue is likely related to this line returning false, because it's reading from the static declaration value instead of reading from the runtime state.

You can access that property here by retrieving the current state of the store

```ts
const isJsonPathsFeatureFlagTrue = useFeaturesStore.getState().features['typeahead-json']
```

src/views/Generator/RuleEditor/Typeahead/Typeahead.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { forwardRef, ComponentProps } from 'react'
+
+import { SuggestionMode, useTypeahead } from './useTypeahead'
+
+export type TypeaheadProps = ComponentProps<typeof TextField.Root> & {
+  options: string[]
+  mode: SuggestionMode
+}
+
+export const Typeahead = forwardRef<HTMLInputElement, TypeaheadProps>(
```
cristianoventura, line 11: I noticed some styling issues when scrolling a long list of suggestions.

Image

src/views/Generator/RuleEditor/Typeahead/useTypeahead.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const handleChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
+    const v = e.target.value ?? ''
+    if (!isControlled) {
+      setInputValue(v)
+    }
+    extOnChange?.(e)
+    const shouldFocus =
+      (mode === 'onDot' && v.includes('.')) ||
+      (mode === 'onFirstKey' && v.length >= 1) ||
+      (mode === 'onThirdKey' && v.length >= 3)
```
cristianoventura: It seems like this logic overlaps with `shouldSuggest`. What do you think about combining them?

src/utils/harToProxyData.ts
MoAqsam: So, there might be some caveat here, are you guys enabling feature flags through a UI ?

I would refrain from polluting util function with state specific calls.
 
So depending on if the UI is triggering it or, if you expect during hot reload (updating the feature flag locally). This will not work either. 
 
I think the way im doing it right now is enabling the feature flag, then running k6. Which is working, as i've tested several recordings.


 Let me know! These may be related to heuristics for feature flagging, perhaps im missing context.

src/views/Generator/RuleEditor/Typeahead/useTypeahead.tsx
MoAqsam: Yep, did a move refactor on the logic and invoked it for both variables [here](https://github.com/grafana/k6-studio/pull/678/commits/5ebb8a533c33be2df6392d7c29734a0f7f604337).

src/views/Generator/RuleEditor/Typeahead/Typeahead.tsx
MoAqsam: Yes, fixed [here](https://github.com/grafana/k6-studio/pull/678/commits/467ea1b78a9e8dcc7809b0b01435ce169f2af3c7).

image


**Context**
Issue was the maxHeight only applied the background on the parent anchor. I had to add a element in between to prevent the parent boxing from causing issues in the child one.

src/utils/harToProxyData.ts
cristianoventura: Yes, we use [use-konami](https://github.com/chris-feist/use-konami?tab=readme-ov-file#use-konami) with the default sequence to activate it: ↑ ↑ ↓ ↓ ← → ← → b a

Hot reload will maintain the feature flag enabled since it's in the application state. (unless you do a window refresh, which is expected because the state is reloaded by then. Since feature flags are only used in Dev builds, this behaviour is okay)

> I would refrain from polluting util function with state specific calls.

Another suggestion would be moving the call to an upper scope. Either way, the state is the source of truth of feature flags during runtime since they are usually enabled through the UI.

MoAqsam: Ah, ok, good to know. I wasn't aware of this, alright i've updated it with your suggestion as i don't want to add a layer of refactoring. So, i'll use the state directly for the time being.

https://github.com/grafana/k6-studio/pull/678/commits/120f8e2b4065ef8b25f7b42f1bde9258d5dd2e96

cristianoventura:
Hi @MoAqsam - The team has identified some UX improvements before considering this production-ready https://github.com/grafana/k6-studio/issues/773.

We will merge this PR soon as it's behind a feature flag, and it'll be promoted as soon as those issues are addressed. Feel free to tackle it on a separate PR if you're interested.

MoAqsam:
> Hi @MoAqsam - The team has identified some UX improvements before considering this production-ready #773.
> 
> We will merge this PR soon as it's behind a feature flag, and it'll be promoted as soon as those issues are addressed. Feel free to tackle it on a separate PR if you're interested.

Understood, i might pick it up!

MoAqsam:
> @MoAqsam Apologies for the back and forth. We had an issue with CI, and it was addressed in [a862ad3](https://github.com/grafana/k6-studio/commit/a862ad3cf93484c2a9bd8b852d5a38c2451c11d9).
> 
> Could you please pull `main` again? 🙏

Yes, boss it has been done! 🗜️

Issue #656, opened 2025-04-14
Use custom element instead of div for in-browser UI


allansson:
# What?

We currently mount the in-browser UI in a div element in the body. This can interfere with the selector generator, causing it to generate a selector that won't work without the UI mounted. For example:

```html
<body>
  <div></div> <!-- Our mount -->
  <div>
    This is some user content.
  </div>
</body>
```

The selector generator might generate `body div:nth-of-type(2)`, which would select the wrong element when run in a script. To lessen the risk we could append the mount _after_ all the content, but this is at odds with loading the UI as fast as possible.

To get around the `nth-of-type` problem we should use a [custom element](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements) e.g. `ksix-studio-mount`, that decreases the risk of weird interactions with the original DOM.

Another improvement would be to make sure our mount is always at the bottom, but this might require us to monitor DOM changes and move the mount to the bottom. It's possible that React won't like this.

# Criteria for considering it done

The mount is using a custom element. Optionally, the mount is continuously moved to be the last element of the body.

allansson:
Turns out `customElements` is not available in content scripts. If we want something like this, then we need to define it by running `executeScript` and that is a bigger undertaking.

PR #684, opened 2025-04-22
internal(browser): In-browser UI breaks selector generation

allansson:
## Description

This PR partially addresses https://github.com/grafana/k6-studio/issues/656. I've outlined the problem in a code comment, but I'll paste it here:

Take the following DOM:

```
<body>
  <div>User element</div>
  <div id="ksix-studio-mount"></div>
  <div>Dynamically added later</div>
</body>
```

If the user was highlighting the dynamically added element, the selector generator could generate a selector like `body > div:nth-child(3)`. But running the generated script would always result in an error because the mount is only present when recording and the correct selector should have been `body > div:nth-child(2)`.

## How to Test

1. Start a recording.
2. Open devtools in the browser
3. Open Elements tab
4. Open e.g. grafana.com
5. You should see a `<div>` element be moved to the button as the page loads.
6. Try all of the tools to make sure that everything works even after the element has been moved.

## Checklist

### DIFF
--- a/extension/src/frontend/view/index.tsx
+++ b/extension/src/frontend/view/index.tsx
@@ -10,11 +10,45 @@ import { InBrowserControls } from './InBrowserControls'
 
 let initialized = false
 
-function createShadowRoot() {
+function createMount() {
   const mount = document.createElement('div')
 
-  document.body.prepend(mount)
+  document.body.appendChild(mount)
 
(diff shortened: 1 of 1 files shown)


cristianoventura (approved):
Thanks for the thorough explanation in the code. LGTM! 🚀

Issue #594, opened 2025-03-25
Recording control overlay


going-confetti:
Add an always-on-top window to control the recording:
- Stop recording
- Add new group (whether the user should be able to rename it right from the window is to be confirmed)

PR #688, opened 2025-04-23
feat: Add stop recording button to in-browser controls

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/594

## Description

Add a "Stop" button to the in-browser controls so that the user wouldn't have to switch back to the app to stop the recording.

## How to Test
* Start a new recording
* Click Stop in the in-browser controls
* Verify that the browser is closed and the recordign is stopped

## Checklist

image

### DIFF
--- a/extension/src/frontend/view/ToolBox.tsx
+++ b/extension/src/frontend/view/ToolBox.tsx
@@ -62,6 +62,7 @@ export function ToolBox({
           <StopIcon />
         </Toolbar.Button>
       </Tooltip>
+      <Toolbar.Separator />
       <Toolbar.ToggleGroup
         type="single"
         value={tool ?? ''}


src/components/primitives/Toolbar.tsx
```diff
@@ -1,7 +1,16 @@
...
+const Button = styled(RadixToolbar.Button)`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  cursor: pointer;
+  color: var(--studio-foreground);
+  background-color: var(--studio-toggle-bg-off);
+  border: none;
+  padding: var(--studio-spacing-2);
+`
```
going-confetti, line 13: Style-wise, it's a cope of `ToggleItem`

extension/src/frontend/view/ToolBox.tsx
```diff
@@ -54,6 +57,11 @@ export function ToolBox({
         border: 1px solid var(--gray-6);
       `}
     >
+      <Tooltip delayDuration={0} asChild content="Stop recording">
+        <Toolbar.Button onClick={onStopRecording}>
+          <StopIcon />
+        </Toolbar.Button>
+      </Tooltip>
```
cristianoventura, line 64: Any thoughts on adding a separator between the Stop and the Inspect buttons?

cristianoventura (approved):
LGTM!

extension/src/frontend/view/ToolBox.tsx
allansson: Agree with Cristiano. ☝️

allansson (dismissed):
Started recording. Clicked button. Magic! 💯 

Maybe there should be some more friction to avoid accidentally stopping the recording e.g. a small popover with an extra button to click, but we can iterate on that.

extension/src/frontend/view/ToolBox.tsx
going-confetti: Oh, it "got lost" when I moved the button to the left. Thanks a lot!

going-confetti:
> Maybe there should be some more friction to avoid accidentally stopping the recording e.g. a small popover with an extra button to click, but we can iterate on that.

Yes, I think this is worth exploring further, I'll create a new ticket to track future work on improving the controls and will mention it there

allansson (approved):
💯

e-fisher (commented):
Works great! I've noticed we don't display the toolbar (and stop button) if browser recording is disabled in settings, should we display the toolbar with just the "stop" button in that case?

allansson:
> Works great! I've noticed we don't display the toolbar (and stop button) if browser recording is disabled in settings, should we display the toolbar with just the "stop" button in that case?

@e-fisher This is tricky to do. The setting is there to be able to disable it if something is messed up by the extension. And the stop button might mess things up as well.

Also, we'd need to have some way of telling the extension that it should show only the stop button but not record anything which we don't have atm. The extension just starts recording asap to not miss any events.

e-fisher (approved):
LGTM! 🚀

PR #689, opened 2025-04-23
internal(browser): Add toggle for browser events to start recording page

allansson:
## Description

This PR adds a button to toggle the browser recording feature when starting a recording. 

## How to Test

1. Start a recording with "Capture browser events" enabled
    * The in-browser UI should show and browser events should be recorded.
2. Start a recording with "Capture browser events" disabled
    * The in-browser UI should not be available and no events should be recorded.

## Checklist

### DIFF
--- a/src/handlers/browser/launch.ts
+++ b/src/handlers/browser/launch.ts
@@ -12,10 +12,11 @@ import os from 'os'
 import path from 'path'
 import { promisify } from 'util'
 
-import { BrowserHandler } from './handlers/browser/types'
-import { getCertificateSPKI } from './proxy'
-import { BrowserServer } from './services/browser/server'
-import { getPlatform } from './utils/electron'
+import { getCertificateSPKI } from '../../proxy'
+import { BrowserServer } from '../../services/browser/server'
+import { getPlatform } from '../../utils/electron'
+
+import { BrowserHandler, LaunchBrowserOptions } from './types'
 
 const createUserDataDir = async () => {
   return mkdtemp(path.join(os.tmpdir(), 'k6-studio-'))
@@ -53,7 +54,7 @@ function getExtensionPath() {
 export const launchBrowser = async (
   browserWindow: BrowserWindow,
   browserServer: BrowserServer,
-  url?: string
+  { url, capture }: LaunchBrowserOptions
 ) => {
   const path = await getBrowserPath()
   console.info(`browser path: ${path}`)
@@ -73,7 +74,7 @@ export const launchBrowser = async (
   const extensionPath = getExtensionPath()
   console.info(`extension path: ${extensionPath}`)
 
-  if (k6StudioState.appSettings.recorder.enableBrowserRecorder) {
+  if (capture.browser) {
     browserServer.start(browserWindow)
   }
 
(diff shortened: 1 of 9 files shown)


src/views/Recorder/Recorder.tsx
```diff
@@ -95,7 +97,7 @@ export function Recorder() {
     try {
       setRecorderState('saving')
 
-      if (proxyData.length === 0) {
+      if (proxyData.length === 0 && browserEvents.length === 0) {
```
allansson, line 100: I did this while experimenting with the toggle for network capture, but I'm keeping it because it fixes another issue that I discovered: if you clear the request log and stop the recording, no recording will be created even if you have browser events.

src/handlers/browser/launch.ts
allansson, line 1: I moved this file into the `handlers/browser` folder because it's only used by the handler and co-locates it with the `LaunchBrowserOptions` type.

src/views/Recorder/EmptyState.tsx
```diff
@@ -32,9 +40,15 @@ const RecorderEmptyStateSchema = z.object({
 type RecorderEmptyStateFields = z.infer<typeof RecorderEmptyStateSchema>
 
 export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
+  const { data: settings } = useSettings()
+
   const proxyStatus = useProxyStatus()
   const { data: isBrowserInstalled } = useBrowserCheck()
 
+  const [captureBrowser, setCaptureBrowser] = useState(true)
```
allansson, line 48: I didn't use `react-form-hooks` here because doing so with the `Checkbox` control was just way to much work for something so simple.

src/views/Recorder/EmptyState.tsx
```diff
@@ -32,9 +40,18 @@ const RecorderEmptyStateSchema = z.object({
 type RecorderEmptyStateFields = z.infer<typeof RecorderEmptyStateSchema>
 
 export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
+  const { data: settings } = useSettings()
+
   const proxyStatus = useProxyStatus()
   const { data: isBrowserInstalled } = useBrowserCheck()
 
+  const [captureBrowser = true, setCaptureBrowser] = useLocalStorage(
```
allansson, line 48: `useLocalStorage` returns `T | undefined` even if you pass an initial value. Doing `captureBrowser = true` is just pleasing the type checker.

Llandy3d (dismissed):
Reviewing from the functionality standpoint, works well! 🙌

src/views/Recorder/EmptyState.tsx
```diff
@@ -85,48 +112,97 @@ export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
...
-            </Button>
-          </Flex>
-        </FieldGroup>
-        <WarningMessage
-          proxyStatus={proxyStatus}
-          isBrowserInstalled={isBrowserInstalled}
-        />
+              <Text size="2">
+                <Flex gap="2" align="center">
+                  <Checkbox
```
cristianoventura, line 166: We can move the `Flex` inside the `Text` component so the label is clickable. I think this improves the UX and makes it consistent with the behaviour of other checkboxes in the app.

https://github.com/user-attachments/assets/609f30cc-10b1-46d0-9814-282ced96de65

cristianoventura (approved):
Works as intended! 🚀  One non-blocking comment that can be addressed later.

src/views/Recorder/EmptyState.tsx
allansson: Good catch! It was working that way, but I was changing things around last minute and broke it. Will fix.

cristianoventura (approved):
LGTM!

PR #690, opened 2025-04-23
refactor(Split main): Move app handlers out of main.ts

cristianoventura:
## Description

This PR is the last one from the series "Move handlers out of main.ts" and addresses the `app` handlers and preload functions. With this final change, we no longer have a `main.ts` file with IPC handlers.

One more PR is coming up that moves some files from the root folder to the `./src/main` folder, making the organization a bit clearer.

## How to Test

- Check that the splashscreen shows and closes as expected
- Modify a generator and close the app without saving. Check that the app prompts you to save before closing

## Checklist

### DIFF
--- a/src/handlers/app/types.ts
+++ b/src/handlers/app/types.ts
@@ -1,0 +1,5 @@
+export enum AppHandler {
+  CLOSE = 'app:close',
+  CHANGE_ROUTE = 'app:change-route',
+  SPLASHSCREEN_CLOSE = 'splashscreen:close',
+}
--- a/src/handlers/app/types.ts
+++ b/src/handlers/app/types.ts
@@ -1,5 +1,5 @@
 export enum AppHandler {
   CLOSE = 'app:close',
   CHANGE_ROUTE = 'app:change-route',
-  SPLASHSCREEN_CLOSE = 'splashscreen:close',
+  SPLASHSCREEN_CLOSE = 'app:splashscreen-close',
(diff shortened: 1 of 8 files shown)


src/handlers/app/types.ts
```diff
@@ -0,0 +1,5 @@
+export enum AppHandler {
+  CLOSE = 'app:close',
+  CHANGE_ROUTE = 'app:change-route',
+  SPLASHSCREEN_CLOSE = 'splashscreen:close',
```
going-confetti, line 4: With the enum in place, it's probably less important now, but shouldn't this be `app:splashscreen-close`?

src/preload.ts
```diff
@@ -1,7 +1,8 @@
 // TODO: https://github.com/grafana/k6-studio/issues/277
 /* eslint-disable @typescript-eslint/no-unsafe-return */
```
going-confetti, line 2: Looks like this can be removed now

going-confetti (approved):

src/preload.ts
cristianoventura: Good catch, I missed that!

src/handlers/app/types.ts
cristianoventura: That makes sense, I changed to `app:splashscreen-close` now!

going-confetti (approved):
LGTM!

PR #701, opened 2025-04-24
chore: Lock node version in package.json

cristianoventura:
## Description

While setting up a fresh environment for testing, I encountered an error with the most recent version of Node.js when running `npm start`, preventing the development environment from starting.

```sh
Failed to load: /www/k6-studio/forge.config.ts
(node:71132) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:71132) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///www/k6-studio/forge.config.ts is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /www/k6-studio/package.json.
```

This PR locks the Node.js version on `22.0.0`, ensuring only the LTS version can run `npm install`.

```sh
$ nvm use 23.11.0
$ npm install

npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'k6-studio@1.2.0',
npm warn EBADENGINE   required: { node: '^22.0.0' },
npm warn EBADENGINE   current: { node: 'v23.11.0', npm: '10.9.2' }
npm warn EBADENGINE }
```

## Checklist


package.json
```diff
@@ -5,6 +5,9 @@
   "description": "k6 Studio",
   "repository": "github:grafana/k6-studio",
   "main": ".vite/build/main.js",
+  "engines": {
+    "node": ">=20.0.0 <=23.0.0"
```
e-fisher, line 9: This config would still allow using exact v23.0.0, which may not be the best, because it won't have any minor or patch updates.

Another question, if we actually need to support this wide range of versions, could we lock it at current LTS (v22) instead? This would also involve updating .nvmrc and git workflows to use v22.

```suggestion
    "node": "^22.0.0"
```

cristianoventura: That's a great suggestion! I just changed it to v22.0.0 in workflow files and package.json.

going-confetti (approved):

e-fisher (approved):

Issue #712, opened 2025-04-29
Add health check for k6 Studio proxy


cristianoventura:
## Description
k6 Studio is unable to establish a connection with the network if the operating system is configured to use a proxy. In this case, additional configuration is needed. We need to implement a health check that reports back to the UI whether or not the proxy was able to reach the user's network, informing them to review their proxy configuration.

## Considerations

- This usually happens when the browser reports a 502 error
- 502 errors can also be reported from server side so it may not be reliable to check only for the status code

## Acceptance criteria

- Perform a health check when k6 Studio proxy launches
- Change proxy status if the health check fails
- Inform user about possible misconfiguration with proxy settings

PR #716, opened 2025-05-05
feat: Implement proxy health check

cristianoventura:
## Description

Some users have faced issues where k6 Studio proxy can't reach their network (because of a proxy misconfiguration or upstream issue with a corporate network). This makes network requests fail without telling the user what the problem is.

This PR implements health checks right after k6 Studio proxy launches. Using the current proxy configuration the user has, the proxy will attempt to connect to the internet, and the status will change to "Unhealthy" upon failure.

## How to Test

There are a couple of ways to reproduce the healthy and unhealthy status of the proxy.

### Option 1. Launch the proxy in regular mode

This should always return healthy, unless internet connection can't be established.

### Option 2. Connect to upstream instance of mitmproxy _with a valid certificate_

- In k6 Studio working directory, launch a new instance of mitmproxy. It's important that the `confdir` parameter points to the `certificates` folder.

```sh
mitmdump --mode regular@7005 --set confdir=./resources/certificates
```

- In k6 Studio, change proxy mode to "upstream"
- Set "Server URL" to `localhost:7005` (or the port you launched mitmdump)
- Set "Certificate path" to the `mitmproxy-ca-cert.pem` located in the `./resources/certificates` folder
- You can also unset the certificate path but check "Skip SSL verification"

### Option 1. Specify a nonexistent upstream address

- In k6 Studio, change proxy mode to "upstream"
- Set "Server URL" to an address that doesn't exist (e.g. `localhost:9999`)

### Option 2. Connect to upstream instance of mitmproxy _without a certificate_

- In k6 Studio working directory, launch a new instance of mitmproxy. It's important that the `confdir` parameter points to the existing `certificates` folder.

```sh
mitmdump --mode regular@7005 --set confdir=./resources/certificates
```

- In k6 Studio, change proxy mode to "upstream"
- Set "Server URL" to `localhost:7005` (or the port you launched mitmdump)
- Do not set "Certificate path"
- Do not check "Skip SSL verification"

## Checklist

image

image

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/712


src/main/proxy.ts
```diff
@@ -257,3 +263,55 @@ export const cleanUpProxies = async () => {
...
+
+const getProxyCertificateContent = () => {
+  const certPath = expandHomeDir(getProxyCertificatePath())
+  if (certPath && existsSync(certPath)) {
+    return readFileSync(certPath)
+  }
+  return undefined
+}
+
+const checkProxyHealth = async () => {
```
going-confetti, line 291: Nit: I think `async` is redundant here

src/views/Recorder/EmptyState.tsx
```diff
@@ -293,6 +296,30 @@ function WarningMessage({
...
+  if (proxyStatus === 'unhealthy') {
+    return (
+      <Callout.Root>
+        <Callout.Icon>
+          <MinusCircle />
+        </Callout.Icon>
+        <Callout.Text>
+          <strong>Proxy is unhealthy</strong>
+          <br />
+          k6 Studio was unable to reach your network. If you have issues
```
going-confetti, line 308: We use the full name in other parts of the UI, so I'd rename it to `Grafana k6 Studio` for consistency reasons.

going-confetti: We check if the proxy can reach quickpizza, so is "was unable to reach your network" factually correct?

Considering we still allow the user to proceed, maybe we should approach this from a different POV: instead of asking the user to try recording and see if they have any issues it tells them to first "verify" that the health check failure is the expected behavior.
Maybe _something_ like this:
```tsx
<>
  Grafana k6 Studio cannot establish connection to the Internet.
  Unless this is expected due to your internal network configuration, check <button>proxy settings</button> or learn more in the <a>troubleshooting guide</a>.
</>
```

src/main/proxy.ts
```diff
@@ -257,3 +263,55 @@ export const cleanUpProxies = async () => {
...
+      agent,
+      ca: certContent,
+      headers: {
+        'k6-Studio-Health-Check': 'true',
+      },
+      rejectUnauthorized: !k6StudioState.appSettings.proxy.sslInsecure,
+    }
+
+    https
+      .get('https://quickpizza.grafana.com', options, (res) => {
```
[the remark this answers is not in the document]
cristianoventura: Thanks for the suggestion! As you mentioned, relying on only one URL may not be ideal, as it could introduce a false unhealthy state when the page isn't accessible.

I looked into the library and think the implementation looks good, except for proxy support, which relies on [global-agent](https://github.com/gajus/global-agent). I'm a bit reluctant to use that approach for a couple of reasons:

- it hasn't been recently maintained
- it globally changes the Node.js [HTTP agent](https://nodejs.org/api/http.html#http_http_request_options_callback) object to use a proxy. While we are not widely using this, it's one more thing we'd have to control/reset for specific scenarios
- some requests rely on DNS lookups that may not be accessible on corporate networks

> We could also implement our own similar lightweight solution, by calling multiple URLs

How would it sound if we used the current implementation to check for multiple URLs instead of just `quickpizza.grafana.com` and, if one of these URLs are successful, we can consider the proxy healthy?

cristianoventura (commented):
> Restarting the proxy twice results in unhealthy status with AggregateError, attaching a video

@e-fisher I wasn't able to reproduce this (that error message only shows up when I intentionally make the health check fail, as a result of not being able to reach the network). Was there any proxy configuration you had set other than the default?

src/main/proxy.ts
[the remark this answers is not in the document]
e-fisher: > How would it sound if we used the current implementation to check for multiple URLs instead of just quickpizza.grafana.com and, if one of these URLs are successful, we can consider the proxy healthy?

Yeah, I think that should do just fine 👍

e-fisher:
> > Restarting the proxy twice results in unhealthy status with AggregateError, attaching a video
> 
> @e-fisher I wasn't able to reproduce this (that error message only shows up when I intentionally make the health check fail, as a result of not being able to reach the network). Was there any proxy configuration you had set other than the default?

No custom configs, except I had port set to 7000, but changing to 6000 didn't solve it. @going-confetti could you check if you are able to reproduce this?

https://github.com/user-attachments/assets/cdb1a763-e55e-41c1-b624-014536d895dd

src/main/healthCheck.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { getProxyCertificateContent, getProxyURL } from './proxy'
+
+export const checkProxyHealth = async () => {
+  try {
+    const results = await Promise.allSettled([
+      isUrlReachable('https://www.google.com/generate_204'),
+      isUrlReachable('https://quickpizza.grafana.com'),
+      isUrlReachable('https://k6.io'),
+    ])
```
cristianoventura, line 13: @e-fisher I'm now using three distinct domains for the health check.

`https://www.google.com/generate_204` is pinged by android devices to check for connectivity and I ended up leaving `quickpizza.grafana.com` as it's lightweight as well.

Any ideas of domains we could include/remove?

src/main/healthCheck.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const isUrlReachable = async (url: string) => {
+  return new Promise((resolve) => {
+    const certContent = getProxyCertificateContent()
+    const agent = new HttpsProxyAgent(getProxyURL())
+    const options: https.RequestOptions = {
+      agent,
+      ca: certContent,
+      headers: {
+        'k6-Studio-Health-Check': 'true',
```
allansson, line 32:
```suggestion
        'X-K6-Studio-Health-Check': 'true',
```
Since the request will be hitting the server, it might be nice to follow convention.

src/main/healthCheck.ts
```diff
@@ -0,0 +1,50 @@
...
+      (r) => r.status === 'fulfilled' && r.value === true
+    )
+    return isProxyHealthy
+  } catch (error) {
+    log.error('Error checking proxy health:', error)
+    return false
+  }
+}
+
+const isUrlReachable = async (url: string) => {
```
allansson, line 24:
```suggestion
const isUrlReachable = (url: string) => {
```

allansson (commented):
I noticed that the state of the health check doesn't get reset when I navigate away from the recording view and then go back.

1. Restart the proxy.
2. Open the Start recording view
3. Disable your internet connection.
4. Move to another view.
5. Open the Start recording view again.
    - No message appears
6. Restart the proxy.
7. Enable your internet connection.
8. Move to another view
9. Open the Start recording view again.
    - The error message is still there

My suggestion is that we poll for the proxy health from the start recording view so that we can clear the message if internet connection returns:

```ts
const { data: isHealthy } = useQuery({
  queryKey: ["proxy-status"],
  queryFn: () => window.studio.proxy.checkHealth(),
  refetchInterval: 2000
})
```

And in the health check code we can retry a few times:

```ts
for (let i = 0; i < 3; ++i) {
  const isHealty = checkHealth()
  
  if (isHealthy) {
    return true
  }
  
  await timeout(500)
}

return false
```

src/main/proxy.ts
```diff
@@ -191,11 +190,9 @@ export const launchProxyAndAttachEmitter = async (
   k6StudioState.proxyEmitter.emit('status:change', 'starting')
 
   return launchProxy(browserWindow, k6StudioState.appSettings.proxy, {
-    onReady: async () => {
-      const isProxyHealthy = await checkProxyHealth()
-      const proxyStatus = isProxyHealthy ? 'online' : 'unhealthy'
+    onReady: () => {
       k6StudioState.wasProxyStoppedByClient = false
-      k6StudioState.proxyEmitter.emit('status:change', proxyStatus)
+      k6StudioState.proxyEmitter.emit('status:change', 'online')
```
cristianoventura: I reverted this back to `online` as the default state. The health check polling will be responsible for changing it if a connection can't be established.

src/main/healthCheck.ts
```diff
@@ -48,3 +48,23 @@ const isUrlReachable = async (url: string) => {
...
+export const startHealthCheckPolling = () => {
+  setInterval(async () => {
+    // Don't check if proxy is offline or starting
+    // Don't check if a recording is in progress
+    // Don't check if not in the /recorder page
+    if (
+      !k6StudioState.currentProxyProcess ||
+      ['starting', 'offline'].includes(k6StudioState.proxyStatus) ||
+      k6StudioState.currentBrowserProcess ||
+      !k6StudioState.currentClientRoute.startsWith('/recorder')
```
cristianoventura, line 61: We could have the health check working across all routes, but I think it'd be too much, so I restricted it to these checks instead.

This means that if there's something wrong with the proxy connection, the user won't be able to know until they go to the Recorder page, which is okay for now since it's the only place using the proxy.

cristianoventura (commented):
Thanks for the feedback! 🙏 

I've implemented something similar as you suggested, but I moved the logic to the main process instead because it'd be easier to control the proxy status from there since it's already being propagated to the client side. It also avoids some back and forth with IPC calls.

This solves:

- when internet connection changes, the proxy status and health check should fail in the "Recorder" page
- race conditions with the proxy status being unhealthy, when it's actually not

Let me know what your thoughts are.

src/main/healthCheck.ts
```diff
@@ -0,0 +1,70 @@
...
+      ca: certContent,
+      headers: {
+        'X-K6-Studio-Health-Check': 'true',
+      },
+      rejectUnauthorized: !k6StudioState.appSettings.proxy.sslInsecure,
+    }
+
+    https
+      .get(url, options, (res) => {
+        console.log(`Proxy health check status: ${res.statusCode} for ${url}`)
```
allansson, line 39: Gentle reminder: remove this.

allansson (dismissed):
The new polling mechanic works like a charm. 🚀

e-fisher (commented):
The false-unhealthy issue I had is no longer reproducible with this approach 🙌 Few comments current implementation:

1. Should we also notify user about unhealthy proxy in validator page and validator dialog? They can't be used if the proxy is misconfigured or there's no internet connection.

2. While the polling provides most up-to-date connection status, it still keeps pinging when connection is healthy. I would probably go with useQuery to get health status from main process and update health check to throw when it fails to rely on useQuery retry mechanism. The hook could be reused in multiple places, it will deduplicate requests automatically and do a background refetch when visiting recorder/validator pages to get the latest health status. The drawback of this is that health status would be separate from other statuses (online/starting/offline), but this would allow removing polling and route conditions from health check function. @allansson since polling was your suggestion - comment if you disagree on this

allansson:
> While the polling provides most up-to-date connection status, it still keeps pinging when connection is healthy. I would probably go with useQuery to get health status from main process and update health check to throw when it fails to rely on useQuery retry mechanism. The hook could be reused in multiple places, it will deduplicate requests automatically and do a background refetch when visiting recorder/validator pages to get the latest health status. The drawback of this is that health status would be separate from other statuses (online/starting/offline), but this would allow removing polling and route conditions from health check function. @allansson since polling was your suggestion - comment if you disagree on this

@e-fisher

If I understand you correctly, the hook would only poll until the first successful ping. The drawback is that it wouldn't detect if the network connection is dropped, but maybe that's acceptable. Using `useQuery` means that it would re-trigger as you navigate the app so you would see it eventually. 🤔

src/main/healthCheck.ts
going-confetti: I would probably remove k6.io or replace it with something else, considering its future is uncertain

src/views/Generator/ValidatorDialog.tsx
```diff
@@ -79,6 +91,14 @@ export function ValidatorDialog({
               <Flex align="center">
                 Validator
                 {isRunning && <Spinner ml="2" />}
+                {!isProxyHealthy && (
+                  <Tooltip content="Grafana k6 Studio cannot establish connection to the Internet. Unless this is expected due to your internal network configuration, check your proxy settings.">
+                    <Badge color="orange" ml="2">
+                      <TriangleAlert />
+                      Proxy health check failed
+                    </Badge>
+                  </Tooltip>
```
cristianoventura, line 100: We don't have an idle state in this component, so we opted for using a Badge component instead of the Callout. Any thoughts?

image

cristianoventura (commented):
I removed the health check polling and moved the logic to the client side. The hook is being called on `Recorder`, `Validator`, and `ValidatorDialog`.

A few trade-offs were made with this approach, such as being unable to detect if internet connection is lost without navigating to other pages and dropping support for the "Unhealthy" proxy status on the sidebar.

@e-fisher I was finally able to reproduce an instance of `AggregateError`, only when pointing the upstream server to a nonexistent address, such as `http://localhost:9999`.

Here are some scenarios for testing the latest changes:

### Health check when connecting to the internet

1. Disconnect your internet
2. Go to the Recorder page
3. Unhealthy warning is displayed ⚠️ 
4. Connect to the internet
5. A new health check is performed and Warning is gone ✅ 

### Disconnecting from the internet

1. Navigate to any page other than the Recorder
2. Disconnect the internet
3. Go back to the Recorder page
4. Unhealthy warning is displayed ⚠️ 

### Changing proxy settings with the Recorder page open

1. With valid proxy settings, navigate to the Recorder page
2. Open settings and change to an invalid upstream address (either misconfigured or nonexistent such as `http://localhost:9999`)
3. Save settings
4. Unhealthy warning is displayed ⚠️

src/hooks/useProxyHealthCheck.ts
```diff
@@ -0,0 +1,27 @@
...
+  })
+
+  useEffect(() => {
+    const fetchProxyHealth = async () => {
+      if (proxyStatus === 'online') {
+        await refetch()
+      }
+    }
+    void fetchProxyHealth()
+  }, [proxyStatus, refetch])
```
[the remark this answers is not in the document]
e-fisher: I think it would be useful to have this as code comment ^

e-fisher (commented):
I've noticed the failed health check message is not displayed in recorder when skipping certificate verification, otherwise looking good!

screencapture 2025-05-19 at 11 27 04

cristianoventura:
> I've noticed the failed health check message is not displayed in recorder when skipping certificate verification, otherwise looking good!

@e-fisher Good catch! I've fixed that and added the code comment as you suggested.

image

e-fisher (approved):
LGTM 🚀

PR #732, opened 2025-05-15
chore: Add more flexible usage tracking service

going-confetti:
## Description
Replace the logic for tracking installations with a more flexible approach, that will allow us to track anonymous usage data (e.g. track the number of recordings created)

## How to Test
At this stage, there's only installation tracking, which is a bit tricky to test (you'll likely need to modify the code to do it).
Review the tests I added and verify that they pass.

## Checklist

- [x] I have updated the information about usage tracking
- [x] I have updated the usage collection endpoint

## Related PR(s)/Issue(s)
k6-docs PR: https://github.com/grafana/k6-docs/pull/1967


src/main.ts
```diff
@@ -187,7 +187,7 @@ app.whenReady().then(
     nativeTheme.themeSource = k6StudioState.appSettings.appearance.theme
     await createSplashWindow()
 
-    await sendReport(k6StudioState.appSettings.telemetry.usageReport)
+    UsageTracker.getInstance()
```
going-confetti: Currently, `getInstance` initializes the service if this hasn't happened yet, which is exactly how it's used here. An alternative would be having both `initialize()` and `getInstance()` - wdyt?

README.md
```diff
@@ -99,7 +99,7 @@ When opening an issue, please include a tail of your log file.
 
 ## Usage collection
 
-By default, Grafana k6 Studio sends an anonymous usage report so that we can track relevant information to be able to build the product making better data-driven decisions. Prioritizing the features that benefit the most and reducing the impact of changes.
+By default, Grafana k6 Studio sends anonymous usage reports so that we can track relevant information to be able to build the product making better data-driven decisions. Prioritizing the features that benefit the most and reducing the impact of changes.
 
 For more information visit the [documentation](https://grafana.com/docs/k6-studio/set-up/usage-collection/).
```
going-confetti, line 104: https://github.com/grafana/k6-docs/pull/1967

src/services/usageTracking/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const eventWithMetadata = await this.createEventWithMetadata(event)
+
+      try {
+        await this.sendEvent(eventWithMetadata)
+      } catch (error) {
+        log.error('Failed to send usage statistic event:', error)
+      }
+    })()
+  }
```
allansson: This is a case where I think using `then` and `catch` produces more readable code.

```suggestion
  static async track(event: UsageEvent) {
    if (!k6StudioState.appSettings.telemetry.usageReport) {
      return
    }
    
    this.createEventWithMetadata(event)
      .then(this.sendEvent)
      .catch(error => log.error('Failed to send usage statistic event:', error))
  }
```

src/services/usageTracking/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  UsageEventWithMetadata,
+} from './types'
+
+const TRACKING_URL = 'https://stats.grafana.org/k6-studio-usage-report'
+const INSTALLATION_ID_FILE = path.join(
+  app.getPath('userData'),
+  '.installation_id'
+)
+
+export class UsageTracker {
```
allansson: Subjective, but I'm not in love with using an all-static class. Classes and objects are containers of state + behavior, but this is only a named bundle of functions.

I would prefer just exporting the methods:

```ts
export async function initTracking() {}

export function trackEvent(event: UsageEvent) {}
```

src/services/usageTracking/types.ts
```diff
@@ -0,0 +1,61 @@
+import { Arch, Platform } from '@/types/electron'
+
+const PREFIX = 'grafana_k6_studio'
+
+export const UsageEventName = {
+  // General
+  AppInstalled: `${PREFIX}_app_installed`,
```
allansson: We could append the prefix before sending the event instead:

```ts
// src/services/usageTracking/index.ts
export class UsageTracker {
  // ...

  private static async sendEvent({ event, ...payload }: UsageEvent) {
    if (!k6StudioState.appSettings.telemetry.usageReport) {
      return
    }

    if (process.env.NODE_ENV === 'development') {
      console.log(event)
      return
    }
    
    const metadata = await this.createEventMetadata()
    
    const body = {
      event: `grafana_k6_studio_${event}`,
      ...payload,
      ...metadata
    }

    const response = await fetch(TRACKING_URL, {
      method: 'POST',
      headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify(body),
    })

    if (!response.ok) {
      throw new Error(`Failed to send event: ${response.statusText}`)
    }
  }

  // ...

  private static async createEventMetadata() {
    return {
      usageStatsId: await this.getInstallationId(),
      timestamp: Date(),
      appVersion: app.getVersion(),
      os: getPlatform(),
      arch: getArch(),
    }
  }
}

// src/services/usageTracking/types.ts
export interface ScriptRunInCloudEvent {
  event: "script_run_in_cloud" // Event names can be simplified
}
```

src/services/usageTracking/index.ts
going-confetti: Thanks, this does look more readable

src/services/usageTracking/index.ts
going-confetti: I think it's a pretty good argument. It wasn't all static in the early iterations, but at this point there's no need to have a class

src/services/usageTracking/types.ts
going-confetti: Done 👍

src/services/usageTracking/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  usageStatsId: string
+  timestamp: string
+  appVersion: string
+  os: Platform
+  arch: Arch
+}
+
+export interface AppInstalledEvent {
+  event: UsageEventName.AppInstalled
+}
```
going-confetti, line 18: More events will be added in a separate PR

src/services/usageTracking/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  UsageEvent,
+  UsageEventMetadata,
+  UsageEventName,
+  UsageEventWithMetadata,
+} from './types'
+
+export async function initEventTracking() {
+  if (process.env.NODE_ENV === 'development') {
+    return
+  }
```
allansson, line 19: I had to comment this out to get the event to be logged to console. Maybe it's not necessary since we block it in `sendEvent`?

allansson (approved):
Just a minor thing, but the functionality works like it should.

Llandy3d (approved):

src/services/usageTracking/index.ts
going-confetti: The logic is pretty much the same as before - If the installation id file already exists, we won't report the installation event, so this early return prevents generation of this file.

We could instead have a separate file for the dev env and even use the dev version of the usage stats endpoint - what do you think?

allansson: I don't think it matters, really. It was just an observation that in order to test the PR I had to remove those lines. Didn't know we wanted to prevent the file from being created. Let's merge as-is, and fix it later if it ever turns out to be a hot path for development. 😬

PR #742, opened 2025-06-05
fix(browser): Extension doesn't load when starting a recording in Chrome

allansson:
## Description

This PR fixes an issue where the browser recorder extension wouldn't load if the user was using Chrome 137+. The reason is that the `--load-extension` used to load the extension has been removed for security reasons.

## How to Test

In both Chrome and Chromium:

1. Start a recording with browser capture enabled.
2. Make sure the in-browser UI is visible.
3. Try to generate some browser events.
4. Check that they appear in k6 Studio
5. Start another recording with browser capture disabled.
6. The in-browser UI should not be shown and no events captured.

## Checklist

### DIFF
--- a/src/handlers/browser/launch.ts
+++ b/src/handlers/browser/launch.ts
@@ -2,7 +2,6 @@ import {
   computeSystemExecutablePath,
   Browser,
   ChromeReleaseChannel,
-  launch,
 } from '@puppeteer/browsers'
 import { exec, spawn } from 'child_process'
 import { app, BrowserWindow } from 'electron'
@@ -13,6 +12,7 @@ import path from 'path'
 import { promisify } from 'util'
 
 import { getCertificateSPKI } from '@/main/proxy'
+import { ChromeDevtoolsClient } from '@/utils/cdp/client'
 
 import { BrowserServer } from '../../services/browser/server'
 import { getPlatform } from '../../utils/electron'
@@ -52,6 +52,18 @@ function getExtensionPath() {
   return path.join(process.resourcesPath, 'extension')
 }
 
+const FEATURES_TO_DISABLE = [
+  'OptimizationGuideModelDownloading',
+  'OptimizationHintsFetching',
+  'OptimizationTargetPrediction',
+  'OptimizationHints',
+]
+
+const BROWSER_RECORDING_ARGS = [
+  '--remote-debugging-pipe',
+  '--enable-unsafe-extension-debugging',
+]
+
 export const launchBrowser = async (
   browserWindow: BrowserWindow,
   browserServer: BrowserServer,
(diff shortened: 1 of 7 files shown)


src/handlers/browser/launch.ts
```diff
@@ -52,6 +52,18 @@ function getExtensionPath() {
...
+  'OptimizationHintsFetching',
+  'OptimizationTargetPrediction',
+  'OptimizationHints',
+]
+
+const BROWSER_RECORDING_ARGS = [
+  '--remote-debugging-pipe',
+  '--enable-unsafe-extension-debugging',
+]
+
```
allansson, line 66: Trying to reduce some noise in the `launch` function.

src/handlers/browser/launch.ts
```diff
@@ -112,28 +110,36 @@ export const launchBrowser = async (
...
     `--ignore-certificate-errors-spki-list=${certificateSPKI}`,
+    `--disable-features=${FEATURES_TO_DISABLE.join(',')}`,
     ...browserRecordingArgs,
-    disableChromeOptimizations,
     url?.trim() || 'about:blank',
   ]
 
-  // if we are on linux we spawn the browser directly and attach the on exit callback
-  if (getPlatform() === 'linux') {
-    const browserProc = spawn(path, args)
```
allansson, line 122: All platforms now use `spawn` instead of puppeteer's `launch`, because Chrome uses file descriptors 3 and 4 for communicating over a pipe and I couldn't find a way to configure these for puppeteer.

Besides, it reduces the complexity of the code.

src/handlers/browser/launch.ts
```diff
@@ -52,6 +52,18 @@ function getExtensionPath() {
...
+const FEATURES_TO_DISABLE = [
+  'OptimizationGuideModelDownloading',
+  'OptimizationHintsFetching',
+  'OptimizationTargetPrediction',
+  'OptimizationHints',
+]
+
+const BROWSER_RECORDING_ARGS = [
+  '--remote-debugging-pipe',
+  '--enable-unsafe-extension-debugging',
```
allansson, line 64: Using `--enable-unsafe-extension-debugging` requires `--remote-debugging-pipe`.

src/utils/cdp/client.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Readable, Writable } from 'stream'
+
+import {
+  ChromeRequest,
+  ChromeRequestMap,
+  ChromeResponse,
+  ChromeResponseSchema,
+} from './types'
+
+export class ChromeDevtoolsClient {
```
allansson, line 11: A super simple client for the Chrome Devtools Protocol. I couldn't find one that communicated over a pipe which is required for `Extensions.loadUnpacked`.

cristianoventura: Was this written from scratch or based on an existing approach?

cristianoventura (approved):
LGTM!

Works on the newest and older versions.

- Chrome `137+` on macOS ✅ 
- Chromium `136` on macOS and Windows ✅

PR #748, opened 2025-06-12
feat: Display recent URLs in Recorder

going-confetti:
## Description
A small QoL improvement we discussed some time ago.

## How to Test
1. Start a recording (or a few) by providing a starting URL
2. Verify that when you open Recorder again, you see max 3 recent starting URLs

https://github.com/user-attachments/assets/46681fc8-d744-4490-bc78-2b2ee99a11b6

### DIFF
--- a/src/views/Recorder/RecentURLs.tsx
+++ b/src/views/Recorder/RecentURLs.tsx
@@ -26,6 +26,7 @@ export function RecentURLs({
         list-style: none;
         padding: 0;
         margin: 0;
+        line-height: 1;
         display: flex;
         flex-wrap: wrap;
         gap: var(--space-1) var(--space-3);
@@ -50,15 +51,6 @@ export function RecentURLs({
               position: relative;
               max-width: 100%;
               overflow: hidden;
-
-              &:hover {
-                background-color: var(--gray-4);
-              }
-
-              &:focus-within > button:last-of-type,
-              &:hover > button:last-of-type {
-                opacity: 1;
-              }
             `}
           >
             <Reset>
@@ -70,6 +62,12 @@ export function RecentURLs({
                   overflow: hidden;
                   text-overflow: ellipsis;
                   white-space: nowrap;
+                  font-weight: 500;
+                  color: var(--gray-11);
+
+                  &:hover {
+                    color: var(--orange-9);
+                  }
 
                   &::after {
                     content: '';
@@ -90,8 +88,9 @@ export function RecentURLs({
                     position: relative;
                     z-index: 1;
                     opacity: 0.6;
-                    margin: calc(var(--space-1) * -0.5)
-                      calc(var(--space-1) * -1.5);
+
+                    margin: calc(var(--space-1) * -0.5);
+                    margin-right: calc(var(--space-1) * -1.5);
                     padding: var(--space-1);
 
                     font-size: var(--font-size-1);
@@ -102,8 +101,8 @@ export function RecentURLs({
 
                     &:hover,
                     &:focus {
+                      color: var(--red-11);
                       opacity: 1;
-                      background: var(--gray-6);
                     }
                   `}
                   type="button"


src/views/Recorder/EmptyState.tsx
```diff
@@ -69,13 +73,17 @@ export function EmptyState({ isLoading, onStart }: EmptyStateProps) {
...
+  const { recentURLs, addURL, removeURL } = useRecentURLs()
+
   const canRecord = proxyStatus === 'online' && isBrowserInstalled === true
 
   const onSubmit = ({ url }: RecorderEmptyStateFields) => {
     if (isLoading || !canRecord) {
       return
     }
 
+    addURL(url)
```
going-confetti, line 85: It's not ideal, because technically the user can enter any string the browser will still start, yet we don't want that erroneous input to be listed. Could be a future improvement 🤔

src/views/Recorder/RecentURLs.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <ul
+      aria-label="Recent URLs"
+      css={css`
+        list-style: none;
+        padding: 0;
+        margin: 0 0 var(--space-3);
+        display: flex;
+        flex-wrap: wrap;
+        gap: var(--space-1) var(--space-3);
```
going-confetti, line 32: Different gaps so it still looks okay on line wrap:
image

src/views/Recorder/RecentURLs.tsx
```diff
@@ -0,0 +1,112 @@
...
+            <Reset>
+              <button
+                type="button"
+                aria-label="Select URL"
+                disabled={disabled}
+                css={css`
+                  flex: 1;
+                  overflow: hidden;
+                  text-overflow: ellipsis;
+                  white-space: nowrap;
```
going-confetti, line 64: Takes care of really long URLs even when the window size is small:
image

going-confetti (commented):
Would like to hear you opinion on the current state of this feature:
1. Is it clear enough that these are recent URLs?
2. Does it look okay when (some or all) URLs are long?
3. How big is the problem of potentially displaying invalid URLs?

src/views/Recorder/RecentURLs.tsx
```diff
@@ -0,0 +1,123 @@
...
+                    opacity: 0.6;
+                    margin: calc(var(--space-1) * -0.5)
+                      calc(var(--space-1) * -1.5);
+                    padding: var(--space-1);
+
+                    font-size: var(--font-size-1);
+                    & > .lucide {
+                      min-width: var(--font-size-1);
+                      min-height: var(--font-size-1);
+                    }
```
going-confetti: Not the prettiest piece of code for sure. It's here to make sure the click target is big enough, while the icon is still subtle

cristianoventura: Can some of these be moved as props to the icon component?

src/hooks/useRecentURLs.ts
```diff
@@ -0,0 +1,33 @@
...
+  )
+
+  const addURL = (url: string) => {
+    const trimmedURL = url.trim()
+
+    if (!trimmedURL || recentURLs.includes(trimmedURL)) {
+      return
+    }
+
+    const updatedURLs = [trimmedURL, ...recentURLs].slice(0, MAX_RECENT_URLS)
```
cristianoventura: If you input a URL that is already at the bottom of the list, I'd expect it to be moved to the top to be considered as "most recent".

Example:

- Recent list has: `k6.io`, `quickpizza.grafana.com`, `grafana.com` (in that order from most recent to older).
- If you input `grafana.com` again, the list should become `grafana.com`, `k6.io`, `quickpizza.grafana.com`.

```suggestion
    if (!trimmedURL) {
      return
    }

    const filteredURLs = recentURLs.filter((u) => u !== trimmedURL)
    const updatedURLs = [trimmedURL, ...filteredURLs].slice(0, MAX_RECENT_URLS)
```

src/hooks/useRecentURLs.ts
```diff
@@ -0,0 +1,33 @@
...
+const MAX_RECENT_URLS = 3
+
+export function useRecentURLs() {
+  const [recentURLs = [], setRecentURLs] = useLocalStorage<string[]>(
+    LOCAL_STORAGE_KEY,
+    []
+  )
+
+  const addURL = (url: string) => {
+    const trimmedURL = url.trim()
```
cristianoventura: As an improvement, we should consider normalizing the URL to lowercase to avoid issues with case sensitivity.

image

cristianoventura (approved):
LGTM!

I left some comments but overall it's a great UX improvement! 🎸

src/views/Recorder/RecentURLs.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  type="button"
+                  disabled={disabled}
+                  aria-label="Remove URL"
+                  onClick={() => onRemoveURL(url)}
+                >
+                  <XIcon />
+                </button>
+              </Reset>
+            </Tooltip>
+          </Badge>
```
allansson: That negative margin trickery irks me for two reasons:

- It's super hard to understand.
- It makes the button sit right next to the text.

Here's a version where we just reset the padding and gap on the badge and set it on the first button.

```suggestion
          <Badge
            size="1"
            variant="soft"
            color="gray"
            css={css`
              position: relative;
              max-width: 100%;
              overflow: hidden;

              padding: 0;
              gap: 0;

              &:hover {
                background-color: var(--gray-4);
              }

              &:focus-within > button:last-of-type,
              &:hover > button:last-of-type {
                opacity: 1;
              }
            `}
          >
            <Reset>
              <button
                type="button"
                disabled={disabled}
                css={css`
                  flex: 1;
                  overflow: hidden;
                  text-overflow: ellipsis;
                  white-space: nowrap;
                  
                  // The padding here gives just a _little_ bit of space on the
                  // right side of the URL text. Just enough so that it doesn't
                  // "touch" the text.
                  padding: calc(var(--space-1) * 0.5);
                  padding-left: calc(var(--space-1) * 1.5);

                  &::after {
                    content: '';
                    position: absolute;
                    inset: 0;
                  }
                `}
                onClick={() => onSelectURL(url)}
              >
                {url}
              </button>
            </Reset>
            <Tooltip content="Remove URL">
              <Reset>
                <button
                  css={css`
                    display: flex;
                    position: relative;
                    z-index: 1;
                    opacity: 0.6;
                    padding: var(--space-1);

                    font-size: var(--font-size-1);
                    & > .lucide {
                      min-width: var(--font-size-1);
                      min-height: var(--font-size-1);
                    }

                    &:hover,
                    &:focus {
                      opacity: 1;
                      background: var(--gray-6);
                    }
                  `}
                  type="button"
                  disabled={disabled}
                  aria-label="Remove URL"
                  onClick={() => onRemoveURL(url)}
                >
                  <XIcon />
                </button>
              </Reset>
            </Tooltip>
          </Badge>
```

allansson (approved):
This is going to save me so much sanity! 😍 

I left a comment on the styling, otherwise everything else looks good 🤌

cristianoventura (approved):
🚀

src/views/Recorder/RecentURLs.tsx
going-confetti: Apologies, I realized that I didn't reply to this comment.
I believe the current min-height/width of 16px was chosen after our initial testing showed that smaller icons looked worse. I'll try to open a separate PR this week to improve the DX though

allansson (approved):

PR #756, opened 2025-06-18
fix: Cloud sign-in fails when user is behind a network proxy

allansson:
## Description

This PR fixes an issue where network requests issued via `fetch` would fail when the user is behind a network proxy. The issue is caused by NodeJS not respecting the system proxy settings out of the box. The solution implemented here is to use the same upstream proxy settings as the recorder proxy.

A future improvement could be to load the system proxy settings from using `os-proxy-config` by default.

## How to Test

1. Set up a HTTP proxy server locally (I used [dev-proxy](https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/overview) but you could use mitmproxy as well)
2. Configure k6 Studio to use the proxy.
3. Try to sign in to cloud

The sign in should work as expected.

## Checklist


src/services/http.ts
```diff
@@ -0,0 +1,53 @@
...
+    return await readFile(path, 'utf-8')
+  } catch (error) {
+    log.error('Failed to read proxy certificate.', error)
+
+    return undefined
+  }
+}
+
+function getAuthorizationHeader({ username, password }: UpstreamProxySettings) {
+  if (username === undefined || password === undefined) {
```
allansson, line 24: The typing here is a bit unfortunate. The schema should have been:

```ts
const CredentialsSchema = z.object({
  username: z.string() // Not optional
  password: z.string() // Not optional
})

const UpstreamProxySettingsSchema = z.object({
  credentials: CredentialsSchema.optional() // Optional

  // no longer required.
  // requiresAuth: z.boolean()
})
```

But that would require a migration, so I'm biting the bullet.

src/services/http.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (path === undefined || path.trim() === '') {
+    return undefined
+  }
+
+  try {
+    return await readFile(path, 'utf-8')
+  } catch (error) {
+    log.error('Failed to read proxy certificate.', error)
+
+    return undefined
```
allansson, line 19: Right now if anything goes wrong reading the certificate, we just treat that as "there was no certificate configured". Not sure if that's the best option, but at least the user can read it in the logs.

src/services/http.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  const certificate = await readCertificate(proxy.certificatePath)
+
+  const agent = new ProxyAgent({
+    uri: proxy.url,
+    headers: getAuthorizationHeader(proxy),
+    requestTls: {
+      ca: certificate,
+      rejectUnauthorized: !proxy.sslInsecure,
```
cristianoventura, line 49: We should probably move the proxy settings to a more generic "Network" tab at some point.

Issue #604, opened 2025-03-26
Add "More Info" helper button to clarify JSON path syntax in correlation


cristianoventura:
When adding a correlation rule, we'd like to add a "more info" button for JSON path selectors to explain which flavor of JSON selectors is being used.

PR #759, opened 2025-06-20
feat: Add JSON path explanation

e-fisher:
## Description

Add an explanation of JSON path selector and examples how to use it. To add syntax highlighting, I've created CodeSnippet wrapper of monaco editor.

I didn't want to use a Tooltip component because of the dark background in light theme and having to hack the styling, so I've used Popover and created PopoverTooltip. That's what radix uses for tooltips in their documentation as well.

I've converted other form tooltips to PopoverTooltip for consistency. 

## Checklist

screencapture 2025-06-20 at 09 42 59

screencapture 2025-06-20 at 09 43 43

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/604


src/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,60 @@
...
+  // Disable monaco editor focus on mount
+  useEffect(() => {
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        if (document.activeElement instanceof HTMLElement) {
+          document.activeElement.blur()
+        }
+      })
+    })
+  }, [])
```
e-fisher, line 29: A workaround to disable autofocus when monaco editor is mounted, this does the trick, but if you know a better way - please let me know.

src/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, [value])
+
+  // Disable monaco editor focus on mount
+  function handleEditorMount() {
+    requestAnimationFrame(() => {
+      if (document.activeElement instanceof HTMLElement) {
+        document.activeElement.blur()
+      }
+    })
+  }
```
e-fisher: A workaround to disable autofocus when monaco editor is mounted, this does the trick, but if you know a better way - please let me know.

going-confetti: I faced a similar issue once, but eventually I managed to find a solution that didn't involve remounting the editor. I think while not ideal, this is probably a decent workaround.

Having said that, I believe Monaco is an overkill for read-only snippets like this – I can only think of 2 benefits:
1. We already have it as a dependency
2. The theme matches the one in the script preview/custom code editor

At the same time, looks like some hacking is needed to make it work okay (side-note: I wish there was a lightweight component with the same/similar API just for displaying code) + it's still a read-only editor, so the user can still put the cursor in there, which feels odd:
image

_____

If we were to build a lightweight syntax highlighting component, we could also use it inline so that small snippets (like `user.hobbies[1]` in this PR) could also have syntax highlighting/be easier to read.

What do you think? Did I miss something?

src/components/Form/FieldGroup.tsx
```diff
@@ -27,23 +29,25 @@ export function FieldGroup({
...
-              {label}
-            </Text>
+            <Label.Root htmlFor={name}>
+              <Text size="2" weight="medium">
+                {label}
+              </Text>
+            </Label.Root>
             {hint && hintType === 'tooltip' && (
-              <Tooltip content={hint}>
+              <PopoverTooltip content={hint}>
```
going-confetti, line 40: Making the tooltips toggleable/clickable seems like a bigger change, and I would prefer to get some feedback from @majavojnoska first – this might be something we want, but it changes the existing UX pattern.

If we do end up making them toggleable, I would suggest using an icon button, so that it's clear that the icon is clickable. Like this example from the radix theme docs:

https://github.com/user-attachments/assets/52b24a74-c6d5-4ba3-b5cc-87113b1fc889

e-fisher:
> I faced a similar issue once, but eventually I managed to find a solution that didn't involve remounting the editor. I think while not ideal, this is probably a decent workaround.

> Having said that, I believe Monaco is an overkill for read-only snippets like this – I can only think of 2 benefits:

> We already have it as a dependency
The theme matches the one in the script preview/custom code editor
At the same time, looks like some hacking is needed to make it work okay (side-note: I wish there was a lightweight component with the same/similar API just for displaying code) + it's still a read-only editor, so the user can still put the cursor in there, which feels odd:
image

>If we were to build a lightweight syntax highlighting component, we could also use it inline so that small snippets (like user.hobbies[1] in this PR) could also have syntax highlighting/be easier to read.

> What do you think? Did I miss something?

You got that right - monaco is already installed, it has our custom theme, and it supports dark/light theme switching. We could use a dedicated highlighting lib, but it would add another dependency and require additional time to set up, more so if we decide to build our own solution. Do you think showing cursor is a dealbreaker here? We could hide it with CSS.

src/components/Form/FieldGroup.tsx
e-fisher: Agree on using icon button. I've picked onClick solution because it's less intrusive and make it easier to select/copy text from tooltips, but let's see what's Maja's opinion on this.

e-fisher:
@going-confetti as discussed - I've updated popover tooltip to be displayed on hover


https://github.com/user-attachments/assets/fb9f90bf-1d25-4c9b-987f-2ccf6eb13b98

cristianoventura (approved):
I'm a bit late for the review and missed some of the discussion. After checking the latest changes, it LGTM! 🤩

going-confetti (approved):

PR #769, opened 2025-07-15
feat: Initial navigation event is sometimes not captured when starting a recording

allansson:
## Description

This PR attempts to fix an issue where the browser recorder wouldn't capture the initial navigation when starting a recording, resulting in a script that is unusable.

I have been able to reproduce the issue once, but not reliably. The likely culprit is that there is a race condition between the browser loading the page and us injecting the extension over CDP (which is why this issue hasn't appeared before since we previously used the `--load-extension` flag).

## How to Test

I'm not sure. It's (potentially) a race condition so it all comes down to luck.

## Checklist

### DIFF
--- a/src/handlers/browser/launch.ts
+++ b/src/handlers/browser/launch.ts
@@ -120,11 +120,17 @@ export const launchBrowser = async (
   })
 
   if (capture.browser) {
-    browserServer.start(browserWindow)
+    try {
+      await Promise.all([
+        new Promise<void>((resolve, reject) => {
+          process.on('spawn', resolve)
+          process.on('error', reject)
+        }),
+        browserServer.start(browserWindow),
+      ])
 
-    const client = ChromeDevtoolsClient.fromChildProcess(process)
+      const client = ChromeDevtoolsClient.fromChildProcess(process)
 
-    process.on('spawn', async () => {
       const response = await client.call({
         method: 'Extensions.loadUnpacked',
         params: {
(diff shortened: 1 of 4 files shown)


extension/src/background/routing.ts
```diff
@@ -10,9 +9,7 @@ const frontend = new BrowserExtensionClient('frontend', new PortTransport())
 
 const studio = new BrowserExtensionClient(
   'studio',
-  STANDALONE_EXTENSION
```
allansson, line 13: Removed this because running the extension standalone no longer works and this line of code affected my testing.

src/handlers/browser/launch.ts
```diff
@@ -154,20 +154,30 @@ export const launchBrowser = async (
...
   if (capture.browser) {
-    browserServer.start(browserWindow)
+    try {
+      await Promise.all([
+        new Promise<void>((resolve, reject) => {
+          process.on('spawn', resolve)
+          process.on('error', reject)
+        }),
+        browserServer.start(browserWindow),
+      ])
```
allansson: I'm attempting to make the startup sequence a bit more reliable:

1. First we wait for the browser process and websocket server to start.
2. Then we inject the extension

vite.extension.config.mts
```diff
@@ -66,7 +66,7 @@ export default defineConfig((env) => {
                 js: ['extension/src/frontend/index.ts'],
               },
             ],
-            permissions: ['webNavigation', 'storage'],
+            permissions: ['webNavigation', 'storage', 'tabs'],
```
allansson, line 69: The `tabs` permission is needed to read the `pendingUrl` and `url` properties on tabs.

Issue #637, opened 2025-04-09
Add assertions on input value


allansson:
# What?

Sometimes you need to assert that an input has a certain value, e.g. after auto-filling a form. Users should be able to add an assertion that checks the value of an input against a value specified by them.

There should be a tool that allows a user to pick an element and add an input value assertion.

# Criteria for considering it done

The user should be able to pick an input or textarea and add an assertion on its value. The event should be visible in the browser event log and a call to `expect().toHaveValue()` should be emitted in the script.

PR #770, opened 2025-07-15
feat: Add assertions of values in text inputs

allansson:
## Description

This PR adds the ability to assert on the value in text inputs (`input` and `textarea`).

## How to Test

1. Start a recording
2. Pick an input or textarea
3. Click "Add value assertion"
4. Stop the recording
5. Run the exported script

The script should contain a `toHaveValue` assertion and should pass/fail correctly.

## Checklist

## Related PR(s)/Issue(s)

Part of https://github.com/grafana/k6-studio/issues/637


extension/src/utils/aria.ts
```diff
@@ -17,32 +24,92 @@ function bySpecificity(a: IntrinsicRole, b: IntrinsicRole) {
   return a.specificity - b.specificity
 }
 
+type AttributePredicate = (element: Element) => boolean
+
+function getConstraintPredicate(
```
allansson, line 29: The existing implementation couldn't detect that `input` elements without a `type` attribute have the `textbox` role, so I rewrote it to something more robust.

Having multiple values comes into play when I add support for `select` elements.

src/codegen/browser/types.ts
```diff
@@ -98,10 +98,16 @@ export interface IsCheckedAssertion {
   expected: CheckState
 }
 
+export interface HasValuesAssertion {
+  type: 'has-values'
+  expected: [string, ...string[]]
+}
```
allansson, line 104: I don't make any distinction whether we're asserting a single value or multiple values here. I instead opted to select which method to use (`toHaveValue` or `toHaveValues`) when emitting the code. Branching here would just mean more code to achieve the same thing.

src/utils/list.ts
```diff
@@ -0,0 +1,8 @@
+export type NonEmptyArray<T> = [T, ...T[]]
+
+export function mapNonEmpty<T, U>(
```
allansson, line 3: Type safe function for mapping a non-empty array.

src/types/aria-query.d.ts
```diff
@@ -0,0 +1,395 @@
+// Adapted from flow types here: https://github.com/A11yance/aria-query/blob/main/flow/aria.js
+declare module 'aria-query' {
```
allansson, line 2: The available types for `aria-query` were outdated and had some key things missing, so I copied and adapted these from their Flow types.

allansson (commented):
Self-review

e-fisher (approved):
LGTM 🚀

PR #771, opened 2025-07-18
chore: Anonymous usage tracking for Recorder, Generator, Validator

going-confetti:
## Description
Track feature usage in Recorder, Generator, and Validator + track successful signins

> [!IMPORTANT]  
> Even though this is not user facing, we should include this in the release notes

## How to Test
- Start the app locally (`npm start`)
- Test one or multiple actions that should be tracked (the list of events can be found in `src/services/usageTracking/types.ts`
- You should see the event logged in the terminal

## Checklist

- [x] I have updated the usage reporting docs (https://github.com/grafana/k6-docs/pull/1967)

## Related PR(s)/Issue(s)

https://github.com/grafana/k6-docs/pull/1967

### DIFF
--- a/src/services/usageTracking/types.ts
+++ b/src/services/usageTracking/types.ts
@@ -3,6 +3,7 @@ import { Arch, Platform } from '@/types/electron'
 export enum UsageEventName {
   // General
   AppInstalled = 'app_installed',
+  UserLoggedIn = 'user_logged_in',
 
   // Recorder
   RecordingCreated = 'recording_created',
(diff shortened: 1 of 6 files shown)


src/services/usageTracking/types.ts
```diff
@@ -3,6 +3,20 @@ import { Arch, Platform } from '@/types/electron'
...
+  RecordingImported = 'recording_imported',
+
+  // Generator
+  GeneratorCreated = 'generator_created',
+  GeneratorUpdated = 'generator_updated',
+
+  // Script
+  ScriptExported = 'script_exported',
+  ScriptValidated = 'script_validated',
+  ScriptRunInCloud = 'script_run_in_cloud',
```
going-confetti, line 19: Let me know if something obvious is missing

src/services/usageTracking/types.ts
```diff
@@ -13,10 +27,51 @@ export interface UsageEventMetadata {
...
+export type UsageEvent =
+  | AppInstalledEvent
+  | UserLoggedInEvent
+  | RecordingCreatedEvent
+  | RecordingImportedEvent
+  | GeneratorCreatedEvent
+  | GeneratorUpdatedEvent
+  | ScriptExportedEvent
+  | ScriptValidatedEvent
+  | ScriptRunInCloudEvent
```
cristianoventura, line 75: Is there any reason to have individual interfaces for each event type? Wondering if all these could be simplified with a single type/interface:

```ts
export type UsageEvent = {
  event: UsageEventName
}
```

src/services/usageTracking/types.ts
cristianoventura: I think these are good to start with, but there are some related to user settings that would be good to have down the road, such as:

- what browser people are using
- is browser recording/events enabled
- proxy mode
- and granular generator changes

src/services/usageTracking/types.ts
going-confetti: I agree it looks quite odd. The reason for this is that I plan to add optional event payloads, so it should make more sense then:

```ts
type RecordingCreatedEvent = {
  event. UsageEventName.RecordingCreated,
  payload: {
    numberOfRequests: number
  }
}
```

PR #782, opened 2025-08-15
fix: Modals close when trying to interact with the toolbar

### LINKED ISSUE #780: Picking tool in toolbar closes dialog/modal in the recorded page
### Describe the bug

If you open up a modal (e.g. [Radix UI dialog](https://www.radix-ui.com/themes/docs/components/dialog)) and try to select a tool from the recording toolbar, the modal will close.

This is because dialogs/modals have some kind of "click outside" or "focus lost" logic and when clicking in the toolbar this is triggered.

1. Open up [a modal](https://www.radix-ui.com/themes/docs/components/dialog)
2. Pick a tool in the toolbar

### Expected behavior

You should be able to pick elements, select text, etc. in the open modal.

### Actual behavior

The modal closes.

### Grafana k6 Studio version

k6 Studio 1.6.0

### OS

Any

allansson:
## Description

This PR fixes (or at least improves on) an issue where modals and dialogs would be closed whenever the user tried to pick a tool from the toolbar. This meant that it was impossible to inspect and add assertions to content inside modals.

## How to Test

I've tested it in three UI frameworks using their examples in their documentation: Radix UI, Material UI and React Aria.

1. Open a modal
2. Pick a tool from the toolbar
3. Interact with the dialog, e.g. pick an element
4. Try to add an assertion

NOTE: I've noticed an issue with Button-elements in React Aria where picking a Button opens our menu but then immediately closes it. It seems like they are doing something funky because it works fine in Radix UI and Material UI.

## Checklist

## Related PR(s)/Issue(s)

Resolves #780

### DIFF
--- a/extension/src/frontend/view/index.tsx
+++ b/extension/src/frontend/view/index.tsx
@@ -7,7 +7,7 @@ import { Theme } from '@/components/primitives/Theme'
 
 import { GlobalStyles } from './GlobalStyles'
 import { InBrowserControls } from './InBrowserControls'
-import { shouldBypassEvent } from './utils'
+import { isUsingTool } from './utils'
 
 let initialized = false
 
@@ -160,6 +160,10 @@ if (document.readyState === 'loading') {
   initialize()
 }
 
+function isInsideBrowserUI(element: Element) {
+  return element.getRootNode() === shadowRoot
+}
+
 // We want to make sure that the user can always interact with the toolbar.
 // This function checks if an event is being dispatched to an element inside
 // our UI and, if so, stops any event listeners outside the our shadow root
@@ -171,9 +175,7 @@ function bypassRecordedPage(event: Event) {
     return
   }
 
-  const root = target.getRootNode()
-
-  if (root !== shadowRoot && !shouldBypassEvent(event)) {
+  if (!isInsideBrowserUI(target)) {
     return
   }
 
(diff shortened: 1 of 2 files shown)


extension/src/frontend/view/index.tsx
```diff
@@ -33,32 +36,48 @@ function createMount() {
...
 function createShadowRoot(mount: Element) {
-  const shadow = mount.attachShadow({
+  shadowRoot = mount.attachShadow({
     mode: 'open',
   })
 
   const root = document.createElement('div')
 
   root.style.cursor = 'initial'
+  root.style.pointerEvents = 'auto'
```
allansson, line 76: This makes sure that the toolbar stays interactive even if `pointer-events` is set to `none` globally (unless they use `none !important` in which case they don't deserve to have their page recorded).

extension/src/frontend/view/utils.ts
```diff
@@ -1,6 +1,12 @@
 import { useInBrowserUIStore } from './store'
 import { Bounds } from './types'
 
+export function shouldBypassEvent(event: Event): boolean {
+  return (
+    event instanceof FocusEvent && useInBrowserUIStore.getState().tool !== null
+  )
+}
+
```
allansson: We want to bypass focus events whenever we have a tool picked. This (hopefully) prevents focus changes from closing modals.

Issue #787, opened 2025-08-19
Use underlying interactive element when recording clicks


allansson:
### Feature description

Recorded clicks are currently recorded on the precise element being clicked, so that if you have a `span` inside a `button` the click is registered on the selector for the `span`. From a user's perspective, it is likely that they expected the clicked element to be the button.

### Suggested solution (optional)

When recording a click, we should walk up the tree to try and find an interactive element (button or link) and use that for generating the selector. Only if no such element is found do we record the precise element being clicked.

PR #791, opened 2025-08-22
feat: Improve selector stability for click events

allansson:
## Description

Previous to this PR, the selector generated for click events would target the exact element that the user click, e.g. the SVG icon inside a button element. This most likely doesn't match the intent of the user. From their point of view they clicked the button and not an icon.

With these changes, we use any interactive element underneath the element that the user clicked which should lead to better selectors and more predictable results.

## How to Test

1. Start a recording.
2. Click an element inside a button or link (e.g. the icon in a button)

The selector generated for the click event should target the button and not the nested element.

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/787


extension/src/utils/dom.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Adapted list of widgets that are interacted with a simple click, regardless where the item
+ * was clicked. Other widgets, such as "scrollbar", can have their behaviour change depending on
+ * where the click happened.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles
+ */
+const SIMPLE_CLICK_WIDGET_ROLES = [
```
allansson, line 8: Not 100% sure if this is a good list, but I don't think it's unreasonable. It _feels_ like these are things that a user would to click.

cristianoventura: Another one that comes to mind is [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role).

allansson: Added!

going-confetti (approved):
LGTM!

PR #796, opened 2025-08-27
chore: Additional usage collection

going-confetti:
> [!IMPORTANT]  
> Despite it being a `chore`, this should be mentioned in the release notes.

> [!NOTE]  
> The usage collection docs should be updated to reflect the changes before we merge this PR.

## Description
- Add a new tracking event for when the entire script is copied
- Add a new `isLoggedIn` event payload property
- Calculate the number of rules of each type before saving the generator
- Internal: Remove the CodeEditor component

## How to Test
- Try copying the script and check the terminal output: the event should be tracked along with the `isLoggedIn` property in the payload
- Try saving a generator - the rule numbers should be logged in the terminal

## Checklist


src/components/Monaco/CodeEditor.tsx
going-confetti, line 1: Seems redundant: only used in 2 places and is basically a wrapper 🤷‍♂️

src/handlers/generator/index.ts
```diff
@@ -45,6 +45,22 @@ export function initialize() {
...
+              (rule) => rule.type === 'parameterization'
+            ).length,
+            verification: generator.rules.filter(
+              (rule) => rule.type === 'verification'
+            ).length,
+            customCode: generator.rules.filter(
+              (rule) => rule.type === 'customCode'
+            ).length,
+          },
+        },
```
going-confetti: Should I extract this into a util function to make it more readable?

src/services/usageTracking/index.ts
```diff
@@ -106,3 +117,13 @@ async function trackInstallation() {
     log.error('Error tracking installation:', error)
   }
 }
+
+async function getLoginState(): Promise<boolean> {
+  try {
+    const { profiles } = await getProfileData()
```
going-confetti, line 123: Looking for suggestions here: `getProfileData` reads data from disk, which is something I'd rather avoid as I wouldn't want the app to read from disk each time a usage event is send

src/views/Validator/ValidatorContent.tsx
```diff
@@ -106,7 +109,11 @@ export function ValidatorContent({
                     flex: 1;
                   `}
                 >
-                  <ReadOnlyEditor language="javascript" value={script} />
+                  <ReadOnlyEditor
+                    language="javascript"
+                    value={script}
+                    onCopy={handleCopy}
+                  />
```
going-confetti, line 116: Nit: `ReadOnlyEditor` should probably also be used in Script Preview, but it's out of scope of this PR.

src/handlers/generator/index.ts
going-confetti: Separate question: should we count disabled rules?

Llandy3d: possibly, but seems still small so it's understandable 🤔 

I would count disabled rules as well

src/services/usageTracking/index.ts
allansson: We could cache the value in the `src/handlers/auth/fs.ts`:

```ts
let profileData: Profile | null = null

export async function getProfileData(): Promise<Profile> {
  if (profile !== null) {
    return profile
  }

  try {
    const file = await readFile(filePath, 'utf-8')
   
    profile = ProfileSchema.parse(JSON.parse(file))

    return profile
  } catch {
    return {
      version: '1.0',
      tokens: {},
      profiles: {
        currentStack: '',
        stacks: {},
      },
    }
  }
}

export async function saveProfileData(profile: Profile) {
  try {
    await writeFile(filePath, JSON.stringify(profile, null, 2))
  } finally {
    // Always invalidate and re-read from disk because the disk is the 
    // single source of truth
    profile = null
  }
}
```

The other alternative is to cache it in frontend, but then you have to invalidate the data whenever there's a change to the login state (e.g. if we sign the user out because their token is rejected by an API).

src/services/usageTracking/index.ts
```diff
@@ -40,15 +41,25 @@ async function sendEvent(event: UsageEventWithMetadata) {
     return
   }
 
+  const payloadWithLoginState = {
+    isLoggedIn: await getLoginState(),
```
Llandy3d, line 45: should we be consistent with casing and have this called `logged_in` ? 🤔

src/handlers/generator/index.ts
allansson: Maybe just a `trackGeneratorUpdated(generator: Generator)` function at the top-level?

Are wa talking disabled per rule type or just a total?

src/services/usageTracking/index.ts
allansson: I think this is consistent with other metadata properties? It's the event types that use underscore.

src/handlers/generator/index.ts
Llandy3d: I interpreted the message as "count disabled rules in the total of each rules type", good question 🤔

going-confetti: Hmm, I meant something like
```ts
generator.rules.filter(
  (rule) => !rule.disabled & rule.type === 'correlation' 
).length
```

src/services/usageTracking/index.ts
going-confetti: Thank you!

Llandy3d (dismissed):

src/handlers/auth/fs.ts
```diff
@@ -11,11 +11,18 @@ const fileName =
 
 const filePath = path.join(app.getPath('userData'), fileName)
 
+let profileDataCache: Profile | null = null
```
allansson, line 14: You need to invalidate the cace in the `saveProfileData` function as well, or the values will go out of sync:

```ts
try {
  await writeFile(filePath, JSON.stringify(profile, null, 2))
} finally {
  // Force profile to be read from disk again
  profileDataCache = null
}
```

allansson (approved):
LGTM! 🚀

Issue #790, opened 2025-08-21
Clicking label of checkbox generates both click and checked events


allansson:
### Describe the bug

If you click on the label of a checkbox/radio control, the checked state of the control will change. k6 Studio will record these clicks as both a click and a check event.

### Steps to reproduce the problem

1. Start a recording
2. Click the label of a checkbox

### Expected behavior

A "check" event should be recorded.

### Actual behavior

A "click" followed by a "check" event is recorded.

### Grafana k6 Studio version

1.6.0

### OS

Any

PR #798, opened 2025-08-27
fix: Clicking the label of a checkbox/radio records two events

allansson:
## Description

This PR fixes some issues with clicking on labels:

1. If the label is associated with a checkbox/radio, both a click and check event would be recorded
2. If the label is associated with a button, two click events would be recorded (one for the label and one for the button)

This caused issues especially for checkboxes, because the duplicated events would toggle the checkbox on and off again, which is not what really happened in the browser.

## How to Test

1. Find a checkbox with a label (quickpizza has one, radix-ui has another)
2. Click the label
3. Check that a single click event was recorded.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6-studio/issues/790

### DIFF
--- a/extension/src/frontend/index.ts
+++ b/extension/src/frontend/index.ts
@@ -3,10 +3,16 @@ import './view'
 import { BrowserEvent } from '@/schemas/recording'
 
 import { generateSelector } from '../selectors'
-import { findInteractiveElement } from '../utils/dom'
+import {
+  findAssociatedElement,
+  findInteractiveElement,
+  isNativeButton,
+  isNativeCheckbox,
+  isNativeRadio,
+} from '../utils/dom'
 
+import { WindowEventManager } from './manager'
 import { client } from './routing'
-import { shouldSkipEvent } from './view/utils'
 
 function getButton(button: number) {
   switch (button) {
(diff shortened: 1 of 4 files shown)


extension/src/frontend/index.ts
```diff
@@ -169,39 +199,27 @@ function handleInputChange(target: HTMLInputElement) {
...
-window.addEventListener(
-  'change',
-  (ev) => {
-    if (shouldSkipEvent(ev)) {
-      return
-    }
-
-    if (ev.target instanceof HTMLTextAreaElement) {
-      handleTextAreaChange(ev.target)
+manager.capture('change', (ev) => {
```
allansson, line 202: Removing that third `{ capture: true }` argument caused a lot of reformatting here. In reality, only the `shouldSkipEvent` was removed because that's handled by the `WindowEventManager`.

extension/src/frontend/manager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type EventBlockerMap = Record<string, WeakSet<EventTarget>>
+
+/**
+ * Since we can't use `preventDefault`, `stopPropagation` or similar methods without
+ * affecting the actual page, we need a way to block events that are side-effects of
+ * other events. This manager adds logic on top of `addEventListener` that allows
+ * blocking the next event of a given type on a given target.
+ */
+export class WindowEventManager {
```
allansson, line 11: This is probably insane... but it works. I think...

extension/src/frontend/manager.ts
```diff
@@ -0,0 +1,73 @@
...
+    // don't do this, we might end up blocking an event we actually want to capture.
+    //
+    // For example, imagine a user clicking a label that is associated with a checkbox.
+    // This would trigger a follow-up `click` event on the checkbox which we don't want
+    // to record. However, if the recorded page calls `preventDefault` on the original
+    // click event, the follow-up click on the checkbox will never happen. So the block
+    // would never be removed and we would end up blocking the next legitimate click.
+    setTimeout(() => {
+      blockedForType.delete(target)
+    }, 1)
```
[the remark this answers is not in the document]
allansson: I asked co-pilot about this in vscode and it said that it was the best way of doing it. It also explained why using a microtask might cause errors. 🤣

Llandy3d (approved):
Seems to work well 🙌

e-fisher (approved):
LGTM

PR #808, opened 2025-09-02
refactor: Separate validator components for script and generator views

allansson:
## Description

I'm going to start implementing browser script validation. Right now the validator view is shared between the generator view and the validator view. But as things progress, these views are going to start to look and behave different. For example: 

- the validator view needs to be able to show details for both requests and browser events, but the generator don't want to show the browser events.
- the validator view will have a tab interface for switching between browser events and requests, whereas the generator view should only have the requests.

What this PR does is create two versions of the previously shared `ValidatorContent` component, one for the generator view and one for the validator view. The overall layout of the views have been lifted out into a shared `ValidatorLayout` component that controls all the allotment panes and some common things like the logs, checks and script panels.

## How to Test

Check that the validator view and the validator dialog in the generator view works like before.

## Checklist


src/components/WebLogView/HttpRequestDetails.tsx
```diff
@@ -9,12 +9,15 @@ import { ProxyData } from '@/types'
...
 import { ResponseDetails } from './ResponseDetails'
 
-interface DetailsProps {
+interface HttpRequestDetailsProps {
   selectedRequest: ProxyData | null
   onSelectRequest: (data: ProxyData | null) => void
 }
 
-export function Details({ selectedRequest, onSelectRequest }: DetailsProps) {
+export function HttpRequestDetails({
```
allansson, line 17: `Details` was very generic and would be confusing when you add details for browser events.

src/views/Generator/ValidatorResult.tsx
allansson, line 1: This is the Generator specific validator component.

src/main.ts
allansson, line 1: Had to fix this in order to commit.

src/views/Validator/ValidatorContent.tsx
allansson, line 1: The `ValidatorContent` is now specific only to the validator view.

src/components/Validator/ExecutionDetails.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const [selectedTab, setSelectedTab] = useState<'logs' | 'checks' | 'script'>(
+    'script'
+  )
+
+  const handleTabChange = (value: string) => {
+    if (value !== 'logs' && value !== 'checks' && value !== 'script') {
+      return
+    }
+
+    setSelectedTab(value)
```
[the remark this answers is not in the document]
allansson: Says to use a type assertion and then proceeds to cast the value twice instead. Classic Co-pilot.

e-fisher (approved):
LGTM

PR #809, opened 2025-09-02
refactor: Use K6Client as interface for running scripts

allansson:
## Description

The browser validator will use an HTTP server to capture actions, screenshots, etc. from the script. To make it easier to handle starting and stopping the server, I need a better interface to manage the lifecycle of a running test.

This PR refactors the `runScript` and `archiveScript` functions to use the same K6Client class that is used by the cloud functionality. The intention is to make this class a JS binding for the k6 CLI.

When using the `run` function, a `TestRun` is returned. The test run monitors the life cycle of the test run, emitting events for each step. It also handles reading and parsing stdout and stderr. I intend for this to also store artifacts sent from the script that we might not want to send directly to the renderer process, e.g. screenshots and DOM dumps.

The script handlers now interact with an instance of a TestRun instead of the raw ChildProcess from `spawn`.

## How to Test

Run scripts in the validator in a test generator and a script. Try starting and stopping tests. Everything should work like it did before.

## Checklist


src/main/script.ts
```diff
@@ -114,26 +71,40 @@ export const runScript = async ({
...
   })
 
-  return k6
+  testRun.on('log', ({ entry }) => {
+    browserWindow.webContents.send(ScriptHandler.Log, entry)
+  })
+
+  testRun.on('done', ({ result, checks }) => {
+    browserWindow.webContents.send(ScriptHandler.Check, checks)
+    browserWindow.webContents.send(ScriptHandler.Finished, result)
```
allansson, line 94: NOTE: I send the result here but it's not handled by the frontend in any way.

src/main/script.ts
```diff
@@ -143,62 +114,30 @@ const parseScript = (input: string) => {
...
-  })
+
+    return TEMP_K6_ARCHIVE_PATH
+  } catch (error) {
+    browserWindow.webContents.send(ScriptHandler.Failed)
+
+    if (error instanceof ArchiveError) {
+      for (const log of error.stderr) {
+        browserWindow.webContents.send(ScriptHandler.Log, log)
+      }
```
allansson, line 136: There's a slight difference in behavior here from the previous code: the previous code would send logs as they happened but this code will dump everything at the end of the process. I don't think that matters a whole lot since it's logging any errors that occurred while archiving the script.

src/main.ts
```diff
@@ -103,7 +103,7 @@ const createSplashWindow = async () => {
 const createWindow = async () => {
   const icon = getAppIcon(process.env.NODE_ENV === 'development')
   if (getPlatform() === 'mac') {
-    app.dock.setIcon(icon)
```
allansson, line 106: Needed to change this in order to commit.

src/utils/k6/testRun.ts
```diff
@@ -0,0 +1,204 @@
...
+  #handleStart = () => {
+    this.emit('start', {})
+  }
+
+  #handleError = (error: Error) => {
+    this.emit('error', { error })
+  }
+
+  #handleClose = (code: number | null) => {
+    switch (code) {
```
allansson, line 168: I've tried to map the exit codes that I thought made sense for our use case. Some things I've mapped because "why not?", others I've just ignored.

src/utils/k6/testRun.ts
```diff
@@ -0,0 +1,204 @@
...
+      case ExitCode.Success:
+        this.emit('done', {
+          result: {
+            passed: true,
+          },
+          checks: this.#checks,
+        })
+        break
+
+      case ExitCode.ScriptAborted:
```
allansson, line 178: This happens when you use `test.abort()` in a script. We're treating this as a test failed status because that's what e.g. `k6-testing` uses it for.

allansson (commented):
Self-review

src/utils/k6/client.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  constructor(code: number | null, stderr: K6Log[]) {
+    super('Failed to archive script')
+
+    this.code = code
+    this.stderr = stderr
+  }
+}
+
+export class K6Client {
```
allansson, line 77: I move the `K6Client` in a separate commit, but sadly GitHub (and I suppose Git) doesn't take that into account when doing diffs. The `archive` and `inspect` methods have only small modifications. The bulk of changes is the `run` method.

PR #811, opened 2025-09-04
internal: Setup server for tracking browser actions when validating scripts

allansson:
## Description

We need a way of tracking browser actions and their outcomes in order to display them when validating browser scripts. This PR sets up a shim for the `k6/browser` module that proxies select functions and reports their result back to an HTTP server hosted by k6 Studio.

## How to Test

1. Run exported browser script.
2. Run exported protocol script.
    - Both should work like normal
3. Uncomment the code beneath `// UNCOMMENT ME FOR TESTING!`
4. Run a browser script

In the console, you should see some actions being reported.

## Checklist


extension/src/utils/events.ts
```diff
@@ -39,14 +40,38 @@ export class EventEmitter<Events extends EventMap<Events>> {
...
-    if (listeners) {
-      for (const listener of listeners) {
-        try {
-          listener(event)
-        } catch (error) {
-          console.log('Error in event listener', error)
+    // The first to emit processes the queue. Emits while processing will be
+    // deferred until after all listeners of the current event have been called.
+    if (this.#queue !== null) {
+      this.#queue.push(this.#defer(type, event))
```
allansson, line 46: We can preserve type safety by queuing functions instead of `[type, event]`.

extension/src/utils/events.ts
```diff
@@ -39,14 +40,38 @@ export class EventEmitter<Events extends EventMap<Events>> {
...
   emit<Type extends keyof Events>(type: Type, event: Events[Type]) {
-    const listeners = this.#listeners[type]
-
-    if (listeners) {
-      for (const listener of listeners) {
-        try {
-          listener(event)
-        } catch (error) {
-          console.log('Error in event listener', error)
+    // The first to emit processes the queue. Emits while processing will be
```
allansson, line 43: I realized this issue when I added the `stop` event to `TestRun`. You would expect that events would be processed in batches so that all listeners of `done` would received the event before any listener received `stop`.

src/main/runner/shims/browser/index.ts
```diff
@@ -0,0 +1,39 @@
+import { Locator, browser as nativeBrowser } from 'k6/browser'
+
+import { createProxy, ProxiedMethods } from './utils'
+
+const locatorProxy = (
+  _target: Locator,
+  selector: string
+): ProxiedMethods<Locator> => {
+  return {
+    click: {
```
allansson, line 10: I've only instrumented two functions for testing purposes the tracking. The rest of the actions will be added once I get to implementing an actual UI.

src/main/runner/instrumentation.ts
```diff
@@ -0,0 +1,182 @@
+import { parse, TSESTree as ts } from '@typescript-eslint/typescript-estree'
```
allansson, line 1: I moved all the script modifications out into its own file to reduce the clutter in `scripts.ts`.

src/main/script.ts
```diff
@@ -58,8 +44,8 @@ export const runScript = async ({
   proxySettings,
   browserWindow,
 }: RunScriptOptions) => {
-  // 1. Get an enhanced version of the script content
-  const modifiedScript = await enhanceScriptFromPath(scriptPath)
+  // 1. Get an instrumented version of the script content
+  const modifiedScript = await instrumentScriptFromPath(scriptPath)
```
allansson, line 48: `enhanceScript` sounded like we somehow improved the script (if anything we were doing the opposite). I think this new name is more appropriate.

src/main/script.ts
```diff
@@ -80,6 +66,18 @@ export const runScript = async ({
     prefix: '',
   })
 
+  const trackingServer = await createTrackingServer().catch(() => null)
```
allansson, line 69: The `catch` here is temporary so that we don't have to worry about this destroying anything if it was merged today and released tomorrowy

src/main/script.ts
```diff
@@ -115,17 +114,14 @@ export const runScript = async ({
...
-    range: true,
+  testRun.on('stop', () => {
+    trackingServer?.dispose()
   })
+
+  return testRun
 }
 
-export const archiveScript = async (
+const archiveScript = async (
```
allansson, line 124: This wasn't being used anywhere else so I just cleaned up the export.

.eslintrc.cjs
```diff
@@ -72,6 +72,7 @@ module.exports = {
         alphabetize: { order: 'asc' },
       },
     ],
+    'import/no-unresolved': ['error', { ignore: ['^k6'] }],
```
allansson, line 75: This rule failed when importing `k6/*` because there isn't any package in `node_modules`.

.gitignore
```diff
@@ -107,3 +107,6 @@ resources/win/x86_64/k6
 
 # Sentry Config File
 .env.sentry-build-plugin
+
+# Shims
+resources/shims
```
allansson, line 112: Since the browser shim is built from source, we don't want it to be committed.

src/main/runner/instrumentation.ts
```diff
@@ -0,0 +1,182 @@
...
+        case 'k6/http':
+          httpImport = node
+          break
+
+        case 'k6/browser':
+          browserImport = node
+
+          // Replace the import source with our shim path.
+          browserImport.source.value = browserShim
+          browserImport.source.raw = JSON.stringify(browserShim)
```
allansson, line 71: We should consider doing something similar for the other `http`-module shim. It reduces the complexity considerably.

PR #824, opened 2025-09-15
fix: Duplicate call to click when recording a submit button click

allansson:
## Description

When clicking the submit button of a form, the generated code will contain two clicks on the submit button. The reason was that we were recording two events: the initial `click` on the button and the `submit` on the form.

Form submits are emitted in the code as a `.click()` because there is no `.submit()` (or even `locator.evaluate(el => el.requestSubmit()`).

This PR ignores the `submit` event if a click on the submitter occurred. 

## How to Test

1. Start a recording.
2. Go to a form with a submit button.
3. Fill in the form.
4. Click the submit button
5. Stop the recording
6. Export a script

The exported script should only contain one call to `.click()` on the submit button.

## Checklist

### DIFF
--- a/extension/src/frontend/index.ts
+++ b/extension/src/frontend/index.ts
@@ -231,7 +231,7 @@ manager.capture('submit', (ev) => {
     return
   }
 
-  // We might have already recorded a click on the submit button. In the futre we might
+  // We might have already recorded a click on the submit button. In the future we might
   // want to replace the previous click with the form submission, but for now we will
   // just skip recording the submission.
   const hasClickedSubmit = manager.history.some(


extension/src/frontend/manager.ts
```diff
@@ -11,6 +12,16 @@ type EventBlockerMap = Record<string, WeakSet<EventTarget>>
 export class WindowEventManager {
   #blockedEvents: EventBlockerMap = {}
 
+  // The browser will execute sequences of events synchronously in a single event loop
+  // tick, so we can keep a history of the events that have happened before and check
+  // that if we need to selectively ignore some of them.
```
allansson, line 17: Co-pilot swears that this is how things work.

PR #825, opened 2025-09-15
feat: Record user granting permission to browser feature

allansson:
## Description

Some features in a browser, such as the microphone or geolocation, require the user to explicitly grant access to them. This PR adds support for recording when user grants permissions and for generating the appropriate code for granting them in exported scripts.

## How to Test

1. Start recording at https://permission.site
2. Grant permission to some features (not all seem to be working, but after debugging it seems like it has something to do with the browser or site and not the implementation).
3. Stop the recording
4. Export a script
5. Run the script

The generated code should call `grantPermissions` with the appropriate permissions. It should constrain the origin to the recorded page. Running the script shouldn't cause any errors.

## Checklist


src/handlers/browser/launch.ts
```diff
@@ -95,6 +95,8 @@ const FEATURES_TO_DISABLE = [
 ]
 
 const BROWSER_RECORDING_ARGS = [
+  `--use-fake-device-for-media-stream`,
+  `--use-fake-ui-for-media-stream`,
```
allansson, line 99: This fakes the microphone for the browser. I added it because in my testing, Chrome just said it needed system permissions to use it even though system permissions were already set.

src/main/script.ts
```diff
@@ -66,6 +66,12 @@ export const runScript = async ({
     prefix: '',
   })
 
+  const browserArgs = [
+    ...proxyArgs,
+    `use-fake-device-for-media-stream`,
+    `use-fake-ui-for-media-stream`,
```
allansson, line 72: Again, faking the microphone.

extension/src/frontend/index.ts
```diff
@@ -240,3 +240,74 @@ manager.capture('submit', (ev) => {
...
+  | 'gyroscope'
+  | 'magnetometer'
+  | 'accessibility-events'
+  | 'clipboard-read'
+  | 'clipboard-write'
+  | 'payment-handler'
+
+type PermissionMap = { [Key in SupportedPermissionName]: Key }
+
+const permissions: PermissionMap = {
```
allansson, line 266: You might think that I could have used an array here, and you'd be right, but! Doing it this way means that if you add another permission to `SupportedPermissionName` you also have to add it to the map, thus making sure that you always iterate over all supported permissions.

allansson (commented):
Self-review

e-fisher:
Closing as inactive

Issue #834, opened 2025-09-16
Add support for emitting `getByTestId` locators


allansson:
### Feature description

Since k6 v1.2.0, the browser module has support for `getByTestId` locators. These locators are a lot more better than CSS selectors and lets engineers instrument their DOM to make QA testing more reliable.

### Suggested solution (optional)

We have the `selectors` object on events already but the only type of selector currently supported is `css`. We should add an optional `testId` property to the selectors and set it to the value of the `data-testid` attribute on the target element, if available.

Once we have the id, we can make a decision during codegen on which selector to use.

PR #849, opened 2025-09-24
feat(Browser): Add getByTestId support

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/834

## How to Test
- Find a website that uses `data-testid` on its elements
- Start a recording and interact with these elements
- Verify that the testId selectors are displayed in the event log
- Verify that, when exported, the browser script uses `getByTestId`

## Checklist

### DIFF
--- a/src/codegen/browser/code/options.ts
+++ b/src/codegen/browser/code/options.ts
@@ -31,7 +31,7 @@ function isBrowserScenario(scenario: ir.Scenario) {
       case 'NewPageExpression':
       case 'GotoExpression':
       case 'ReloadExpression':
-      case 'NewCSSLocatorExpression':
+      case 'NewCssLocatorExpression':
       case 'NewTestIdLocatorExpression':
       case 'ClickExpression':
       case 'ClickOptionsExpression':
--- a/src/codegen/browser/code/options.ts
+++ b/src/codegen/browser/code/options.ts
@@ -31,7 +31,8 @@ function isBrowserScenario(scenario: ir.Scenario) {
       case 'NewPageExpression':
       case 'GotoExpression':
       case 'ReloadExpression':
-      case 'NewLocatorExpression':
+      case 'NewCSSLocatorExpression':
+      case 'NewTestIdLocatorExpression':
       case 'ClickExpression':
       case 'ClickOptionsExpression':
       case 'FillTextExpression':
--- a/src/codegen/browser/test.ts
+++ b/src/codegen/browser/test.ts
@@ -86,14 +86,16 @@ function buildBrowserNodeGraph(events: BrowserEvent[]) {
     // await input.focus()
     // await input.type("Hello")
     // await input.press("Enter")
+
+    // TODO: is it enough to compare just the CSS selector?
     if (
-      previousLocator?.selector !== selector.css ||
+      previousLocator?.selector.css !== selector.css ||
       previousLocator?.inputs.page.nodeId !== page.nodeId
     ) {
       previousLocator = {
         type: 'locator',
         nodeId: crypto.randomUUID(),
-        selector: selector.css,
+        selector,
         inputs: {
           page,
         },
--- a/src/codegen/browser/test.ts
+++ b/src/codegen/browser/test.ts
@@ -87,7 +87,6 @@ function buildBrowserNodeGraph(events: BrowserEvent[]) {
     // await input.type("Hello")
     // await input.press("Enter")
 
-    // TODO: is it enough to compare just the CSS selector?
     if (
       previousLocator?.selector.css !== selector.css ||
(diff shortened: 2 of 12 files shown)


src/codegen/browser/test.ts
```diff
@@ -86,14 +86,16 @@ function buildBrowserNodeGraph(events: BrowserEvent[]) {
     // await input.focus()
     // await input.type("Hello")
     // await input.press("Enter")
+
+    // TODO: is it enough to compare just the CSS selector?
```
going-confetti: 💭

src/components/BrowserEventList/EventDescription/Selector.tsx
```diff
@@ -44,7 +44,7 @@ export function Selector({ selector, onHighlight }: SelectorProps) {
       onMouseEnter={handleMouseEnter}
       onMouseLeave={handleMouseLeave}
     >
-      {selector.css}
+      {selector.testId ?? selector.css}
```
going-confetti: Not sure about the presentation. What do you think about something like this:

Image

src/codegen/browser/test.ts
allansson: I think it's good enough for now. We could start assigning unique ids to elements while recording, but that doesn't have to be part of this PR.

src/components/BrowserEventList/EventDescription/Selector.tsx
allansson: I think it's a good idea. The icon in the first one (`</>`) melts together with the text though. Maybe a separator between the icon and the selector value? Ascii art:

```
______________________________
| </> | .flex > .text-center |
------------------------------
```

I think we keep it simple for now and then we can expand on it as we add additional selectors.

_<sup>(We also have the hover highlighting to consider, but let's pretend that thing doesn't exist and continue using the CSS selector for now.)</sup>_

src/codegen/browser/code/options.ts
```diff
@@ -31,7 +31,8 @@ function isBrowserScenario(scenario: ir.Scenario) {
       case 'NewPageExpression':
       case 'GotoExpression':
       case 'ReloadExpression':
-      case 'NewLocatorExpression':
+      case 'NewCSSLocatorExpression':
```
allansson: This is pure opinion, but I usually refer to Google's [guidelines for camel-casing](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) (though this is technically pascal case). According to the guidelines, the name should be `NewCssLocatorExpression` which to me is easier to read.

src/components/BrowserEventList/BrowserEventList.tsx
```diff
@@ -39,6 +39,9 @@ export function BrowserEventList({
                   <EventIcon event={event} />
                   <div
                     css={css`
+                      display: flex;
+                      align-items: center;
+                      gap: 0.25rem;
```
going-confetti, line 44: Needed this to make sure everything fits now that `Selector` uses the `Badge` component:

Image

We'll likely need to add a tooltip in future.

src/components/BrowserEventList/EventDescription/Selector.tsx
going-confetti: This is the latest iteration:
image

src/components/BrowserEventList/EventDescription/Selector.tsx
```diff
@@ -28,23 +30,25 @@ export function Selector({ selector, onHighlight }: SelectorProps) {
   }
 
   return (
-    <strong
+    <Badge
```
allansson: Fun fact: you need to create a custom component for this since `@radix-ui/themes` doesn't work in the browser extension. 😞

Image

src/components/BrowserEventList/EventDescription/Selector.tsx
```diff
@@ -28,23 +30,25 @@ export function Selector({ selector, onHighlight }: SelectorProps) {
...
+          background-color: ${isRecording ? 'var(--gray-4)' : undefined};
+        }
       `}
       onMouseEnter={handleMouseEnter}
       onMouseLeave={handleMouseLeave}
     >
-      {selector.css}
-    </strong>
+      {selector.testId ? <TestTubeDiagonalIcon /> : <BracesIcon />}
+      <Code variant="ghost" truncate>
```
allansson: Same here...

src/components/BrowserEventList/EventDescription/Selector.tsx
going-confetti: Fixed this, but did not extract the badge into a separate component under `/primitives` (mainly because I didn't know it to be configurable yet) - let me know if you have a strong opinion on this

image

PR #853, opened 2025-10-02
feat: Add Regex support to Verification rule

going-confetti:
## Checklist

### DIFF
--- a/src/rules/verification.utils.ts
+++ b/src/rules/verification.utils.ts
@@ -12,6 +12,8 @@ export function getValueFromRule(rule: VerificationRule, response: Response) {
       return getRecordedValue(rule.target, response)
     case 'string':
       return `'${rule.value.value}'`
+    case 'regex':
+      return `new RegExp(${JSON.stringify(rule.value.regex)})`
     case 'variable':
       return `VARS['${rule.value.variableName}']`
     case 'number':
@@ -37,6 +39,8 @@ export function getCheckExpression(
       return `${target}.includes(${value})`
     case 'notContains':
       return `!${target}.includes(${value})`
+    case 'matches':
+      return `${value}.test(${target})`
     default:
       return exhaustive(operator)
   }
@@ -58,6 +62,8 @@ export function getCheckDescription(
       return `${rule.target} contains ${valueDescription}`
     case 'notContains':
       return `${rule.target} does not contain ${valueDescription}`
+    case 'matches':
+      return `${rule.target} matches ${valueDescription}`
     default:
       return exhaustive(operator)
   }
@@ -109,6 +115,8 @@ function getValueDescription(rule: VerificationRule, value: string | number) {
       return `variable "${rule.value.variableName}"`
     case 'number':
       return rule.value.number
+    case 'regex':
+      return `/${rule.value.regex}/`
     default:
       return exhaustive(rule.value)
   }
--- a/src/schemas/generator/v2/rules.ts
+++ b/src/schemas/generator/v2/rules.ts
@@ -33,6 +33,21 @@ export const NumberValueSchema = z.object({
     .nonnegative({ message: 'Must be a positive number' }),
 })
 
+export const RegexValueSchema = z.object({
+  type: z.literal('regex'),
+  regex: z.string().refine(
+    (value) => {
+      try {
+        new RegExp(value)
+        return true
+      } catch {
+        return false
+      }
+    },
+    { message: 'Invalid regular expression' }
+  ),
+})
+
 export const FilterSchema = z.object({
   path: z.string(),
 })
@@ -142,6 +157,8 @@ const VerificationOperator = z.enum([
   'notEquals',
   'contains',
   'notContains',
+  // Regex only
+  'matches',
 ])
 
 const VerificationTarget = z.enum(['body', 'status'])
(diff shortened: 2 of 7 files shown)


src/rules/shared.ts
```diff
@@ -87,7 +87,8 @@ export const matchRegex = (value: string, regexString: string) => {
   const regex = new RegExp(regexString)
   const match = value.match(regex)
   if (match) {
-    return match[1]
```
going-confetti, line 90: This caused a bug when the regex selector didn't have capture groups, leading to `match[1]` always being `undefined`.

src/rules/shared.ts
```diff
@@ -87,7 +87,8 @@ export const matchRegex = (value: string, regexString: string) => {
...
+    // If no capture groups, return the full match; otherwise return the first capture group
+    return match.length === 1 ? match[0] : match[1]
```
going-confetti: Image

src/schemas/generator/v2/rules.ts
```diff
@@ -33,6 +33,21 @@ export const NumberValueSchema = z.object({
...
+  regex: z.string().refine(
+    (value) => {
+      try {
+        new RegExp(value)
+        return true
+      } catch {
+        return false
+      }
+    },
+    { message: 'Invalid regular expression' }
```
going-confetti, line 47: This is the only validation so far

src/views/Generator/RuleEditor/VerificationEditor/ValueEditor.utils.ts
```diff
@@ -11,21 +11,37 @@ export const VALUE_LABELS: Record<VerificationRule['value']['type'], string> = {
...
 
 export const OPERATOR_LABELS = {
   equals: getLogicalOperatorLabelAndIcon('equals'),
   notEquals: getLogicalOperatorLabelAndIcon('notEquals'),
   contains: getLogicalOperatorLabelAndIcon('contains'),
   notContains: getLogicalOperatorLabelAndIcon('notContains'),
+  matches: {
+    label: 'Matches',
+    icon: '.*',
+  } as const,
```
going-confetti, line 25: I could've extended `getLogicalOperatorLabelAndIcon`, but it's not really a logical operator, so I decided not to.

src/views/Generator/TestRuleContainer/TestRule/TestRuleSelector.tsx
```diff
@@ -106,11 +106,7 @@ function SelectorLabel({ selector }: { selector: ReplacerSelector }) {
         </>
       )
     case 'regex':
-      return (
-        <>
-          <Strong>(.*) {stringFallback(selector.regex)}</Strong>
-        </>
-      )
+      return <Strong>{new RegExp(selector.regex).toString()}</Strong>
```
going-confetti, line 109: Using the same way of displaying the expression in both the regex selector and the verification rule content.

src/schemas/generator/v2/rules.ts
```diff
@@ -142,6 +157,8 @@ const VerificationOperator = z.enum([
   'notEquals',
   'contains',
   'notContains',
+  // Regex only
+  'matches',
```
going-confetti, line 161: I believe the current schema is flawed, because certain constraints are only enforced in the UI, but not in the schema:
for example, it's technically valid to have a rule that checks whether the response body doesn't contain the recorded value, but it won't be reflected in the UI.

I decided not to address this in this PR and create a ticket for future instead. But first I'd like to hear your opinion on this.

src/views/Generator/RuleEditor/VerificationEditor/VerificationEditor.tsx
```diff
@@ -24,7 +24,7 @@ export function VerificationEditor() {
       <Text size="2" as="p" mb="2" color="gray">
         Add verification checks to validate response data.
       </Text>
-      <Grid columns="2" gap="3">
+      <Grid columns="1fr auto 1fr" gap="4">
```
going-confetti, line 27: Image

src/rules/verification.utils.ts
```diff
@@ -109,6 +117,8 @@ function getValueDescription(rule: VerificationRule, value: string | number) {
       return `variable "${rule.value.variableName}"`
     case 'number':
       return rule.value.number
+    case 'regex':
+      return new RegExp(escapeRegExp(rule.value.regex)).toString()
```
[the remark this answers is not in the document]
going-confetti: The usage is correct, because this piece of code is used to generate a text description of the value. I may need to add a comment I guess.
image

image
image

src/rules/shared.ts
e-fisher: While this change makes perfect sense for parameterization and verification rules, I think it might cause some confusion in correlation, take this example:

screencapture 2025-10-03 at 14 54 36@2x

Previously this expression wouldn't match since there are no capture groups, but now it takes match[0] and replaces "csrf" cookie key with correlation variable, which resolves to the same "csrf" string. While not breaking anything, it makes it looks like some value is correlated.

I see 2 possible ways to address this:
1. Keep previous logic for the correlation rule and only add fallback to match[0] for parameterization and verification rules
2. Add validation rule in zod to require capture groups for regular expressions in correlation rules. I think we've talked about this at some point, but can't find anything in code. 

Currently, we don't seem to communicate about capture groups being required in correlation, the validation message would address it.

e-fisher (commented):
The input focus outline looks a bit out of place, not sure when this happened, but it affects other inputs with slots as well. Should we try to make it look the same way as in radix docs?

screencapture 2025-10-03 at 14 14 38@2x

screencapture 2025-10-03 at 14 14 52@2x

Also, the input is not full width in the verification rule form, while it is in correlation and parameterization

screencapture 2025-10-03 at 15 14 03@2x

src/rules/shared.ts
going-confetti: I like your second suggestion with a small addition: we should also try to make it clear before the user inputs a value, not just show them a validation error afterwards. I'll bring the topic of capture groups in the next meeting

going-confetti: Following our recent discussion, I reverted this change for now

PR #860, opened 2025-10-10
internal: Add ability to manage AI API key

e-fisher:
## Description

This is the first PR from Auto Correlation series - splitting to make it easier to review. 
Add new settings tab "AI" for managing API key. 

- User's can add their OpenAI API key through settings
- Key is encrypted using electron's safeStorage method and stored in app settings
- Once the key is saved, it's no longer possible to edit it because of encryption, but it can be removed by clicking "remove" button

## How to Test

- Enable feature toggle
- Open Settings -> AI
- Add any string and save settings
- **You should see prompt to allow keychain access**
- Open settings again
- **You should see "API key configured" message and input should be disabled**
- Click "change or remove key" 
- **You should be able to save settings without key or add a new one**

There's no logic to use the decrypted key in this PR, this will be added in a follow-up PR.

screencapture 2025-10-10 at 15 20 14@2x

screencapture 2025-10-10 at 15 20 08@2x

## Checklist

### DIFF
--- a/src/main/settings.ts
+++ b/src/main/settings.ts
@@ -209,6 +209,7 @@ export async function applySettings(
 
   if (modifiedSettings.ai) {
     k6StudioState.appSettings.ai = modifiedSettings.ai
+    // TODO: re-initialize AI client with new settings
   }
 }
 
--- a/src/handlers/settings/index.ts
+++ b/src/handlers/settings/index.ts
@@ -29,11 +29,6 @@ export function initialize() {
       const { windowState: _, ...settings } = data
       const modifiedSettings = await saveSettings(settings)
 
-      if (modifiedSettings.ai && 'apiKey' in modifiedSettings.ai) {
-        console.info('OpenAI API key changed, resetting AI client')
-        // TODO: reset AI client
-      }
-
       await applySettings(modifiedSettings, browserWindow)
 
       sendToast(browserWindow.webContents, {


src/main/settings.ts
```diff
@@ -12,6 +12,7 @@ import { getPlatform } from '../utils/electron'
 import { safeJsonParse } from '../utils/json'
 import { getExecutableNameFromPlist } from '../utils/plist'
 
+import { encrypString, decryptString } from './encryption'
```
going-confetti:
```suggestion
import { encryptString, decryptString } from './encryption'
```

src/main/settings.ts
```diff
@@ -199,4 +206,34 @@ export async function applySettings(
...
+  apiKey: string | undefined
+): string | undefined {
+  if (apiKey === undefined || apiKey.trim() === '') {
+    return undefined
+  }
+
+  try {
+    return encrypString(apiKey)
+  } catch (error) {
+    // TODO: show UI error when encryption is not available
```
going-confetti: I would also check if `safeStorage.isEncryptionAvailable()` returns `false` (and maybe if `safeStorage.getSelectedStorageBackend()` returns `basic_text`) and disable the tab/API key form.

src/components/Settings/AiSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+    })
+  }
+
+  return (
+    <SettingsSection>
+      <FieldGroup
+        name="ai.apiKey"
+        label="OpenAI API Key"
+        errors={errors}
+        hint="Your API key will be encrypted using your operating system's secure storage and used to generate AI-powered test scripts."
```
going-confetti: I think the hint is good for when the users opens this tab for the first time, but can be made clearer:
* `will be` -> `is`
* Test scripts themselves are not AI-powered. I'd suggest something along the lines of "used for AI-powered test script generation" 🤔

src/components/Settings/AiSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+                      </Text>
+                    </Flex>
+                    <Button
+                      variant="outline"
+                      color="red"
+                      size="2"
+                      onClick={handleClearApiKey}
+                    >
+                      <TrashIcon size={16} />
+                      Change or remove API key
```
going-confetti: I find this button confusing, because it's not clear what it actually does: is it removing or changing the key? Should it simply be "Remove"/"Reset"?

src/components/Settings/AiSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  </TextField.Root>
+                </Flex>
+
+                {isApiKeyConfigured && (
+                  <Flex gap="2" justify="between" align="center">
+                    <Flex align="center" gap="2">
+                      <CheckCircleIcon color="green" size={16} />
+                      <Text size="2" color="green">
+                        API key is configured and encrypted
+                      </Text>
```
going-confetti: IMO this is redundant: the "encrypted" part is mentioned in the field description, the "configured" part is implied by the field having value. I would suggest adding some kind of a health check/test integration functionality, that would send a request to `https://api.openai.com/v1/models` and surface any 4** to the user.

src/components/Settings/AiSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+          />
+
+          <Box>
+            <Text size="1">
+              <strong>Need an API key?</strong> Create one at{' '}
+              <ExternalLink href="https://platform.openai.com/api-keys">
+                platform.openai.com/api-keys
+              </ExternalLink>
+              .
+            </Text>
```
going-confetti: This is a good zero-state, but I'm not sure it's needed once the key has been added.
If you have a strong opinion that it should be here, could you please add `color="gray"` to `Text`? The default text color is darker (black?) than what's used in the field hint and it's pretty noticeable.

src/components/Settings/AiSettings.tsx
going-confetti: I would even go further and replace it with an icon button, because that's the pattern we have both in GCk6 and in k6 Studio:
Image

src/main/settings.ts
e-fisher: Added a callout when encryption is not available

screencapture 2025-10-14 at 10 06 57

src/components/Settings/AiSettings.tsx
e-fisher: Agree, replaced with icon button ✅

src/components/Settings/AiSettings.tsx
e-fisher: Thought about this too, but just removing "api key configured" message for now and will keep a note to consider adding ping in polishing phase

PR #861, opened 2025-10-10
chore: Browser events schema v2

going-confetti:
> [!WARNING]  
> Please test this thoroughly with any "unusual" HAR recordings you may have (e.g. created in the old browser recorder extension or exported from a browser)

### Description
- Some future-proofing to allow us to introduce changes in how browser events are stored in the recording.
- Use a zod schema to parse HAR files

### Testing instructions
Note that some old recordings created while testing earlier iterations of certain browser recording feature may not work, because they don't follow our Recording schema.

Verify that the following works as expected:
1. The app builds and starts without any errors
2. Previously created recordings can be opened and used in Generator
3. Previously created recordings can be used to create browser test scripts
4. You should be able to create a new recording and use it to generate a browser test script

Note that now a zod schema is used to parse HAR files

### DIFF
--- a/src/schemas/recording/har/index.ts
+++ b/src/schemas/recording/har/index.ts
@@ -1,91 +1,6 @@
 import { z } from 'zod'
 
-// Based on the historical knowledge collected in https://github.com/grafana/har-to-k6/blob/master/li-har.spec.md
-
-export const HarHeaderSchema = z.object({
-  name: z.string(),
-  value: z.string(),
-})
-
-const HarCookieSchema = z.object({
-  name: z.string(),
-  value: z.string(),
-})
-
-const HarQueryParamSchema = z.object({
-  name: z.string(),
-  value: z.string(),
-})
-
-const HarPostDataSchema = z.object({
-  mimeType: z.string(),
-  text: z.string().optional(),
-  params: z
-    .array(
-      z.object({
-        name: z.string(),
-        value: z.string().optional(),
-        fileName: z.string().optional(),
-        contentType: z.string().optional(),
-      })
-    )
-    .optional(),
-})
-
-export const HarRequestSchema = z
-  .object({
-    method: z.string(),
-    url: z.string(),
-    httpVersion: z.string().optional(),
-    headers: z.array(HarHeaderSchema).optional(),
-    queryString: z.array(HarQueryParamSchema).optional(),
-    postData: HarPostDataSchema.optional(),
-    cookies: z.array(HarCookieSchema).optional(),
-    headersSize: z.number().optional(),
-    bodySize: z.number().optional(),
-  })
-  .passthrough()
-
-export const HarContentSchema = z.object({
-  size: z.number().optional(),
-  mimeType: z.string().optional(),
-  text: z.string().nullable().optional(),
-  encoding: z.string().optional(),
-})
-
-export const HarResponseSchema = z
-  .object({
-    status: z.number(),
-    statusText: z.string(),
-    httpVersion: z.string(),
-    headers: z.array(HarHeaderSchema),
-    content: HarContentSchema,
-    cookies: z.array(HarCookieSchema),
-    redirectURL: z.string(),
-    headersSize: z.number(),
-    bodySize: z.number(),
-  })
-  .passthrough()
-
-export const HarEntrySchema = z.object({
-  pageref: z.string().optional(),
-  comment: z.string().optional(),
-  startedDateTime: z.string().datetime({ offset: true }).optional(),
-  time: z.number().min(0).optional(),
-  timings: z.object({}).passthrough().optional(),
-  request: HarRequestSchema,
-  response: HarResponseSchema.optional(),
-  cache: z.object({}).passthrough().optional(),
-})
-
-export const HarPageSchema = z
-  .object({
-    id: z.string(),
-    title: z.string(),
-    startedDateTime: z.string().datetime({ offset: true }).optional(),
-    pageTimings: z.object({}).passthrough().optional(),
-  })
-  .passthrough()
+import { HarEntry, HarPage } from '@/types/recording'
 
 export const LogSchema = z.object({
   version: z.string(),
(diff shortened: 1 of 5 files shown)


src/schemas/recording/index.ts
```diff
@@ -1,2 +1,45 @@
...
+import { exhaustive } from '../../utils/typescript'
+
+import * as RecordingV1 from './v1/recording'
+import * as RecordingV2 from './v2/recording'
+
+export * from './v2/browser'
+
+type AnyRecording = RecordingV1.Recording | RecordingV2.Recording
+
+const AnyRecordingSchema = z.unknown().transform((data): AnyRecording => {
```
allansson: What if we move the migrated browser events schema inside the `AnyRecordingSchema`?

```ts

const BrowserEventsSchemaV1 = BrowserEventSchema.array()

const BrowserEventsSchemaV2 = z.object({
  version: z.literal(1),
  events: BrowserEventSchema.array()
})

const AnyBrowserEventsSchema = z.union([
  BrowserEventsSchemaV1,
  // Future proofing
  z.discriminatedUnion("version", [BrowserEventsSchemaV2])
])

// Schema for any version stored on disk
type AnyBrowserEvents = z.infer<typeof AnyBrowserEventsSchema>

function migrate(browserEvents: AnyBrowserEvents) {
  if ("version" in browserEvents === false) {
    // v1.migrate should return a v2. Apply recursively as always. 😛 
    return migrate(v1.migrate(browserEvents))
  }

  return browserEvents
}

// The canonical schema for k6 Studio
const BrowserEventSchema = BrowserEventsSchema.transform(migrate)


const AnyRecordingSchema = z.object({
  log: {
    entries: z.unknown().transform(value => value as Entry[])
    // etc.
    
   // This makes _browserEvents versioned individually from the HAR file.
    _browserEvents: BrowserEventsSchema.optional()
  }
})

```

src/schemas/recording/v2/recording.ts
```diff
@@ -0,0 +1,28 @@
...
+    _browserEvents: z
+      .object({
+        version: z.literal('2'),
+        events: BrowserEventSchema.array(),
+      })
+      .optional(),
+  }),
+})
+
+export type Recording = z.infer<typeof RecordingSchema>
```
allansson: Love the name change here. I've wanted to do that since forever ❤️

src/schemas/recording/index.ts
```diff
@@ -1,2 +1,47 @@
-export * from './v1/browser'
-export { Recording } from './v1/recording'
+import type { Page } from 'har-format'
+import { z } from 'zod'
+
+import { EntryWithOptionalResponse } from '@/types/har'
+
+import * as v1 from './browser/v1'
+import * as v2 from './browser/v2'
```
going-confetti, line 4: @allansson I think you suggestion makes a lot of sense, because we're much less likely to change the schema of the HAR log. I added the changes and went a few steps further: created the `browser` sub-folder with folders for each version and move the browser event schemas there - wdyt?

src/schemas/recording/index.ts
going-confetti: Responded in a [separate thread](https://github.com/grafana/k6-studio/pull/861#discussion_r2432363603)

src/codegen/browser/test.ts
```diff
@@ -137,7 +137,7 @@ function buildBrowserNodeGraph(events: BrowserEvent[]) {
           modifiers: event.modifiers,
           inputs: {
             previous,
-            locator: getLocator(event.tab, event.selector),
+            locator: getLocator(event.tab, event.target.selectors),
```
going-confetti, line 140: `LocatorNode` also needs a similar change, but I would rather do it in a separate PR

src/schemas/recording/index.ts
```diff
@@ -1,2 +1,52 @@
...
+}
+
+export const RecordingSchema = z.object({
+  // HAR log format version 1.2
+  log: z.object({
+    version: z.literal('1.2'),
+    creator: z.object({
+      name: z.literal('k6-studio'),
+      version: z.string(),
+    }),
```
allansson: Do we want to impose this restriction? It does mean that trying to open any other kind of HAR file will fail.

src/schemas/recording/index.ts
```diff
@@ -1,2 +1,52 @@
...
+    pages: z.unknown().transform((value) => value as Page[]),
+    entries: z
+      .unknown()
+      .transform((value) => value as EntryWithOptionalResponse[]),
```
allansson: Maybe we should leave a comment here explaining why we're doing the `.transform` magic.

allansson (approved):
LGTM! Left some food for thought. If you decide to do anything about it, feel free to re-request review.

src/types/index.ts
```diff
@@ -49,7 +49,7 @@ export interface Response extends Metadata {
   cookies: Cookie[]
   reason: string
   statusCode: number
-  content: string
+  content: string | null
```
going-confetti, line 52: Learned this the hard way 😒

src/main/proxy.ts
```diff
@@ -98,6 +98,7 @@ export const launchProxy = (
       return
     }
 
+    // TODO: add zod schema validation
```
going-confetti, line 101: Seems pretty important TBH

src/schemas/recording/har/index.ts
```diff
@@ -0,0 +1,98 @@
+import { z } from 'zod'
+
+// Based on the historical knowledge collected in https://github.com/grafana/har-to-k6/blob/master/li-har.spec.md
```
going-confetti: 🙌

src/schemas/recording/har/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  })
+  .passthrough()
+
+export const LogSchema = z.object({
+  version: z.string(),
+  creator: z.object({
+    name: z.string(),
+    version: z.string(),
+  }),
+  pages: z.array(HarPageSchema).optional(),
```
allansson: After some thinking, I think we should stick with the transform-as hack here and make a separate PR for the general HAR schema. I think this PR is big enough without it and since the new parsing is potentially breaking, it would be good to get more eyes on it than mine.

allansson (changes requested):
See my comment

allansson (approved):
LGTM! 👍

PR #869, opened 2025-10-23
chore: Azure trusted signing for windows binaries

Llandy3d:
## Description

This PR migrates from the use of certificates file to sign the windows binaries to using the Azure Trusted signing cloud service.

## How to Test

Make a release or test the draft release `1.8.0-azure2` on a windows machine, confirm that no warning appear regarding the application being unsigned and confirm that it mentions `by Grafana Labs`.

## Checklist

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -7,11 +7,11 @@ import { MakerZIP } from '@electron-forge/maker-zip'
 import { FusesPlugin } from '@electron-forge/plugin-fuses'
 import { VitePlugin } from '@electron-forge/plugin-vite'
 import type { ForgeConfig, ForgeMakeResult } from '@electron-forge/shared-types'
-import { spawn } from 'node:child_process'
 import path from 'path'
 
 import { CUSTOM_APP_PROTOCOL } from './src/main/deepLinks.constants'
 import { getPlatform, getArch } from './src/utils/electron'
+import { spawnSignFile } from './src/utils/signing'
 import { windowsSign } from './windowsSign'
 
 function getPlatformSpecificResources() {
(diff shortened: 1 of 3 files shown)


.github/workflows/release-test-version.yml
```diff
@@ -105,12 +102,61 @@ jobs:
...
       - name: publish Windows
         if: startsWith(matrix.platform, 'windows-')
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           # windows cert
-          WINDOWS_CERTIFICATE_PATH: ./certificate.pfx
+          TRUSTED_SIGNING_ACCOUNT: grafana-premium-eastus
+          TRUSTED_SIGNING_ENDPOINT: https://eus.codesigning.azure.net/
+          TRUSTED_SIGNING_PROFILE: grafana-production
+          SIGNTOOL_PATH: ${{ steps.install-sign-tool.outputs.sign-tool }}\sign
```
Llandy3d, line 157: We need to use a specific version of sign tool for making the authentication work with the cloud service

forge.config.ts
```diff
@@ -25,7 +27,75 @@ function getPlatformSpecificResources() {
...
+
+    await Promise.all(signingPromises)
+    return options
+  }
+}
+
 const config: ForgeConfig = {
+  hooks: {
+    postMake: getPostMakeHook(),
+  },
```
Llandy3d, line 54: This is an hack for signing the binary because by default there are quite a few issue with the `MakerSquirrel` maker that we use for the windows installer.

The main one is that it seems to have a bug that ignore configurations, so if you specify a signtool_path and arguments for it, they will be ignored.
The work around was for it to fail to sign files (might be unrelated) and proceed to manually sign the files in the `postmake` hook of electron-forge on windows.

Although it's working, we should invest more time into finding a proper solution to this issue but it will require diving deeper into the libraries that forge uses for signing.

```
❯ Making a squirrel distributable for win32/x64
stderr: Some files do not exist.  Try using a different --base-directory value or a fully qualified file path.

stderr:     C:\Users\runneradmin\AppData\Local\Temp\electron-packager\tmp-NH4s0z\ffmpeg.dll
```
This is an example of failure of signing from squirrel even though we are not specifying that it needs to sign files directly in the maker config.

There is also these errors popping out in the CI that seem to auto-solve while signing attempts are made:

```
fail: Azure.Identity[10]
      False MSAL 4.73.1.0 MSAL.NetCore .NET 8.0.21 Microsoft Windows 10.0.26100 [2025-10-23 00:46:47Z - 3b1e043c-8aac-4cbd-84e8-47d5faf6dd92] Error message: [Managed Identity] Authentication unavailable. Either the requested identity has not been assigned to this resource, or other errors could be present. Ensure the identity is correctly assigned and check the inner exception for more details. For more information, visit https://aka.ms/msal-managed-identity.
      Status: BadRequest
      Content:
      {"error":"invalid_request","error_description":"Identity not found"}
  ```

File: .github/workflows/release-test-version.yml
Comment (going-confetti) [inline .github/workflows/release-test-version.yml]: These are no longer needed, it seems

File: forge.config.ts (hunk 2 above)
Comment (going-confetti) [reply forge.config.ts:54]: I think it's worth adding a comment with a concise description of the hack.

> we should invest more time into finding a proper solution to this issue but it will require diving deeper into the libraries that forge uses for signing

Agree, let's create a ticket (I'd even add a link to it in the aforementioned comment)

File: forge.config.ts [hunk 3]
```diff
@@ -25,7 +27,75 @@ function getPlatformSpecificResources() {
...
 
+function getPostMakeHook() {
+  // we use the hook function only on windows to sign the binary so in
+  // all other cases we just return an empty object
+  if (getPlatform() !== 'win') {
+    return (forgeConfig: ForgeConfig, options: ForgeMakeResult[]) =>
+      Promise.resolve(options)
+  }
+
+  return async (forgeConfig: ForgeConfig, options: ForgeMakeResult[]) => {
```
going-confetti: I realize that this is a pretty nitpicky comment, but it should probably be `makeResults` and not `options` 🤔

windowsSignHook.ts
```diff
@@ -0,0 +1,39 @@
...
+    '-td',
+    'sha256',
+    '-fd',
+    'sha256',
+    '--trusted-signing-account',
+    process.env.TRUSTED_SIGNING_ACCOUNT!,
+    '--trusted-signing-certificate-profile',
+    process.env.TRUSTED_SIGNING_PROFILE!,
+    '--trusted-signing-endpoint',
+    process.env.TRUSTED_SIGNING_ENDPOINT!,
```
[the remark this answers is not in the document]
going-confetti: I think this is a good comment,adding explicit validation can help us debug the workflow, if anything breaks.

forge.config.ts
```diff
@@ -25,7 +27,75 @@ function getPlatformSpecificResources() {
...
+          '-td',
+          'sha256',
+          '-fd',
+          'sha256',
+          '--trusted-signing-account',
+          process.env.TRUSTED_SIGNING_ACCOUNT!,
+          '--trusted-signing-certificate-profile',
+          process.env.TRUSTED_SIGNING_PROFILE!,
+          '--trusted-signing-endpoint',
+          process.env.TRUSTED_SIGNING_ENDPOINT!,
```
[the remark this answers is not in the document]
going-confetti: Another good comment

.github/workflows/release-test-version.yml
going-confetti, line 1: Not related to this particular PR, but should probably look into reducing duplication across `release-test-version` and `release` in future: both workflows are quite long and hard to follow, I think there's a risk of them getting out of sync.

forge.config.ts
Llandy3d: no I think it makes complete sense, it was options from an example I took but then I had to add typing and it's 100% clear now that it's not an "option" at all 🙌

PR #870, opened 2025-10-24
internal: Add auto correlation feature

e-fisher:
## Description

Use OpenAI to create correlation rules automatically. Demo:

https://github.com/user-attachments/assets/0ae99eb6-2899-439e-90e1-99a7d15f7a59

## How to Test

You'll need an OpenAI API key to test it. There are 2 ways to add your key:
  - Open settings and select AI tab
  - Click "auto correlate" in a generator, and you'll see a button to open settings, if you haven't added a key before.

Use a recording that requires correlation and test it out. You may get 3 different outcomes - 
- Validation passes -> Correlation successful
- Some values were correlated, but validation doesn't pass -> Partial success with an explanation
- No values were correlated -> Failure with an explanation.

Make sure to test the app with auto correlation feature disabled, everything should work as before.

## Limitations and next steps
This is an internal preview and the feature will need additional polishing:
- Decide if we want to change correlation dialog layout, I've bootstrapped a 3-panel layout, but it's open for suggestions
- The prompt and tool/type description might require some tweaking
- Add more detailed introduction text
- Add user-friendly errors
- Add ability to retry auto correlation
- Limit max number of requests to AI - currently it bails out after few failed attempts and doesn't keep on looping if it cannot solve correlation, but we still need a fail-safe
- Handle cases when correlation is not needed (validation passes without any rules), show a message to user and a button to "continue anyway"
- Refactor and address TODOs.
- Collect and address internal feedback

## Checklist

### DIFF
--- a/src/utils/validateScript.ts
+++ b/src/utils/validateScript.ts
@@ -5,7 +5,10 @@ import { processProxyData } from './proxyData'
 /**
  * Validates a k6 script by running it and collecting proxy data
  */
-export async function validateScript(script: string): Promise<ProxyData[]> {
+export async function validateScript(
+  script: string,
+  signal?: AbortSignal
+): Promise<ProxyData[]> {
   let collectedData: ProxyData[] = []
 
   return new Promise((resolve, reject) => {
@@ -18,6 +21,15 @@ export async function validateScript(script: string): Promise<ProxyData[]> {
       unsubscribeProxyData()
       unsubscribeScriptFinished()
       unsubscribeScriptFailed()
+    }
+
+    // Handle abort signal
+    if (signal) {
+      signal.addEventListener('abort', () => {
+        cleanup()
+        window.studio.script.stopScript()
+        reject(new DOMException('Aborted', 'AbortError'))
+      })
     }
 
     // Set up proxy data listener
--- a/src/views/Generator/AutoCorrelation/useGenerateRules.ts
+++ b/src/views/Generator/AutoCorrelation/useGenerateRules.ts
@@ -1,5 +1,5 @@
 import { useChat } from '@ai-sdk/react'
-import { useCallback, useRef, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 
 import { applyRules } from '@/rules/rules'
 import {
@@ -32,6 +32,7 @@ export const useGenerateRules = ({
     useState<CorrelationStatus>('not-started')
   const [outcomeReason, setOutcomeReason] = useState('')
   const suggestedRulesRef = useRef(suggestedRules)
+  const abortControllerRef = useRef<AbortController | null>(null)
   const recording = useGeneratorStore(selectFilteredRequests)
   const generator = useGeneratorStore(selectGeneratorData)
 
@@ -129,7 +130,10 @@ export const useGenerateRules = ({
       recording
     )
 
-    const validationResult = await validateScript(script)
+    const validationResult = await validateScript(
+      script,
+      abortControllerRef.current?.signal
+    )
 
     const result = validationMatchesRecording(recording, validationResult)
     setIsValidationSuccessful(result.success)
@@ -147,26 +151,42 @@ export const useGenerateRules = ({
     setIsValidationSuccessful(false)
     setCorrelationStatus('validating')
 
-    const validationResult = await runValidation()
+    try {
+      const validationResult = await runValidation()
 
-    if (validationResult.success) {
-      setCorrelationStatus('correlation-not-needed')
-      setOutcomeReason(
-        'The script validation passed successfully. The current generator configuration is sufficient to handle all requests in the recording, so no additional auto-generated correlation rules are required.'
-      )
-      return
+      if (validationResult.success) {
+        setCorrelationStatus('correlation-not-needed')
+        setOutcomeReason(
+          'The script validation passed successfully. The current generator configuration is sufficient to handle all requests in the recording, so no additional auto-generated correlation rules are required.'
+        )
+        return
+      }
+
+      // TODO: we are sending just one failing request for context, should we send all?
+      return sendMessage({
+        text: `${systemPrompt} \n\n Validation result: ${JSON.stringify(validationResult)}`,
+      })
+    } catch (error) {
+      if (error instanceof Error && error.name === 'AbortError') {
+        return
+      }
+      console.error(error)
+      setCorrelationStatus('error')
+      setOutcomeReason('An error occurred during auto-correlation.')
     }
-
-    // TODO: we are sending just one failing request for context, should we send all?
-    return sendMessage({
-      text: `${systemPrompt} \n\n Validation result: ${JSON.stringify(validationResult)}`,
-    })
   }
   function stop() {
-    window.studio.script.stopScript()
     void stopGeneration()
     setCorrelationStatus('aborted')
+    abortControllerRef.current?.abort()
   }
+
+  useEffect(() => {
+    abortControllerRef.current = new AbortController()
+    return () => {
+      abortControllerRef.current?.abort()
+    }
+  }, [])
 
   return {
     start,


src/store/features/useFeaturesStore.ts
```diff
@@ -11,7 +11,8 @@ interface FeaturesStore {
 export const defaultFeatures: Record<Feature, boolean> = {
   'dummy-feature': false,
   'typeahead-json': false,
-  'auto-correlation': false,
+  // TODO: change to false before mergin
+  'auto-correlation': true,
```
e-fisher: Will remove this before merge - keeping it enabled by default to make it easier to test and create a custom build to share

.eslintrc.cjs
```diff
@@ -41,6 +41,7 @@ module.exports = {
   rules: {
     'react/no-unknown-property': ['error', { ignore: ['css'] }],
     'no-param-reassign': 'error',
+    'unused-imports/no-unused-imports': 'warn',
```
e-fisher, line 44: A small QoL change: pre-commit hook will remove unused imports automatically

src/views/Generator/AutoCorrelation/utils/IPCChatTransport.ts
```diff
@@ -0,0 +1,89 @@
...
+  UIMessageChunk,
+} from 'ai'
+
+import { StreamChatRequest, StreamChatChunk } from '@/handlers/ai/types'
+
+/**
+ * Custom ChatTransport implementation that uses Electron IPC for communication
+ * between renderer and main process for AI streaming responses.
+ */
+export class IPCChatTransport<UI_MESSAGE extends UIMessage>
```
e-fisher: If we decide to move logic from main process to an API, it would be a matter of changing transport

src/views/Generator/TestRuleContainer/TestRuleContainer.tsx
```diff
@@ -25,25 +31,72 @@ export function TestRuleContainer() {
...
+                      onClick={() => setIsAutoCorrelationDialogOpen(true)}
+                    >
+                      <WandSparkles />
+                      Auto correlate
+                    </Button>{' '}
+                    <NewRuleMenu
+                      variant="solid"
+                      size="2"
+                      color="orange"
+                      text="Add rule manually"
```
e-fisher, line 93: Changed button text as a way to promote auto correlation, "do it automatically OR do it manually"

src/main/settings.ts
```diff
@@ -98,7 +99,9 @@ export async function saveSettings(settings: Partial<AppSettings>) {
   const currentSettings = await getSettings()
   const newSettings = { ...currentSettings, ...settings }
 
-  newSettings.ai.apiKey = processApiKeyForStorage(newSettings.ai.apiKey)
+  if (currentSettings.ai.apiKey !== newSettings.ai.apiKey) {
+    newSettings.ai.apiKey = processApiKeyForStorage(newSettings.ai.apiKey)
+  }
```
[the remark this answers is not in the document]
e-fisher: If the condition is removed, the API key will be encrypted every time the settings are saved, even when key didn't change, resulting in broken API key. This is working as expected - when the user adds a new key, it won't be equal to the stored encrypted key.

src/types/autoCorrelation.ts
```diff
@@ -0,0 +1,113 @@
...
+  and used in subsequent requests.
+
+  Examples:
+
+  1. A session token is returned in a login response and used in subsequent requests
+  2. A user ID is returned in a create user response and used in subsequent requests
+  3. A resource ID is returned in a create resource response and used in subsequent requests
+  4. A CSRF token is returned in a form response and used in subsequent form submissions
+
+  The matched values will be replaced in subsequent request by exact text match
```
going-confetti, line 115: Does this mean it currently doesn't support custom replacers?

src/main/settings.ts
```diff
@@ -98,7 +99,9 @@ export async function saveSettings(settings: Partial<AppSettings>) {
   const currentSettings = await getSettings()
   const newSettings = { ...currentSettings, ...settings }
 
-  newSettings.ai.apiKey = processApiKeyForStorage(newSettings.ai.apiKey)
+  if (currentSettings.ai.apiKey !== newSettings.ai.apiKey) {
```
going-confetti, line 83: Totally forgot to add this when reviewing the previous PR, but there's a small issue in the Settings form: when you insert a key, it shows the following warning in the console

Image

src/types/autoCorrelation.ts
```diff
@@ -0,0 +1,113 @@
...
+])
+
+const CorrelationExtractorSchema = z.object({
+  filter: FilterSchema,
+  selector: ExtractorSelectorSchema,
+  extractionMode: z.enum(['single', 'multiple']).default('single'),
+})
+
+export const AiCorrelationRuleSchema = z.object({
+  id: z.string().describe('Unique identifier for the rule'),
```
going-confetti, line 100: Somehow, it feels odd to ask an LLM to generate a unique id, but maybe I'm overthinking this 🤔

src/types/auto-correlation.ts
```diff
@@ -0,0 +1,113 @@
...
+  BeginEndSelectorSchema,
+  RegexSelectorSchema,
+  JsonSelectorSchema,
+  HeaderNameSelectorSchema,
+])
+
+const CorrelationExtractorSchema = z.object({
+  filter: FilterSchema,
+  selector: ExtractorSelectorSchema,
+  extractionMode: z.enum(['single', 'multiple']).default('single'),
```
going-confetti: Should this also be described? Wouldn't it ever even pick `multiple` otherwise? Maybe we can even simplify by using `z.enum(['single'])`.

src/views/Generator/AutoCorrelation/IntroductionMessage.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {isAiConfigured && (
+          <Button onClick={onStart} size="4">
+            <WandSparkles />
+            Analyze recording
+          </Button>
+        )}
+        {!isAiConfigured && (
+          <>
+            <Text size="2" color="gray" align="center">
+              To use auto correlation, you{'`'}ll need to configure your OpenAI
```
going-confetti: An apostrophe (`{"'"}`) should be used here instead of the grave accent/backtick (`you'll` instead of ``you`ll``)

src/views/Generator/AutoCorrelation/IntroductionMessage.tsx
```diff
@@ -0,0 +1,81 @@
...
+    >
+      <img
+        src={grotIllustration}
+        role="img"
+        aria-label="Grafana mascot illustration"
+        css={{ maxWidth: 250 }}
+      />
+
+      <Flex direction="column" align="center" gap="4" maxWidth="500px" mt="4">
+        <Text size="3" color="gray" align="center">
```
going-confetti: When I saw this component for the first time, I thought it was zoomed in. Here's why:
1. We don't use font-size: 16px for body text often, so this particular paragraph looks bigger than what you typically see in the app (having said that, I did find one similar case in the auth dialog)
2. The CTAs are too big - this includes both the buttons size itself and the font size of the button text.

Is this intentional? 

Image

src/views/Generator/AutoCorrelation/IntroductionMessage.tsx
```diff
@@ -0,0 +1,81 @@
...
+          Let AI analyze your recording and automatically create correlation
+          rules to make your test repeatable.
+        </Text>
+
+        <Flex direction="column" gap="2" align="start" width="100%" mb="6">
+          <ListItem>Validation will run automatically</ListItem>
+          <ListItem>Recording will be analyzed for dynamic values</ListItem>
+          <ListItem>
+            Correlation rules will be created to make your test repeatable
+          </ListItem>
```
going-confetti: Nit (because this isn't user facing yet): the general recommendation in our UX writing guide is to [use active voice](https://grafana.com/docs/writers-toolkit/write/style-guide/ux-writing/#use-active-voice). I think it would be an improvement, but I don't have any concrete suggestions at the moment.

src/views/Generator/AutoCorrelation/AutoCorrelationDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+  open,
+  onOpenChange,
+}: AutoCorrelationDialogProps) {
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
+      <Dialog.Content maxWidth="1200px" height="750px" size="2" asChild>
+        <Flex direction="column">
+          <Dialog.Title>
+            <Flex justify="between">
+              <Flex align="center">Auto correlation</Flex>
```
going-confetti: I would like to hear a native speaker's opinion on this, but I believe this should either be `autocorrelation` or `auto-correlation` (similar to [auto-complete](https://www.merriam-webster.com/dictionary/auto-complete))

src/views/Generator/AutoCorrelation/AutoCorrelation.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const handleDiscard = () => {
+    close()
+  }
+
+  if (correlationStatus === 'not-started') {
+    return <IntroductionMessage onStart={start} />
+  }
+
+  return (
```
going-confetti, line 72: I think the UI needs some work to better utilize the space and reduce duplication. I'll add concrete examples in this thread.

going-confetti: ### 1. First run
Both panels on the left don't seem useful at this stage: the top one is only used to convey the process status (One suggestion I have is to move it to the dialog title); the bottom one is only really useful at the very end of the process (and only if any rule suggestions have been made).

I think Allotment is not needed here: you can't really inspect requests in the dialog (the whole requests view gets blurred out at the end) and adjustments can be made to ensure there's enough horizontal space to fit the rule items.

Image

going-confetti: ### 2. Second run/re-validation(?)

If the script is already valid, It immediately tells me the validation was successful, but then runs it again in the background. Is this expected for all scripts or only for those that aren't valid yet? It looks a bit broken, because the "validation successful" message stays, but the requests list gets cleared and re-populated again. At the same time, it's completely blurred out, so it's not even immediately clear what's being added behind the blurred overlay:

Image

going-confetti: ### 3. End state
I think there's too much space used for essentially the same thing: Validation successful/Auto-correlation completed. I would argue that the latter is more important as it's the status of the whole process, but the former has a bigger and bolder styling. IMO The blurred out requests don't add any value either and as I suggested in a comment above, more space should be used for rules instead.

Speaking of rules, I noticed that sometimes the LLM becomes a bit too enthusiastic about correlating different dynamic values (see screenshot), so naturally I wanted to see if all these rules were still needed, but I couldn't find a way to re-run the validation. I believe this is an important missing piece of functionality because we offer the user choice that can change the validity of the script (they can theoretically uncheck the only "required" suggestion from the list on the screenshot, which would mean "validation successful" no longer reflect the actual state of things).

Image

Also, if the process is unsuccessful, but the LLM tried configuring rules, it keeps them in the list, even though they (supposedly) don't make the test better:

Image

going-confetti: Forgot to add one more comment about the dialog/its content: it doesn't work on all window sizes we support:

Image

I used the following hack in Settings some time ago:
```tsx
<Dialog.Content
  // This constraings could be anything. For example `1200px` and `750px` as it is in you core rn
  maxWidth="800px"
  maxHeight="640px"
  // This is the trick. It makes sure the dialog always fits on the screen 
  width="calc(100vw - 100px)"
  height="calc(100vh - 100px)"
>
```

src/types/autoCorrelation.ts
e-fisher: Correct, for the sake of keeping things simple I'm not including it in this stage, but will revisit before public preview

src/types/autoCorrelation.ts
e-fisher: Initially, I've planned to allow AI to edit rules as well, and since AI gives descriptive IDs (e.g., 'correlate_csrf') it would make it easier to navigate previously added rules. It seems to do this well, and I didn't encounter duplicates.

src/views/Generator/AutoCorrelation/IntroductionMessage.tsx
e-fisher: It was intentional, I wanted to give this some more "pop", but perhaps went too far :) Changed to smaller font and button 👍

src/views/Generator/AutoCorrelation/AutoCorrelationDialog.tsx
e-fisher: Looks like "autocorrelation" is the correct form, updated

src/views/Generator/AutoCorrelation/AutoCorrelation.tsx
e-fisher: I've moved status to the dialog title per your suggestion, and added outcome reason after the rule list. 

As for the "validation successful" message and blurring - I wanted to emphasize the moment when AI correlated successfully, the blur indicating you no longer need to inspect requests manually, it's taken care of. I agree it looked bad when correlation was not needed and requests were appearing behind the blur, but now that's fixed. Let me know if you have a strong preference about removing it. 

No revalidate functionality at this point, but I've added it to public preview tasklist.

src/utils/validateScript.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+      if (unsubscribeScriptFinished) {
+        unsubscribeScriptFinished()
+        unsubscribeScriptFinished = null
+      }
+      if (unsubscribeScriptFailed) {
+        unsubscribeScriptFailed()
+        unsubscribeScriptFailed = null
+      }
+    }
```
allansson:
```suggestion
    let unsubscribeProxyData = () => {}
    let unsubscribeScriptFinished = () => {}
    let unsubscribeScriptFailed = () => {}

    // Cleanup function to remove all listeners
    const cleanup = () => {
        unsubscribeProxyData()
        unsubscribeScriptFinished()
        unsubscribeScriptFailed()
    }
```

src/views/Generator/AutoCorrelation/utils/IPCChatTransport.ts
allansson: CAPS_CASE isn't really standard convention for generic names. I see that ChatTransport uses all caps, but I don't think we should. Maybe just `Message`?

src/views/Generator/AutoCorrelation/AutoCorrelation.tsx
```diff
@@ -0,0 +1,129 @@
...
+  useEffect(() => {
+    return () => {
+      window.studio.script.stopScript()
+      void stop()
+    }
+  }, [stop])
+
+  useEffect(() => {
+    onCorrelationStatusChange(correlationStatus)
+  }, [correlationStatus, onCorrelationStatusChange])
```
allansson, line 49: This kind of sync makes me think that the control over the correlation process (i.e. caller of `useGenerateRules`) should be owned by however passed the `onCorrelatationStatusChange` callback.

src/views/Generator/AutoCorrelation/Status.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <Flex align="center" gap="2">
+      <Text size="5" color={color} asChild>
+        {icon}
+      </Text>
+      <Text size="3" weight="medium" color={color}>
+        {text}
+      </Text>
+    </Flex>
+  )
+}
```
allansson: This could be inverted to make it feel more like React:

```tsx

export function Status({ correlationStatus }) {
  switch (correlationStatus) {
    case "not-started":
      return <StatusContent />

    case "correlation-not-needed":
      return (
        <StatusContent color="grass" icon={<CircleCheckIcon />}>
          Correlation not needed
        </StatuContent>
      )

    /* etc. */

  }

}

function StatusContent({ color, icon, children }) {
  return (
    <Flex align="center" gap="2">
      <Text size="5" color={color} asChild>
        {icon}
      </Text>
      <Text size="3" weight="medium" color={color}>
        {text}
      </Text>
    </Flex>
  )
}

```

src/views/Generator/AutoCorrelation/SuggestedRules.tsx
```diff
@@ -0,0 +1,89 @@
...
+  suggestedRules,
+  isLoading,
+  checkedRuleIds,
+  setCheckedRuleIds,
+  correlationStatus,
+}: SuggestedRulesProps) {
+  useEffect(() => {
+    // Suggested rules checked by default
+    setCheckedRuleIds(suggestedRules.map((rule) => rule.id))
+  }, [suggestedRules, setCheckedRuleIds])
```
allansson: Both `suggestedRules` and the `setCheckedRuleIds` callback are passed to this component. Maybe this logic should belong to the parent component?

src/views/Generator/AutoCorrelation/SuggestedRules.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { TestRuleInlineContent } from '../TestRuleContainer/TestRule/TestRuleInlineContent'
+import { TestRuleTypeBadge } from '../TestRuleContainer/TestRule/TestRuleTypeBadge'
+
+import { CorrelationStatus } from './types'
+
+interface SuggestedRulesProps {
+  suggestedRules: CorrelationRule[]
+  isLoading: boolean
+  checkedRuleIds: string[]
+  setCheckedRuleIds: (ruleIds: string[]) => void
```
allansson: `onCheckRules`

src/utils/validateScript.ts
allansson: Would this work?

package.json
```diff
@@ -77,9 +77,11 @@
...
   "private": true,
   "dependencies": {
-    "@dnd-kit/core": "^6.3.1",
-    "@dnd-kit/modifiers": "^9.0.0",
-    "@dnd-kit/sortable": "^10.0.0",
+    "@ai-sdk/openai": "^2.0.23",
+    "@ai-sdk/react": "^2.0.30",
+    "@dnd-kit/core": "^6.1.0",
+    "@dnd-kit/modifiers": "^7.0.0",
+    "@dnd-kit/sortable": "^8.0.0",
```
allansson: Am I reading this correctly or have the @dnd-kit versions been lowered? Maybe `main` hasn't been merged recently?

allansson (commented):
Left some comments on the code. Overall it looks good. 👍 

Some thoughts on the UX:

1. `Close` is a text button and not an X icon like in other dialogs.
2. The top border shouldn't be there in that initial "promotional" view (compare with Cloud auth dialog). It makes sense while the auto-correlation is running so it's just a case of the wrong elements having the border I think.
3. The animations are a bit too much for my taste. We don't really have any animations in other places either.
4. Seeing the progress is great, but I think the blur sticks out a bit too much. If the information in the view isn't needed we can just throw it away and remove it. An ugly mockup (make it work, designers):
    Image
    
    I think this gives it a bit more of a wizard feel. 

4. Either way, I think the dialog buttons should be to the bottom right of the dialog, because that's the convention. There should also be a Cancel button there until validation has finished.

    Image

5. If a recording is missing, the "Auto-correlate" button is disabled but "Add rule" is not. It looks a bit off. 

    Image

    Maybe the auto-correlate could have a step to select the recording if there is none?

I'm not sure if we want to act on any of these now or if we should just merge and gather feedback from more people. It might be a waste of time to do it now only to have to go back and change it again later.

And just a reflection on what I'd imagine the workflow would look like 99% of the time:

1. Open a recording
2. Click "Create test" and pick "HTTP test"
3. I'm taken through a wizard to configure the generator
    1. Select domains (allow list)
    2. "Would you like to run the auto-correlation?"
        1. Configure your token, if not configured (save token/don't save token)
        2. Run correlation using the UI currently in the dialog
    4. Done
4. I'm shown the rule editor

_(If you create a test generator via New generator, then we'd just add a "Pick or start a recording" step as the first step of the wizard)_.

In this flow, the auto-correlation wouldn't need to be promoted as much because promoting it would be a part of the normal flow when going from a recording to a generator.

We also wouldn't need a dialog. Running the auto-correlation would just show the auto-correlation wizard step.

allansson (dismissed):
I've tried it out and all in all I think it's good enough for internal testing.

One thing I've noticed though is that stopping the correlation by clicking stop or X doesn't seem to stop anything. So if I run the correlation again before the previous correlation completed, then both correlations start sending requests to the process and it becomes chaos:

https://github.com/user-attachments/assets/579d9fea-2a2f-42a3-bb49-51780307429e

It also seems flaky and can get stuck. Running it twice yields different results (which is to be expected, but the quality can range from decent to abysmal). 

Here it fails to detect that the id in the `/api/streak/complete/6493e877-4e48-4c8a-b9b8-bb129dfc3668` request came from an earlier response. Instead it focuses down on trying to correlate the cookie for every request.

https://github.com/user-attachments/assets/b838a0f8-1dca-438a-bb2e-f0ad41ee2369

Regarding the feature as a whole there's one thing I would like it to do which it doesn't do right now and that is to suggest correlations that are _likely_ to be related. 

Here's the workflow I tested (using my vibe coded test app):

1. Place an order
2. Copy the order number
3. Go into the admin pages
4. Find the order
5. Mark it as shipped

You'd want the app to suggest that the order id that was generated when placing the order should be used in the request to mark it as shipped.

If understand things correctly, we currently check that all the status codes line up correctly. In this case, the request for marking it as shipped will still succeed because the order that was created while recording still exists.

https://github.com/user-attachments/assets/0b7bd6ff-504d-404f-ad32-b9b7ba5873eb

This being said, I think it's still worth merging and then we can iterate more on the prompt, etc. in separate PRs.

PR #871, opened 2025-10-27
feat(browser): Record browser interactions over Chrome DevTools Protocol (experimental)

allansson:
## Description

This PR implements support for recording user interactions over Chrome DevTools Protocol instead of the browser extension. The new recorder is behind a setting for now, but in the long term the intention is to move away from using an extension at all.

### Network only

1. Go to Settings -> Recorder
2. Disable browser recording
3. Start a recording

The in-browser UI should not load and network traffic should be recorded as usual.

### Browser extension

1. Go to Settings -> Recorder
2. Under Browser Recording, select "Using browser extension"
3. Start a recording

The in-browser UI should load and both network and user interactions should be recorded as usual. Highlighting selectors and clicking links in k6 Studio should work as previously.

### CDP

1. Go to Settings -> Recorder
2. Under Browser Recording, select "Using Chrome DevTools Protocol"
3. Start a recording

The in-browser UI should load and both network and user interactions should be recorded as usual. Highlighting selectors and clicking links in k6 Studio should work as previously.

> #### NOTE: One difference between the extension and CDP is that when you click a link in the event log, CDP will record a navigation while the extension will not. Making CDP behave like the extension adds considerable complexity and since there isn't really a _right_ answer to whether we should record a navigation or not I intend to leave it like it is.

## Checklist

### DIFF
--- a/extension/src/frontend/recording.ts
+++ b/extension/src/frontend/recording.ts
@@ -10,6 +10,7 @@ import {
   isNativeButton,
   isNativeCheckbox,
   isNativeRadio,
+  isNonButtonInput,
 } from '../utils/dom'
 
 import { WindowEventManager } from './manager'
@@ -64,7 +65,7 @@ export function startRecording(client: BrowserExtensionClient) {
     // We don't want to capture clicks on form elements since they will be
     // interacted with using e.g. the `selectOption` or `type` functions.
     if (
-      clickTarget instanceof HTMLInputElement ||
+      isNonButtonInput(clickTarget) ||
       clickTarget instanceof HTMLTextAreaElement ||
       clickTarget instanceof HTMLSelectElement ||
       clickTarget instanceof HTMLOptionElement
--- a/extension/src/utils/dom.ts
+++ b/extension/src/utils/dom.ts
@@ -131,5 +131,21 @@ export function isNativeButton(element: Element) {
     return false
   }
 
-  return element.type === 'button' || element.type === 'submit'
+  return (
+    element.type === 'button' ||
+    element.type === 'submit' ||
+    element.type === 'reset'
+  )
+}
+
+export function isNonButtonInput(element: Element) {
+  if (element instanceof HTMLInputElement === false) {
+    return false
+  }
+
+  return (
+    element.type !== 'button' &&
+    element.type !== 'submit' &&
+    element.type !== 'reset'
+  )
 }
--- a/vite.extension.config.mts
+++ b/vite.extension.config.mts
@@ -20,6 +20,7 @@ export default defineConfig((env) => {
   ]
 
   const build: BuildOptions = {
+    minify: false,
     outDir: `.vite/build/extension`,
     sourcemap: 'inline',
   }


extension/src/cdp/index.ts
```diff
@@ -0,0 +1,16 @@
+import { startRecording } from '../frontend/recording'
+import { initializeView } from '../frontend/view'
+
+import { client } from './routing'
+import { configureStorage } from './storage'
+import { isInFrame } from './utils'
+import { trackTabFocus } from './window'
+
+// CDP will inject this script into all frames, but we only want to run it in the top frame.
```
allansson, line 9: This is at least true until we get around to adding support for frames.

extension/src/cdp/window.ts
```diff
@@ -0,0 +1,35 @@
+import { BrowserExtensionClient } from '../messaging'
+
+let wasFocused = false
+
+export function trackTabFocus(client: BrowserExtensionClient) {
```
allansson, line 5: CDP doesn't have a way of tracking which window or tab is currently active, so we have to report this from the browser itself.

extension/src/cdp/window.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  window.addEventListener(
+    'focus',
+    () => {
+      checkFocus()
+    },
+    true
+  )
+
+  setInterval(checkFocus, 200)
```
allansson, line 33: Knowing the active tab is only useful when interacting with the browser from studio, e.g. when clicking a link in the event log. A 200ms interval might be overkill, but what's the harm?

extension/src/frontend/recording.ts
```diff
@@ -0,0 +1,254 @@
...
+  findInteractiveElement,
+  isNativeButton,
+  isNativeCheckbox,
+  isNativeRadio,
+} from '../utils/dom'
+
+import { WindowEventManager } from './manager'
+import { getTabId } from './utils'
+
+export function startRecording(client: BrowserExtensionClient) {
```
allansson, line 19: This is essentially the same code that was in `extension/src/frontend/index.ts`, except:

- It's been wrapped in a function
- The `tab` property of recorded events use `getTabId` instead of `''`.

scripts/cdp/generateCdpClient.ts
allansson, line 1: You might ask "Why go through the trouble of creating a generator for the entire CDP when we only need a few methods?". With a generator we can detect breaking changes in CDP by:

1. Updating the `devtools-protocol` package to the latest version
2. Generating a new client

If CDP has breaking changes impacting the methods we're using, then we'll get type errors because the client will be different.

allansson: I've tried to keep the code here clean, but I don't think we should pay _too_ much attention to it. It just needs to be good enough to squeeze out a working client.

src/handlers/browser/recorders/cdp/index.ts
```diff
@@ -0,0 +1,247 @@
...
+
+type InitState = 'init' | 'spawned' | 'ready'
+
+const DEBUGGER_PORT = 9222
+
+const BROWSER_CDP_ARGS = [
+  '--disable-back-forward-cache',
+  // Disable web security to allow our script to be executed in sandboxed iframes.
+  '--disable-web-security',
+  '--allow-running-insecure-content',
```
allansson, line 32: The browser will load the script into every frame on a page. If the frame is sandboxed then the script will fail to load, an error be logged to the console and then the browser will wait for a few seconds and try again. This leads to an error being written every few seconds.

src/handlers/browser/recorders/cdp/page.ts
```diff
@@ -0,0 +1,192 @@
...
+      'frameStartedLoading',
+      'frameStoppedLoading',
+    ]
+
+    for (const eventName of events) {
+      this.#client.page.on(eventName, ({ data }) => {
+        console.log(eventName, data)
+      })
+    }
+  }
```
allansson, line 184: The behavior of these events are poorly documented so I've been using this to try and figure out which event fires when. I'm leaving it to aid in future debugging.

src/handlers/browser/index.ts
```diff
@@ -2,12 +2,13 @@ import { ipcMain, shell } from 'electron'
...
 import { waitForProxy } from '@/main/proxy'
-import { BrowserServer } from '@/services/browser/server'
 import { browserWindowFromEvent } from '@/utils/electron'
+import { HighlightSelector } from 'extension/src/messaging/types'
 
+import { BrowserLaunchError } from './recorders/types'
 import { BrowserHandler, LaunchBrowserOptions } from './types'
 
-export function initialize(browserServer: BrowserServer) {
+export function initialize() {
```
allansson, line 10: The `BrowserServer` is now created on-demand when starting a recording.

src/handlers/browser/launch.ts
allansson, line 1: I split this into separate files, one for each way of running a test:

- Network only
- Browser with extension
- Browser with CDP

src/main/k6StudioState.ts
```diff
@@ -1,16 +1,17 @@
...
+
 import { ProxyStatus } from '../types'
 import { AppSettings } from '../types/settings'
 
 import { type ProxyProcess } from './proxy'
 import { defaultSettings } from './settings'
 
 export type k6StudioState = {
-  currentBrowserProcess: ChildProcess | null
+  currentRecordingSession: RecordingSession | null
```
allansson, line 14: Use an abstraction instead of working with the `ChildProcess` directly. This solves the problem of disposing e.g. the browser server.

src/schemas/settings/v4/index.ts
```diff
@@ -0,0 +1,129 @@
+import { z } from 'zod'
+
+const RegularProxySettingsSchema = z.object({
```
allansson, line 3: Previous migrations were all referencing a bunch of schemas in `v1`. This makes it easy to accidentally change previous versions of the schema. From `v4` and onward each version should contain a complete copy of the entire schema.

src/schemas/settings/v4/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  })
+
+const RecorderDetectBrowserPathSchema = z.object({
+  detectBrowserPath: z.literal(true),
+  browserRecording: z.union([
+    z.literal('extension'),
+    z.literal('cdp'),
+    z.literal('disabled'),
+  ]),
```
allansson, line 67: This is the only change that's been made.

src/schemas/settings/README.md
```diff
@@ -6,10 +6,13 @@ If you're simply adding new options to the settings file, a migration is not nee
 
 ## Creating a new migration
 
-For when a new migration is needed:
+Let's say the current version is `v2` and you want to add a `v3` schema.
 
-1. Create a directory for the new version (e.g. `/v2`).
-2. Declare the new schema with appropriate changes.
+1. Create a copy of the `v2` directory and rename the copy to `v3`
```
allansson, line 11: I tried to clarify how to add migrations.

src/types/global.d.ts
```diff
@@ -0,0 +1,5 @@
+declare global {
+  export type EmptyObject = Record<string, never>
```
allansson, line 2: I'm probably the only one with this problem, but I got tired of typing `Record<string, never>` for e.g. empty event objects. It's a nice convenience for when you need something that contains absolutely nothing.

src/utils/cdp/transports/webSocket.ts
allansson, line 1: I initially used a websocket to connect to the browser because I thought there would be some benefit to using it, but as I dug deeper I couldn't find any good reason not to use stdio.

But we have the transport and it's working so I see no point in removing it. In the future we might want to have an option to select which transport to use.

src/utils/cdp/client.ts
allansson, line 1: This file is automatically generated, so there's no point in reviewing it.

src/utils/resources.ts
```diff
@@ -0,0 +1,21 @@
+import { app } from 'electron'
+import { readFile } from 'fs/promises'
+import path from 'path'
+
+const RESOURCES_ROOT_PATH = MAIN_WINDOW_VITE_DEV_SERVER_URL
+  ? path.join(app.getAppPath(), 'resources')
+  : process.resourcesPath
+
+const RESOURCE_INDEX = {
+  'browser-script': 'browser/index.js',
```
allansson, line 10: My idea is that we can add every resource we need to load here, e.g. the checks/group snippets.

vite.main.config.ts
```diff
@@ -57,7 +57,6 @@ export default defineConfig((env) => {
         org: process.env.SENTRY_ORG,
         project: process.env.SENTRY_PROJECT,
       }),
-      tsconfigPaths(),
```
allansson, line 60: Noticed that this plugin was duplicated, so I removed one instance of it.

allansson (commented):
Self-review

src/handlers/browser/launch.ts
```diff
@@ -1,255 +1,42 @@
...
-    })
-  } catch (e) {
-    log.error(e)
-    return undefined
-  }
+import { RecorderSettings } from '@/types/settings'
+import { exhaustive } from '@/utils/typescript'
+
+import { launchBrowserWithDevToolsProtocol } from './recorders/cdp'
+// import { launchBrowserWithExtension } from './recorders/extension'
```
going-confetti: 👀

scripts/cdp/generateCdpClient.ts
going-confetti: I read this comment, but by the time I reached `client.ts`, I'd already forgotten about it. So, the reason for storing `client.ts` in the repo instead of generating it at build time is clear to me now, but I think it might be worth adding a comment here in `generateCdpClient.ts`

going-confetti (commented):
This is really solid work and it works well 👏 
In my testing, I've only found one issue so far: if you use CDP with starting URLs, the in-browser toolbar doesn't appear until you refresh the page.
Code LGTM

allansson:
> LGTM, I wasn't able to reproduce starting URL issue @going-confetti mentioned

@e-fisher For some reason, this only happened on some specific pages (e.g. https://mui.com).

There were two issues:

- The `crypto` namespace wasn't initialized causing the app to throw when calling `randomUUID`. This has been fixed by using `nanoid` (https://github.com/grafana/k6-studio/pull/871/commits/8db0a98939600622afaf63fd89d3cdc3095ffd5e). This would have been an issue on any HTTP page, because the crypto namespace is only initialized in secure pages (i.e. HTTPS or localhost).
- For some reason, our script would mount the UI to an empty document which was then replaced with the actual document from the loaded page. I have no idea why this happens, but I managed to find a workaround (https://github.com/grafana/k6-studio/pull/871/commits/3fb97a2b875a63dc5e0020c3a607b0d7dfe1a7e7).

Could you do another review?

scripts/cdp/generateCdpClient.ts
allansson: This could actually be made a GitHub workflow check. Something like:

```sh
npm install
npm update --save devtools-protocol
npm run gen:cdp
npx run typecheck
```

Definitely a separate PR though.

Issue #767, opened 2025-07-15
Browser event drawer opens and closes when picking an element


allansson:
### Describe the bug

If the event drawer in the browser UI is open while picking an element, the drawer closes and then re-opens.

### Steps to reproduce the problem

1. Start a new recording.
2. Toggle the event drawer.
3. Select the picker tool.
4. Pick and element.

### Expected behavior

The element popup is shown and the drawer remains open.

### Actual behavior

The event drawer closes and then immediately opens again, showing the open animation.

### Grafana k6 Studio version

v1.5.0

### OS

Any

PR #874, opened 2025-10-30
fix(browser): Browser event drawer opens and closes when picking an element

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/767

## How to Test
- Start a new recording
- Open the event drawer
- Using the element inspector tool to pick an element on the page - the drawer should stay open without any flashing/sliding animation

## Checklist

### DIFF
--- a/extension/src/frontend/view/EventDrawer.tsx
+++ b/extension/src/frontend/view/EventDrawer.tsx
@@ -84,9 +84,10 @@ export function EventDrawer({ open, editing, onOpenChange }: EventDrawerProps) {
   return (
     <RecordingContext recording>
       <Dialog.Root modal={false} open={open} onOpenChange={onOpenChange}>
-        <Dialog.Portal container={container}>
+        <Dialog.Portal container={container} forceMount>
           <Dialog.Overlay />
           <Dialog.Content
+            forceMount
             css={css`
               position: fixed;
               top: 0;


extension/src/frontend/view/EventDrawer.tsx
```diff
@@ -123,16 +104,20 @@ export function EventDrawer({ open, editing, onOpenChange }: EventDrawerProps) {
               overflow-y: hidden;
               overscroll-behavior: contain;
 
+              /* Default closed state */
+              transform: translateX(100%);
+              transition: transform 0.3s cubic-bezier(0.22, 1, 0.36, 1);
+
               &[data-state='open'] {
```
going-confetti, line 114: I don't like this fix, because I can't fully explain why it works (and because it doesn't address the root cause at all). My current hypothesis is that when the element inspector popover opens, the drawer content is remounted (not sure why exactly, maybe something related to radix focus traps?), so the slide-in animation plays one more time (even though the value of the `open` prop doesn't change.

allansson (commented):
The drawer no longer slides in because the entire container is being unmounted. This was the reason I used animations to begin with. [Radix UI listens for animation events and only unmount components once the `animationend` event is fired.](https://www.radix-ui.com/primitives/docs/guides/animation). 

https://github.com/user-attachments/assets/75ff88ef-2799-464d-8fcc-fe90b999a4a9

This is still an improvement though. Better to have no animations at all than to have animations that don't work.

Question is: do we simply remove the slide or do we dig deeper until we figure it out?

PR #875, opened 2025-10-30
feat: Expand selection to parent elements when adding text assertions

allansson:
## Description

Before this PR, the text selection popover allowed users to type an arbitrary CSS selector. Now the popover uses the same element selector as the menu when picking elements. This makes more sense, since the text input would allow you to select elements which will never contain the selected text.

The reason for this change is to be able to generate `getBy` selectors when adding assertions. Allowing arbitrary input makes this near impossible.

### Before

https://github.com/user-attachments/assets/8961f820-7c98-427a-ab09-989bcf2a5d9c

### After

https://github.com/user-attachments/assets/25e98865-05ad-466b-a51d-2f2913b0b734

## How to Test

1. Start a recording
2. Select text selection tool and highlight some text
3. Use the left/right buttons in the popover header to expand and contract the selection.
4. Select the element picker tool and pick an element
5. Use the left/right buttons in the popover header to expand and contract the selection.

## Checklist

## Related PR(s)/Issue(s)

BEGIN_COMMIT_OVERRIDE
feat(browser): Expand selection to parent elements when adding text assertions
END_COMMIT_OVERRIDE

### DIFF
--- a/extension/src/frontend/view/ElementInspector/ElementInspector.hooks.ts
+++ b/extension/src/frontend/view/ElementInspector/ElementInspector.hooks.ts
@@ -1,91 +1,12 @@
-import { last } from 'lodash-es'
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 
-import { ElementSelector } from '@/schemas/recording'
-import { uuid } from '@/utils/uuid'
-import { ElementRole, getElementRoles } from 'extension/src/utils/aria'
-
-import { generateSelectors } from '../../../selectors'
 import { useGlobalClass } from '../GlobalStyles'
 import { useHighlightDebounce } from '../hooks/useHighlightDebounce'
 import { usePreventClick } from '../hooks/usePreventClick'
 import { Bounds, Position } from '../types'
-import { getElementBounds } from '../utils'
 
-function* getAncestors(element: Element) {
-  let current: Element | null = element.parentElement
-
-  while (current !== null) {
-    yield current
-
-    if (current === document.documentElement) {
-      break
-    }
-
-    current = current.parentElement
-  }
-}
-
-function findLabelFor(label: HTMLLabelElement): Element | null {
-  const id = label.getAttribute('for')
-
-  if (id === null) {
-    return null
-  }
-
-  return document.getElementById(id)
-}
-
-function findInChildren(label: HTMLLabelElement): Element | null {
-  return label.querySelector('input, select, textarea')
-}
-
-function findLabelledBy(label: HTMLLabelElement): Element | null {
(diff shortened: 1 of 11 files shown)


extension/src/frontend/view/ElementInspector/assertions/TextAssertionEditor.tsx
```diff
@@ -83,21 +47,6 @@ export function TextAssertionEditor({
 
   return (
     <AssertionForm onCancel={onCancel} onSubmit={handleSubmit}>
-      {canEditSelector && (
```
allansson, line 86: Feels good to remove this special-case

extension/src/frontend/view/ElementInspector/ElementInspector.hooks.ts
```diff
@@ -1,91 +1,12 @@
...
-
-function findLabelledBy(label: HTMLLabelElement): Element | null {
-  if (label.id === '') {
-    return null
-  }
-
-  return document.querySelector(`[aria-labelledby="${label.id}"]`)
-}
-
-export function findRelatedInput(element: Element): TrackedElement | null {
```
allansson, line 51: This function was not used so it's been removed.

extension/src/frontend/view/TextSelectionPopover.tsx
```diff
@@ -26,39 +27,45 @@ function TextSelectionPopoverContent({
...
 
   const handleChange = (assertion: TextAssertionData) => {
     setAssertion(assertion)
   }
 
   const handleSubmit = (assertion: TextAssertionData) => {
-    onAdd(assertion)
+    onAdd({
+      ...assertion,
+      selector: targetElement.selector.css,
```
allansson, line 49: Not the prettiest, but I think it beats trying to sync the selected element using `useEffect`.

extension/src/frontend/view/TextSelectionPopover.tsx
```diff
@@ -26,39 +27,45 @@ function TextSelectionPopoverContent({
...
 
-  useEffect(() => {
-    return () => {
-      client.send({
-        type: 'highlight-elements',
-        selector: null,
-      })
-    }
-  }, [])
+  const targetElement = selected ?? selection.element
```
allansson, line 38: This removes the need for multiple null checks.

e-fisher (approved):
Looks great! 🤩 

One small NIT: long selectors change popover width, you have to move cursor to keep expanding the selector, otherwise you'll close the popover by clicking on background. Perhaps it could be fixed width and show (...) when selector is too long?

https://github.com/user-attachments/assets/5bd21059-ad6f-453e-a9fb-ba6474a098f3

Issue #873, opened 2025-10-30
Strict mode violation when recording element with `data-testid`


allansson:
### Describe the bug

When recording a page that has multiple elements with the same `data-testid`, interacting with one of them will generate a script that causes the following error:

```
ERRO[0009] Uncaught (in promise) clicking on "internal:attr=[data-testid='data-testid Nav menu item']": strict mode violation, multiple elements returned for selector query  executor=shared-iterations scenario=default
```

The reason is that we're not checking whether the test id is unique while recording.

### Steps to reproduce the problem

1. Sign in to a Grafana stack
2. Expand an item in the sidebar menu
4. Export and run the script

### Expected behavior

The script should pass

### Actual behavior

Script fails with a strict mode violation:
```
ERRO[0009] Uncaught (in promise) clicking on "internal:attr=[data-testid='data-testid Nav menu item']": strict mode violation, multiple elements returned for selector query  executor=shared-iterations scenario=default
```

### Grafana k6 Studio version

v1.8.0

### OS

Any

Issue #833, opened 2025-09-16
Add support for emitting `getByRole` locators


allansson:
### Feature description

# What?

Since k6 v1.2.0, the browser modules support `getByRole` locators. These locators are significantly easier to understand than CSS selectors and are likely to be more reliable.

### Suggested solution (optional)

We already have the necessary code to determine the role of an element (currently used to provide context sensitive options in the element picker). What we need to do is determine the role along with the [accessible name ](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/) and record those in the events.

We should add this information in a new `aria` property where we could store all accessibility related information.

Once we have that information, we can make an informed decision during codegen on which type of locator to use.

Issue #876, opened 2025-11-03
Generate nested selectors for non-unique getBy-selectors


allansson:
### Feature description

We want to avoid using CSS locators as much as possible, and instead rely on the `getBy`-family of locators. In some cases we're unable to rely on these selectors because they don't uniquely match an element. 

### Suggested solution (optional)

The idea is to find a landmark element within which the selector _is_ unique and generate a nested locator from that.

Given the following DOM:

```html
<body>
  <header>
    <nav>
      <button>Create account</button>
    </nav>
  </header>
  <main>
    What are you waiting for? <button>Create account</button>
  </main>
</body>
```

Imagine that the user clicked the button in the main body. Starting from the `<body>` element, we'd query for `getByRole("button", { name: "Create account" })`. If the match isn't unique, then we move one step down the hierarchy, closer to the element and query again.

In this case, we'd be done in two steps:

1. `body`: 2 matches
2. `main`: 1 match

Once we know the landmark element, we generate a selector for it and use it for nesting. The generated code could look something like:

```ts
await page.locator("main")
  .getByRole("button", { name: "Create account", exact: true })
```

Even if the selector for the landmark element ends up being a CSS selector, the end results should still be more readable and stable than falling back to a single CSS selector.

PR #877, opened 2025-11-03
feat(browser): Generate getByRole selectors

allansson:
## Description

This PR adds support for recording ARIA-information and using it to emit `getByRole` selectors.

The recorder will take the role and accessible name of the target element and query the DOM to see if they are unique. If so, a role selector will be recorded.

And since I added a uniqueness check for getByRole I took the opportunity to do the same for `getByTestId`, fixing https://github.com/grafana/k6-studio/issues/873 in the process.

### Caveat

In some cases a unique selector can't be found, e.g. because the page has multiple buttons with the same text ('Get started', "Create account'). In this case no role selector will be recorded and the app will fallback to one of the other selectors.

The precedence of selectors is currently:

1. `getByRole`
2. `getByTestId`
3. `locator`

I already have a plan to improve on this: https://github.com/grafana/k6-studio/issues/876

## How to Test

1. Start a recording
5. Navigate a page
6. Stop the recording
7. Generate a script
8. Run the script

Whenever possible, the script should generate `getByRole` selectors and execute (reasonably) well.

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6-studio/issues/833
Fixes https://github.com/grafana/k6-studio/issues/873


extension/src/utils/aria.ts
```diff
@@ -134,3 +141,36 @@ export function getElementRoles(element: Element): Set<ElementRole> {
...
+  }
+
+  return Array.from(element.labels)
+    .map((label) => label.textContent)
+    .filter((label) => label !== null)
+}
+
+export function getAriaDetails(element: Element): AriaDetails {
+  const roles = [...getElementRoles(element)]
+    .filter((r) => !ABSTRACT_ROLES.has(r.role))
```
allansson, line 165: Some roles in ARIA are abstract, e.g. `input` which is implemented by `checkbox`, `radio` and `textbox`, which means that they aren't specific enough to be used for querying the DOM.

extension/src/target.ts
```diff
@@ -0,0 +1,89 @@
...
+    return undefined
+  }
+
+  const testId = element.getAttribute('data-testid')
+
+  if (testId === null || testId.trim() === '') {
+    return undefined
+  }
+
+  const matches = queryAllByTestId(document.body, testId)
```
allansson, line 62: Check for uniqueness of the test id, fixing https://github.com/grafana/k6-studio/issues/873

src/codegen/browser/codegen.test.ts
```diff
@@ -746,3 +746,120 @@ it('should assert that input has single value using toHaveValue', async ({
     '__snapshots__/browser/assertions/input/text-input-has-value.ts'
   )
 })
+
+it('should emit a getByTestId locator', async ({ expect }) => {
```
allansson, line 856: We didn't have a test for `getByTestId` so I added it.

src/codegen/browser/codegen.test.ts
```diff
@@ -746,3 +746,120 @@ it('should assert that input has single value using toHaveValue', async ({
...
+    },
+    scenarios: {},
+  })
+
+  await expect(script).toMatchFileSnapshot(
+    '__snapshots__/browser/locators/get-by-role.ts'
+  )
+})
+
+it('should emit a css locator', async ({ expect }) => {
```
allansson, line 940: Also didn't have it for `locator`, so added that too.

src/codegen/browser/selectors.ts
```diff
@@ -0,0 +1,75 @@
...
+    selector: selectors.css,
+  }
+}
+
+export function getNodeSelector(selector: ElementSelector): NodeSelector {
+  return (
+    getRoleSelector(selector) ??
+    getTestIdSelector(selector) ??
+    getCssSelector(selector)
+  )
```
allansson, line 60: Selects the first non-null selector in order of execution. `getCssSelector` is guaranteed to always be a selector.

src/components/BrowserEventList/EventDescription/Selector.tsx
```diff
@@ -1,23 +1,66 @@
...
+function SelectorIcon({ selector }: SelectorComponentProps) {
+  switch (selector.type) {
+    case 'css':
+      return <BracesIcon />
+
+    case 'test-id':
+      return <TestTubeDiagonalIcon />
+
+    case 'role':
+      return <SpeechIcon />
```
allansson: Suggestions for a better icon are welcome.

src/schemas/recording/browser/v2/index.ts
```diff
@@ -1,12 +1,25 @@
...
+})
+
 const ElementSelectorSchema = z.object({
   css: z.string(),
   testId: z.string().optional(),
+  role: RoleElementSelectorSchema.optional(),
 })
 
-const EventTargetSchema = z.object({
+const BrowserEventTargetSchema = z.object({
```
allansson, line 20: Renamed this to `BrowserEventTarget` to avoid conflicts with `EventTarget` already available in browsers.

allansson (commented):
Self-review

src/components/BrowserEventList/EventDescription/Selector.tsx
e-fisher: Perhaps a different icon for each role?

Image

Here are the icons I've used in the screenshot, role is typed as string so not sure if this covers all the cases
```ts
switch (selector.role) {
  case 'button':
    return <SquareMousePointer />
  case 'textbox':
    return <TextCursorInput />
  case 'link':
    return <Link2Icon />
  default:
    return <SpeechIcon />
}
```

allansson: Wow, that looks awesome! 😍 There are tons of roles, but I think we can cover the most common ones and then fallback to a generic for all the other ones.

Issue #784, opened 2025-08-19
Insert `waitForNavigation` when action is followed by a navigation


allansson:
### Feature description

Submitting a form or clicking a link will (often) cause a hard page navigation. In these cases, any actions following the navigation will be dependent on the navigation having completed.

To handle this [the docs](https://grafana.com/docs/k6/latest/using-k6-browser/recommended-practices/simulate-user-input-delay/#pagewaitfornavigation) recommend using `Promise.all` and `page.waitForNavigation` to block further execution until the next page has loaded. We should try to detect these interactions and generate the appropriate code.

### Suggested solution (optional)

Right now, we're only recording navigation that is initiated by the address bar or using the back/forward and reload buttons. We would need to expand upon the recorded navigations to include what we could call _implicit navigations_: navigations that happened as a side effect as some action performed on the page.

When generating code we could check if an action is directly followed by an implicit navigation and, if it is, generate code to wait both action and navigation at the same time:

```ts
await Promise.all([
  page.waitForNavigation(),
  locator("button[type=submit]").click()
])
```

PR #879, opened 2025-11-05
feat: Add waitForNavigation on link clicks and form submissions

e-fisher:
## Description

- Record implicit navigation events 
- Don't emit goto expressions when navigation is implicit
- Wrap actions that caused navigation in Promise.all and and waitForNavigation

### Link click
**Before:**

```ts
const page = await browser.newPage();

await page.goto("https://quickpizza.grafana.com/");

await page.locator(".font-bold > a").click(); // Navigation link

await page.locator("#username").fill("default");
```

**After:**

```ts
const page = await browser.newPage();

await page.goto("https://quickpizza.grafana.com/");

await Promise.all([
  page.waitForNavigation(),
  page.locator(".font-bold > a").click(), // Navigation link
]);

await page.locator("#username").fill("default");

```

### Form submit

**Before:**
```ts
const page = await browser.newPage();

await page.goto("https://quickpizza.grafana.com/my_messages.php");

await page.locator('[name="login"]').fill("admin");

await page.locator('[name="password"]').fill("123");

await page.locator("input:nth-of-type(3)").click(); // Submit button
```

**After:**
```ts
const page = await browser.newPage();

await page.goto("https://quickpizza.grafana.com/my_messages.php");

await page.locator('[name="login"]').fill("admin");

await page.locator('[name="password"]').fill("123");

await Promise.all([
  page.waitForNavigation(),
  page.locator("input:nth-of-type(3)").click(), // Submit button
]);
```

screencapture 2025-11-05 at 11 17 32@2x

## How to Test

Click links that cause hard page navigation, test with both extension recording and CDP

## Checklist

## Related PR(s)/Issue(s)

Resolves https://github.com/grafana/k6-studio/issues/784

### DIFF
--- a/src/components/BrowserEventList/EventDescription/PageNavigationDescription.tsx
+++ b/src/components/BrowserEventList/EventDescription/PageNavigationDescription.tsx
@@ -70,7 +70,7 @@ export function PageNavigationDescription({
       return <>Navigated to {url} using the browser history</>
 
     case 'implicit':
-      return null
+      return <>Navigated to {url} after a user action</>
 
     default:
       return exhaustive(event.source)


src/codegen/browser/code/scenario.ts
```diff
@@ -261,6 +261,35 @@ function emitExpectExpression(
...
+    .done()
+}
+
+function emitPromiseAllExpression(
+  context: ScenarioContext,
+  expression: ir.PromiseAllExpression
+): ts.Expression {
+  const expressions = expression.expressions.map((expr) =>
+    // No need to await inside Promise.all
+    new ExpressionBuilder(emitExpression(context, expr)).removeAwait().done()
```
e-fisher, line 283: We don't need to add `await` in `Promise.all` so it gets removed for all child expressions. I've considered 2 other options:
- Add condition to ExpressionBuilder.await and do an early return if parent expression is `Promise.all` - but we don't utilize `parent` property here
- Pass `Promise.all` expression to `emitExpression` as an additional `parent` argument and add a condition in `emitClickExpression` to skip .await when parent is `Promise.all` - but this would require adding similar condition to other expression which could potentially be wrapped in `Promise.all` in the future.

While adding and then removing await doesn't sound ideal, I think it was the most optimal solution out of those 3 - it will ensure that any expression wrapped in future won't have an await.

allansson: I had a similar solution in mind:

```ts
function unawait(expression: ts.Expresssion): ts.Expression {
  if (expression.type === "AwaitExpression") {
    return expression.value
  }

  return expression
}

const expressions = expression.expressions
  .map((expr) => emitExpression(context, expr))
  .map(unawait)
```

At some point I'd like to refactor things so that it's the consumer that decides whether it needs to await, but that's a bigger task.

extension/src/background/navigation.ts
```diff
@@ -22,6 +22,17 @@ function isHistoryNavigation({
...
+function isImplicitNavigation({
+  transitionType,
+  transitionQualifiers,
+}: WebNavigation.OnCommittedDetailsType) {
+  return (
+    (transitionType === 'link' &&
+      !transitionQualifiers.includes('forward_back')) ||
+    transitionType === 'form_submit'
+  )
+}
```
allansson, line 34: One thing to consider is that a navigation could be caused by a script as well, e.g. 

```tsx
return <button onClick={() => { window.location.href = "https://test.com/dashboard" }>Go to dashboard</button>
```

It's poor practice, but not unheard of. 

Not sure which combination of `transitionType` and `transitionQualifiers` this would produce but maybe it's worth doing a quick test to see if we can include it?

I'm also fine with leaving it be for now. Maybe a future PR?

src/codegen/browser/test.ts
```diff
@@ -130,14 +134,20 @@ function buildBrowserNodeGraph(events: BrowserEvent[]) {
         }
 
       case 'click': {
```
allansson, line 136: I was thinking of going for a more general approach of detecting whether two things happened in parallel and wrap things in `Promise.all` if that was the case.

But this is MUCH simpler and I like that. 😬

extension/src/background/navigation.ts
e-fisher: One of those rare cases when something worked unexpectedly 😄 I've created a page with js redirect button to test it out: https://testjsnav.oneapp.dev/

screencapture 2025-11-06 at 12 11 08@2x

src/codegen/browser/code/scenario.ts
e-fisher: I liked your approach better because unawait wouldn't need to be part of ExpressionBuilder, but had some issue with types, leaving as is for now

PR #881, opened 2025-11-06
feat(browser): Improved debugging experience for browser scripts

### LINKED ISSUE #839: VIsualize each action when running browser scripts in the validator
### Feature description

It is currently hard to know what is going on when executing browser scripts. Users are left hoping that any errors will be logged to the console. We need to provide a better way of debugging browser scripts so that users can understand why their tests aren't working.

Similar to how we instrument the `k6/http` module we should instrument the `k6/browser` module so that it reports back the status of each operation in the browser script. The actions could then be visualized as a table in the Validator UI with a success/fail status.

_No response_

allansson:
## Description

This PR implements a new view for debugging browser scripts. When running a browser script, the user is shown a list of each action performed and its status. 

https://github.com/user-attachments/assets/9e6d4e18-f3d6-470b-8a58-c9982cf7b1dd

Some overall UX changes:

* There are two top-level tabs: Script and Debugger. The view opens into the Script tab.
* The Validate button is now called "Debug script"
* The primary action is to run the script in cloud

The type of debugger to be displayed is determined by looking at the `options` export of the script. If it has browser configured, then the `BrowserDebugger` will be used.

Caveats:

* The actions explicitly covered in this PR are limited to:
    * `page.goto`
    * `page.waitForNavigation`
    * `page.reload`
    * `locator.check`
    * `locator.clear`
    * `locator.click`
    * `locator.dblclick`
    * `locator.fill`
    * `locator.focus`
    * `locator.hover`
    * `locator.press`
    * `locator.selectOption`
    * `locator.setChecked`
    * `locator.tap`
    * `locator.type`
    * `locator.uncheck`
    * `locator.waitFor`
* Value returning functions such as `innerHTML` will just have a generic `Call to `X` on page`.
* Nested locators are not supported and will be added in a future PR
* Assertions are not supported and requires some work in `k6-testing`
 
## How to Test

Run various scripts, both HTTP and browser, and check the debugger.

The most important part is to make sure the HTTP debugger works as expected.

## Checklist

## Related PR(s)/Issue(s)

Closes #839

### DIFF
--- a/src/views/Validator/Browser/BrowserDebugger.tsx
+++ b/src/views/Validator/Browser/BrowserDebugger.tsx
@@ -155,20 +155,20 @@ export function BrowserDebugger({
                     </Flex>
                   )}
                 </Flex>
-                {session.state === 'pending' && (
-                  <DebuggerEmptyState onDebugScript={onDebugScript}>
-                    Debug the script to inspect browser actions.
-                  </DebuggerEmptyState>
-                )}
-                {session.state !== 'pending' && (
-                  <AutoScrollArea
-                    tail={session.state === 'running' && tailActions}
-                    items={session.browserActions.length}
-                    onScrollBack={handleActionsScrollBack}
-                  >
+                <AutoScrollArea
+                  tail={session.state === 'running' && tailActions}
+                  items={session.browserActions.length}
+                  onScrollBack={handleActionsScrollBack}
+                >
+                  {session.state === 'pending' && (
+                    <DebuggerEmptyState onDebugScript={onDebugScript}>
+                      Debug the script to inspect browser actions.
+                    </DebuggerEmptyState>
+                  )}
+                  {session.state !== 'pending' && (
                     <BrowserActionList actions={session.browserActions} />
-                  </AutoScrollArea>
-                )}
+                  )}
+                </AutoScrollArea>
               </Flex>
             </Panel>
           </Group>
--- a/src/views/Validator/Browser/BrowserActionStatusIcon.tsx
+++ b/src/views/Validator/Browser/BrowserActionStatusIcon.tsx
@@ -1,3 +1,4 @@
+import { css } from '@emotion/react'
 import { Spinner } from '@radix-ui/themes'
 import { CircleCheckIcon, CircleMinusIcon, CircleXIcon } from 'lucide-react'
 
@@ -12,8 +13,26 @@ export function BrowserActionStatusIcon({
   event,
 }: BrowserActionStatusIconProps) {
   if (event.type === 'begin') {
-    return <Spinner />
+    return (
+      <div
+        css={css`
+          /* 
+          * Together with the width and height, this makes a total size 
+          * of 20px which matches the size of the status icons.
+          */
+          padding: 2px;
+        `}
+      >
+        <Spinner
+          css={css`
+            width: 16px;
+            height: 16px;
+          `}
+        />
+      </div>
+    )
   }
+
   switch (event.result.type) {
     case 'success':
       return <CircleCheckIcon color="var(--green-11)" />


src/codegen/browser/intermediate/index.ts
```diff
@@ -100,6 +100,13 @@ function emitLocatorNode(context: IntermediateContext, node: m.LocatorNode) {
       })
       break
 
+    case 'alt':
+    case 'label':
+    case 'text':
+    case 'placeholder':
+    case 'title':
+      throw new Error('Not implemented yet.')
```
allansson: I jumped the gun on these selectors a bit just so that we won't have to go back and update the debugger with selectors as we add them. I didn't want to go so far as to implement the codegen for them though.

src/components/Browser/ClickText.tsx
```diff
@@ -0,0 +1,64 @@
...
+      return exhaustive(button)
+  }
+}
+
+interface ClickTextProps {
+  pastTense?: boolean
+  details: ClickDetails
+}
+
+export function ClickText({ pastTense = false, details }: ClickTextProps) {
```
allansson: Shared between event list and actions list. Don't like the name so suggestions are welcome.

src/components/Browser/Locator.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return exhaustive(locator)
+  }
+}
+
+interface LocatorProps {
+  locator: NodeSelector
+  onHighlightChange?: (highlighted: boolean) => void
+}
+
+export function Locator({ locator, onHighlightChange }: LocatorProps) {
```
allansson, line 97: Split from `<Selector />` so that it can be shared between the event and action list. I'm referring to it as a `Locator` here because it's closer to what it's called in k6/browser.

src/components/Browser/RoleLocatorIcon.tsx
```diff
@@ -0,0 +1,120 @@
...
+  tree: ListTreeIcon,
+  treegrid: LayoutPanelLeftIcon,
+  treeitem: ListTreeIcon,
+}
+
+interface RoleLocatorIcon extends LucideProps {
+  selector: GetByRoleSelector
+}
+
+export function RoleLocatorIcon({ selector, ...props }: RoleLocatorIcon) {
```
allansson: Moved and renamed `RoleSelectorIcon` to `RoleLocatorIcon`.

src/components/Browser/SelectOptions.tsx
```diff
@@ -0,0 +1,44 @@
...
+  value?: string
+  label?: string
+  index?: number
+}
+
+interface SelectOptionsProps {
+  options: Array<SelectOption | string>
+}
+
+export function SelectOptions({ options }: SelectOptionsProps) {
```
allansson, line 13: Shared between event and action list. It was the `formatOptions` function before.

src/components/BrowserEventList/EventDescription/RoleSelectorIcon.tsx
allansson, line 1: This is `RoleLocatorIcon` now.

src/components/primitives/Kbd.tsx
```diff
@@ -7,6 +7,7 @@ export const Kbd = styled.kbd`
   font-size: 0.9em;
   color: var(--gray-12);
   padding: 0.2em 0.4em;
+  z-index: 0;
```
allansson, line 10: Without this the `z-index: -1` on the `::before` psuedo element would cause it to render behind some underlying element, meaning it wasn't visible in the action list.

src/components/Validator/ChecksSection.tsx
```diff
@@ -44,7 +44,7 @@ export function ChecksSection({ checks, isRunning }: ChecksSectionProps) {
                 >
                   <Table.Header>
                     <Table.Row>
-                      <Table.ColumnHeaderCell></Table.ColumnHeaderCell>
+                      <Table.ColumnHeaderCell width="20px"></Table.ColumnHeaderCell>
```
allansson, line 47: Without this, the column width would be distributed equally between all columns making the status icon column needlessly large.

src/components/Validator/ExecutionDetails.tsx
```diff
@@ -74,18 +76,20 @@ export function ExecutionDetails({
...
-          flex: 1;
-        `}
-      >
-        <ReadOnlyEditor
-          language="javascript"
-          value={script}
-          onCopy={handleCopy}
-        />
-      </Tabs.Content>
+      {script !== undefined && (
```
allansson, line 79: Unlike the validator, the `HttpDebugger` doesn't want to show the script here since it's in a separate top-level tab.

src/handlers/script/index.ts
```diff
@@ -44,8 +46,13 @@ export function initialize() {
       flag: 'r',
     })
 
+    const options = await new K6Client()
+      .inspect({ scriptPath: resolvedScriptPath })
+      .catch(() => ({}))
```
allansson, line 51: 💡Since we know the options we know which scenarios it has. It'd be pretty simple to add a dropdown to the debugger to allow debugging specific scenarios.💡

src/components/Browser/ClickPill.tsx
```diff
@@ -0,0 +1,100 @@
...
+      return exhaustive(button)
+  }
+}
+
+interface ClickPillProps {
+  pastTense?: boolean
+  details: ClickDetails
+}
+
+export function ClickPill({ pastTense = false, details }: ClickPillProps) {
```
allansson, line 73: Not in love with these names, tbh. But I can't think of anything better that is as concise.

src/components/BrowserEventList/BrowserEventList.tsx
```diff
@@ -46,6 +46,7 @@ export function BrowserEventList({
                       overflow: hidden;
                       white-space: nowrap;
                       text-overflow: ellipsis;
+                      font-size: var(--studio-font-size-1);
```
allansson, line 49: The action list in the debugger has the same font-size as the request list. Figured I'd update it here so that the browser event list is consistent with everything else.

src/main/runner/shims/browser/proxies/locator.ts
```diff
@@ -0,0 +1,143 @@
...
+import { ProxyOptions } from '../utils'
+
+export function locatorProxy(
+  target: Locator,
+  locator: ActionLocator
+): ProxyOptions<Locator> {
+  return {
+    target,
+    tracking: {
+      check(options) {
```
allansson, line 16: These are the methods that take a call, e.g. `locator.check()`, and convert them into an event payload.

src/main/runner/shims/browser/proxies/locator.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      waitFor(options) {
+        return {
+          method: 'locator.waitFor',
+          locator,
+          options,
+        }
+      },
+
+      $default() {
```
allansson: This would be a catch-all for any unimplemented functions. It's empty for now because I don't want to deal with this complexity right now.

src/types/global.d.ts
```diff
@@ -1,5 +1,7 @@
 declare global {
   export type EmptyObject = Record<string, never>
+
+  export type Falsy<T> = T | false | 0 | '' | null | undefined
```
allansson, line 4: Just a nice to have...

src/views/Validator/Browser/BrowserDebugger.tsx
```diff
@@ -0,0 +1,107 @@
...
+            </Flex>
+          </Allotment.Pane>
+          <Allotment.Pane minSize={40} maxSize={40}>
+            <BrowserDebugDrawer.List session={session} />
+          </Allotment.Pane>
+          <Allotment.Pane snap minSize={150} preferredSize={300}>
+            <Box height="100%" width="100%">
+              <BrowserDebugDrawer.Content session={session} />
+            </Box>
+          </Allotment.Pane>
```
allansson: This is a bit of a hack. The tab list is in its own fixed size pane so that we can snap the bottom drawer while still showing the tabs

https://github.com/user-attachments/assets/3904d0b6-7ed0-4433-a75f-7e52f5a209a0

src/views/Validator/Validator.tsx
```diff
@@ -117,22 +150,39 @@ function Content({ scriptPath }: ValidatorProps) {
...
+          </TabContent>
+          <TabContent value="debugger">
+            <Flex
+              css={css`
+                flex: 1 1 0;
+              `}
+              justify="center"
+            >
+              <Debugger
+                key={session?.id}
```
allansson: Clears component state whenever a new debug session has started.

vite.shims.config.ts
```diff
@@ -1,22 +1,27 @@
...
-      preserveEntrySignatures: 'strict',
-      input: {
-        browser: 'src/main/runner/shims/browser/index.ts',
-      },
-      output: {
-        entryFileNames: '[name].js',
+import { getBuildConfig } from './vite.base.config'
+
+export default defineConfig((env) => {
+  return mergeConfig(getBuildConfig(env as ConfigEnv<'build'>), {
```
allansson, line 7: Merges the electron-forge config so that hot reload works for the browser shim.

src/components/AutoScrollArea.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useAutoScroll } from '@/hooks/useAutoScroll'
+
+interface AutoScrollAreaProps extends Omit<ScrollAreaProps, 'onScroll'> {
+  tail?: boolean
+  items?: number
+  children?: ReactNode
+  onScrollBack?: () => void
+}
+
+export function AutoScrollArea({
```
going-confetti, line 13: 👏

going-confetti (approved):
LGTM

going-confetti (dismissed):

e-fisher (dismissed):
Got it, I think we can leave this for the future as it's non-blocking. LGTM! 🚀

natydej:
> Here's an updated version where the tabs only appear after clicking "Debug script" + the word wrap button removed:

Awesome, thanks for representing a new version to illustrate my thoughts.

One more thing, is there a case where people want to see the script and also see the debugger running at the same? if yes, I don't see the point to have tabs and switching them all the time, just thinking if tab is the best component to use here.

Why don't we offer to overview the script all the time and also provide the option to see the debugger on a side?

Something that we see in Visual Studio (see img attached), probably not as a bottom panel, but as a Sidebar panel, this avoids forcing users to choose between seeing the script and seeing the debugger.

image

On the UX side, tabs are good when users need two separate things, but here the debugger exists to provide feedback about the script.

One suggestion could be: 

- Left side: Script
- Right side: Debugger (collapsible)
- Toggle button (not sure where is the best placement yet): "Show debug script" or "Show debug panel"

> If we use icons on buttons, we should use them consistently throughout imo (again, the Grafana icon is a bit different because it's a promotion as well as a button).

Yes, exaclty. So why don't we keep the icon on the "Debug script" CTA next to the Grafana CTA? I think people are familiar with the icon of debug universally, since we are using icons buttons for the primary action as well. Same happen with the create test CTA and new recording (but that's for another discussion)
 
> It doesn't seem like a part of the PR though and something we could discuss further in an upcoming engineering sync.

Agreed, let's take this discussion for the weekly sync next week and create a specific issue for it.

going-confetti (dismissed):

PR #883, opened 2025-11-07
refactor: Simplify code when opening scripts

allansson:
## Description

This PR makes some simplifications to how scripts are opened.

1. It removes storing external script paths in push state. Both internal and external scripts now store the file path in the `filename` param.
2. It removes the `absolute` parameters from both `openScript` and `runScript` and instead relies on `path.isAbsolute`.
4. Whether or not a file is external is determined based on the absolute path of the script: if it is in the `SCRIPTS_PATH` folder, then it is not an external script.
5. The script is loaded using `useQuery` which simplifies the loading logic.

The only functional difference is that clicking `Open script` on the Home page will open a file dialog rather than taking the user to an empty script view. This reduces the complexity of the code and, frankly, leads to a better UX.

## How to Test

1. Open an internal script and run it
2. Open an external script from the Home page and run it
3. Open an external script from the menu in the Validator view and run it
4. Run a script in cloud

## Checklist

## Related PR(s)/Issue(s)

This PR is breaking out some of the changes from the implementation of #839, to make reviews more manageable.

### DIFF
--- a/src/handlers/script/index.ts
+++ b/src/handlers/script/index.ts
@@ -10,8 +10,7 @@ import { trackEvent } from '@/services/usageTracking'
 import { UsageEventName } from '@/services/usageTracking/types'
 import { browserWindowFromEvent, sendToast } from '@/utils/electron'
 import { TestRun } from '@/utils/k6/testRun'
-
-import { isExternalScript } from '../utils'
+import { isExternalScript } from '@/utils/workspace'
 
 import { ScriptHandler } from './types'
 
--- a/src/handlers/utils.ts
+++ b/src/handlers/utils.ts
@@ -1,7 +1,4 @@
 import { ipcMain, IpcMainEvent, ipcRenderer, IpcRendererEvent } from 'electron'
-import path from 'node:path'
-
-import { SCRIPTS_PATH } from '@/constants/workspace'
 
 // Create listener and return clean up function to be used in useEffect
 export function createListener<T>(
@@ -55,8 +52,4 @@ export function waitFor<T>({ event, signal, timeout }: WaitForOptions) {
 
     ipcMain.once(event, handleMessage)
   })
-}
-
-export function isExternalScript(scriptPath: string) {
-  return path.dirname(scriptPath) !== path.normalize(SCRIPTS_PATH)
 }
--- a/src/main/logger.ts
+++ b/src/main/logger.ts
@@ -3,7 +3,7 @@ import { BrowserWindow } from 'electron'
 import log from 'electron-log/main'
 import fs from 'fs/promises'
 import { spawn } from 'node:child_process'
-import path from 'node:path'
+import path from 'path'
 
 import { getPlatform } from '../utils/electron'
 
--- a/src/main/settings.ts
+++ b/src/main/settings.ts
@@ -2,7 +2,7 @@ import { app, BrowserWindow, dialog, nativeTheme } from 'electron'
 import log from 'electron-log/main'
 import { existsSync, readFileSync } from 'fs'
 import { writeFile, open } from 'fs/promises'
-import path from 'node:path'
+import path from 'path'
 
 import { configureSystemProxy } from '@/services/http'
 
--- a/src/utils/workspace.ts
+++ b/src/utils/workspace.ts
@@ -1,5 +1,6 @@
 import { existsSync } from 'fs'
 import { mkdir } from 'fs/promises'
+import path from 'path'
 
 import {
   DATA_FILES_PATH,
@@ -34,4 +35,8 @@ export const setupProjectStructure = async () => {
   if (!existsSync(DATA_FILES_PATH)) {
     await mkdir(DATA_FILES_PATH)
   }
+}
+
+export function isExternalScript(scriptPath: string) {
+  return path.dirname(scriptPath) !== path.normalize(SCRIPTS_PATH)
 }


src/handlers/script/index.ts
```diff
@@ -30,50 +30,50 @@ export function initialize() {
...
-      const script = await readFile(resolvedScriptPath, {
-        encoding: 'utf-8',
-        flag: 'r',
-      })
-
-      return script
+  ipcMain.handle(ScriptHandler.Open, async (_, scriptPath: string) => {
+    console.log(`${ScriptHandler.Open} event received`)
+
+    const absolute = path.isAbsolute(scriptPath)
```
allansson, line 37: Instead of passing whether the path is absolute or not, we just check it. This should also make it easier to refactor so that k6 Studio treats all paths as absolute which should simplify things further.

src/handlers/script/index.ts
```diff
@@ -30,50 +30,50 @@ export function initialize() {
...
+      : path.join(SCRIPTS_PATH, scriptPath)
+
+    const script = await readFile(resolvedScriptPath, {
+      encoding: 'utf-8',
+      flag: 'r',
+    })
+
+    return {
+      script,
+      isExternal: path.dirname(resolvedScriptPath) !== SCRIPTS_PATH,
```
allansson: The `isExternal` check is done here rather than in the renderer because we don't have access to things like `dirname` there and I don't want to bring in a full polyfill for the `path` module.

src/handlers/script/types.ts
```diff
@@ -11,3 +11,8 @@ export enum ScriptHandler {
   Check = 'script:check',
   RunFromGenerator = 'script:run-from-generator',
 }
+
+export interface OpenScriptResult {
+  script: string
+  isExternal: boolean
+}
```
allansson, line 18: In the implementation of #839, this interface will contain data to determine whether the script is a browser script or not. It could also contain the list of scenarios in the test e.g. to allow users to pick a specific scenario to validate.

src/views/Home/Home.tsx
```diff
@@ -73,11 +97,9 @@ export function Home() {
...
             description="Debug and validate your k6 script"
           >
-            <Button variant="ghost" asChild>
-              <Link to={getRoutePath('validator', {})}>
-                <CircleCheckIcon />
-                Validate script
-              </Link>
+            <Button variant="ghost" onClick={handleOpenScript}>
+              <CircleCheckIcon />
+              Open script
```
allansson, line 89: Sneaking in a small change in terminology here. 😬

src/views/Validator/ValidatorControls.tsx
```diff
@@ -7,8 +7,7 @@ import { useProxyStatus } from '@/hooks/useProxyStatus'
 
 interface ValidatorControlsProps {
   isRunning: boolean
-  isScriptSelected: boolean
```
allansson, line 10: Since we show a file dialog when users open a script from the Home page, they can no longer end up in the validator without a script selected. So the `isScriptSelected` logic no longer applies.

src/views/Validator/ValidatorControls.tsx
```diff
@@ -7,8 +7,7 @@ import { useProxyStatus } from '@/hooks/useProxyStatus'
...
-  isExternal: boolean
+  canDelete: boolean
```
allansson, line 10: Renamed this property to make it clearer what it does.

src/handlers/script/index.ts
```diff
@@ -30,50 +30,50 @@ export function initialize() {
...
+      browserWindow,
+      scriptPath: resolvedScriptPath,
+      proxySettings: k6StudioState.appSettings.proxy,
+      usageReport: k6StudioState.appSettings.telemetry.usageReport,
+    })
+
+    trackEvent({
+      event: UsageEventName.ScriptValidated,
+      payload: {
+        isExternal: path.dirname(resolvedScriptPath) !== SCRIPTS_PATH,
```
allansson: There's a very slight change in tracking here that's _VERY_ unlikely to occur: if you use the "Open script" dialog to open a script in the `SCRIPTS_PATH` it will register as if you selected the script from the sidebar.

src/views/Validator/ValidatorEmptyState.tsx
```diff
@@ -29,17 +25,6 @@ export function ValidatorEmptyState({
     return window.studio.proxy.launchProxy()
   }
 
-  if (!isScriptSelected) {
```
allansson, line 32: There's no longer any way of getting to the validator without having picked a script.

src/views/Validator/Validator.tsx
```diff
@@ -127,22 +113,14 @@ export function Validator() {
...
-    resetProxyData()
-    resetLogs()
-    resetChecks()
-  }, [script, resetProxyData, resetLogs, resetChecks])
-
   return (
     <View
       title="Validator"
-      subTitle={file ? <FileNameHeader file={file} /> : null}
+      subTitle={<FileNameHeader file={file} canRename={!data?.isExternal} />}
```
allansson, line 119: I think it makes sense to show the file name for external scripts as well, so instead we just toggle the ability to rename it.

src/views/Validator/Validator.tsx
```diff
@@ -184,3 +162,9 @@ export function Validator() {
     </View>
   )
 }
+
+export function ValidatorView() {
+  const scriptPath = useScriptPath()
+
+  return <Validator key={scriptPath} scriptPath={scriptPath} />
```
allansson: Using a key to reset state is more reliable than using `useEffect`.

PR #886, opened 2025-11-11
feat(browser): Generate `getByAltText`, `getByLabel`, `getByPlaceholder` and `getByTitle` selectors

### LINKED ISSUE #835: Add support for `getByLabel` locators
### Feature description

Since k6 v1.2.0, the browser module has support for `getByLabel` locators. These locators are a lot more reliable when working with inputs than CSS selectors.

We should add a new `aria` property to all events that include ARIA information. One value in this property would be the label of the target element, if any.

Once we have the data, we can make a decision during codegen on which kind of locator to use.

_No response_

### LINKED ISSUE #836: Add support for `getByPlaceholder` locators
### Feature description

Since k6 v1.2.0, the browser module has support for `getByPlaceholder` locators. Although not as robust as some other `getBy*` locators, they are a lot more reliable when working with inputs than CSS selectors.

We should add a new `aria` property to all events that include ARIA information. One value in this property would be the placeholder of the target element, if any.

Once we have the data, we can make a decision during codegen on which kind of locator to use.

_No response_

### LINKED ISSUE #837: Add support for `getByAltText` locators
### Feature description

Since k6 v1.2.0, the browser module has support for `getByAltText` locators. These can be helpful to find image elements in a more reliable way than e.g. CSS selectors.

We should add a new `aria` property to all events that include ARIA information. One value in this property would be the alt text of the target element, if any.

Once we have the data, we can make a decision during codegen on which kind of locator to use.

_No response_

### LINKED ISSUE #838: Add support for `getByTitle` locators
### Feature description

Since k6 v1.2.0, the browser module has support for `getByTitle` locators. These can be helpful to find elements that rely on the `title` attribute to convey information. While not as strong as some other `getBy*` locators, it is still better than CSS.

We should add a new `aria` property to all events that include ARIA information. One value in this property would be the title of the target element, if any.

Once we have the data, we can make a decision during codegen on which kind of locator to use.

_No response_

allansson:
## Description

This PR adds support for recording and emitting various `getBy` selectors:

* `getByAltText`
* `getByLabel`
* `getByPlaceholder`
* `getByTitle`

## How to Test

Most of these are pretty hard to test IRL.

* `getByAltText`: in 99.99% of cases a `getByRole("img", { name: "" })` will be used. I hacked this by commenting out the `getRoleSelector` function in `getNodeSelector`.
* `getByLabel`: password inputs don't have an aria role, so they are good targets.
* `getByPlaceholder`: I used the search box [in our docs](https://grafana.com/docs/k6/latest/using-k6-browser/options/).
* `getByTitle`: I went to some pages and searched the DOM for the title attribute in the console `document.querySelectorAll('[title]')`.

## Checklist

## Related PR(s)/Issue(s)

Closes #837 
Closes #838 
Closes #835
Closes #836

### DIFF
--- a/src/codegen/browser/__snapshots__/browser/locators/get-by-alt-text.ts
+++ b/src/codegen/browser/__snapshots__/browser/locators/get-by-alt-text.ts
@@ -14,5 +14,5 @@ export const options = {
 export default async function () {
   const page = await browser.newPage();
 
-  await page.getByAltText("Grot is happy").click();
+  await page.getByAltText("Grot is happy", { exact: true }).click();
 }


extension/src/target.ts
```diff
@@ -48,6 +55,116 @@ function generateRoleSelector(
...
+
+  const placeholder = element.placeholder.trim()
+
+  if (placeholder === '') {
+    return undefined
+  }
+
+  const matches = queryAllByPlaceholderText(document.body, placeholder, {
+    exact: true,
+  })
```
allansson, line 131: I'm a bit on the fence about using exact matching. It's the most reliable for generating selectors, but it's not necessarily the most reliable when authoring tests. I think in general you want to be pretty loose so that you don't have to take e.g. alt texts in the accessible name into account.

One way is to first do a loose match and if that doesn't produce a unique result do an exact match. Anyway, just some food for thought.

PR #920, opened 2025-12-02
internal: Improve correlation prompts and tools

e-fisher:
## Description

This PR introduces several improvements to the autocorrelation feature:

- Significantly reduced token usage for large recordings: Up to an ~80% decrease. For example, a grafana.net recording with 116 requests previously used over 300k tokens, and now uses ~50k.
- More consistent prompt behavior: updated prompts reduce the likelihood of generating unnecessary correlation rules.
- Upgraded to GPT-5 for improved correlation accuracy.
- New tools for recording navigation: added utilities to get an overview of a recording and to search within it.
- Value diffing between recording and validation: helps identify dynamic values and better extract them.
- Token usage indicator: displays how many tokens are consumed during autocorrelation.

### Quickpizza
1. Record the flow: **login → order pizza → rate pizza**.
2. Verify that autocorrelation correctly extracts the CSRF token for login and the pizza ID for `/api/ratings`.

### Personal Grafana Stack
In testing, autocorrelation successfully logged into a Grafana stack via grafana.com in roughly 9 out of 10 attempts.  
To reproduce this setup, you’ll need:  
1. The PR that **disables redirect following** (required, as correlation does not work correctly with redirects enabled): [#910](https://github.com/grafana/k6-studio/pull/910)  
2. A personal Grafana stack (see the redirect PR description for setup details).

## Checklist

### DIFF
--- a/src/codegen/codegen.utils.ts
+++ b/src/codegen/codegen.utils.ts
@@ -1,3 +1,5 @@
+import { flow } from 'lodash-es'
+
 import { canonicalHeaderKey } from '@/rules/utils'
 import { ProxyData } from '@/types'
 import { getLocationHeader, getUpgradeHeader } from '@/utils/headers'
@@ -122,7 +124,7 @@ export const removeWebsocketRequests = (recording: ProxyData[]) => {
 }
 
 export function cleanupRecording(recording: ProxyData[]) {
-  return removeWebsocketRequests(recording)
+  return flow(removeWebsocketRequests, mergeRedirects)(recording)
 }
 
 export function shouldIncludeHeaderInScript(key: string) {
--- a/src/handlers/ai/index.ts
+++ b/src/handlers/ai/index.ts
@@ -1,7 +1,6 @@
 import { OpenAIResponsesProviderOptions } from '@ai-sdk/openai'
 import { convertToModelMessages, streamText } from 'ai'
 import { ipcMain, IpcMainEvent } from 'electron'
-import log from 'electron-log/main'
 
 import { setupAiModel } from './model'
 import { streamMessages } from './streamMessages'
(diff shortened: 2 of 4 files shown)


src/utils/json.ts
```diff
@@ -4,7 +4,6 @@ export function safeJsonParse<T extends object>(value: string) {
   try {
     return JSON.parse(value) as T
   } catch (error) {
-    console.error('Failed to parse JSON', error)
```
e-fisher, line 7: Reduce noise

going-confetti: Could you remind me why it's so noisy? Is that related to the proxy recorder or is it because some websites have invalid JSON?

src/views/Generator/AutoCorrelation/TokenUsageIndicator.tsx
```diff
@@ -0,0 +1,23 @@
...
+  tokenUsage,
+}: {
+  tokenUsage?: TokenUsage
+}) {
+  if (tokenUsage?.totalTokens === undefined) {
+    return null
+  }
+
+  const formatted = Intl.NumberFormat('en', { notation: 'compact' }).format(
+    tokenUsage.totalTokens
```
going-confetti, line 15: I'm not super familiar with how other tools handle this, but would it make sense to report the input/output token numbers considering the different price-per-token?

src/utils/json.ts
e-fisher: We had some issues with APIs that returned JSON but didn't have correct content-type headers set, breaking response body targeted rules. The solution was to "try" parsing content as JSON without relying on headers, console log here would trigger every time the body is not JSON.

src/views/Generator/AutoCorrelation/TokenUsageIndicator.tsx
e-fisher: Agree, will add this in a separate PR

PR #925, opened 2025-12-08
chore: Fix shim import path generation on Windows

### LINKED ISSUE #923: Instrumentation tests are failing in windows OS due to path separators in browser shim imports
The test suite is this "should inject browser options when browser module is imported"
This file: instrumentation.test.ts

The error
```
+ import browser from "\\mock\\resources\\shims\\browser.js";
  export default function () {}
  export function handleSummary() {
    console.log("This is the handleSummary shim");
  }
  export const options = {

 ❯ src/main/runner/instrumentation.test.ts:200:5
    198|     const result = await instrumentScript({ script, shim
    199| 
    200|     await expect(result).toMatchFileSnapshot(
       |     ^
    201|       './__snapshots__/script/browser-options/with-brows
    202|     )
```

This is a simple fix for the windows path for forward slashes(/). I have already pushed  

## Bug Description
The instrumentation tests were failing on Windows because the browser shim import path was using backslashes (`\`) instead of forward slashes (`/`). This caused test snapshots to not match between Windows and Unix-based systems.

## Root Cause
The `getShimPath()` function in `src/main/runner/instrumentation.ts` was using `path.join()` which returns platform-specific path separators:
- Windows: `C:\mock\resources\shims\browser.js`
- Linux/Mac: `/mock/resources/shims/browser.js`

Since JavaScript import statements require forward slashes regardless of platform, the generated imports were inconsistent.

1. use a windows 10 laptop 
3. git clone this repo
4. Install the dependencies
5. Run `npm test`
6. Observe the logs

### Expected behavior

The tests for "should inject browser options when browser module is imported" should pass on windows OS

### Actual behavior

The test suite "should inject browser options when browser module is imported" is failing because the path separator in the import statement is using backslashes (\) on Windows instead of forward slashes (/).

### Grafana k6 Studio version

v1.9.0

### OS

Windows 10

mohswell:
## Description
Windows 10 test error for istrumentation and also my machine was tracking the k6.exe file so I changed the gitignore file to use the windows .exe extension

So JavaScript import statements require forward slashes regardless of OS, causing Windows test snapshots to mismatch.
I just normalized all path separators to forward slashes by adding replace() functions. Forward slashes work correctly on all platforms (Windows, Linux, Mac).

image

## How to Test
npm test -- instrumentation.test.ts

## Checklist

## Related PR(s)/Issue(s)
https://github.com/grafana/k6-studio/issues/923

 
 Resolves #923

### DIFF
--- a/src/main/runner/instrumentation.ts
+++ b/src/main/runner/instrumentation.ts
@@ -160,8 +160,7 @@ const getShimPath = (name: string) => {
   // @ts-expect-error We are targeting CommonJS so import.meta is not available
   if (!import.meta.env.PROD) {
     // Module specifiers are URL-like and must always use forward slashes (/) regardless of OS
-    return path.posix.
-      .join(app.getAppPath(), 'resources', 'shims', name)
+    return path.posix.join(app.getAppPath(), 'resources', 'shims', name)
   }
 
   return path.posix.join(process.resourcesPath, 'shims', name)


src/main/runner/instrumentation.ts
```diff
@@ -159,10 +159,12 @@ const getSnippetPath = (snippetName: string) => {
 const getShimPath = (name: string) => {
   // @ts-expect-error We are targeting CommonJS so import.meta is not available
   if (!import.meta.env.PROD) {
-    return path.join(app.getAppPath(), 'resources', 'shims', name)
+    return path
+      .join(app.getAppPath(), 'resources', 'shims', name)
+      .replace(/\\/g, '/')
```
going-confetti: May I suggest a different approach? By using [path.posix](https://nodejs.org/api/path.html#pathposix), we could make it look more intentional:

```suggestion
    // Module specifiers are URL-like and must always use forward slashes (/) regardless of OS
    return path.posix.
      .join(app.getAppPath(), 'resources', 'shims', name)
```

Issue #473, opened 2025-02-12
Add confirmation prompts for destructive actions


going-confetti:
* Deleting generators/recordings/scripts

PR #928, opened 2025-12-08
feat: add dialog confirmation for destructive actions

mohswell:
## Description
This adds a confirmation dialog before deleting a file, using the existing Radix dialog components.

image

## Checklist

## Related PR(s)/Issue(s)
Solves this https://github.com/grafana/k6-studio/issues/473


e-fisher (changes requested):
Thank you for your contribution @mohswell! While this solution works, I think the solution could be improved:

## Code
The whole `wrappedTrigger` and element cloning logic is a bit hacky, this can be simplified:

1. Add a `preventDefault` to `DropdownMenuItem` when it's wrapped in dialog:

```ts
<DeleteFileDialog
  key={item.label}
  file={file}
  onConfirm={item.onClick}
  trigger={
    <DropdownMenu.Item
      color="red"
      onClick={(e) => e.preventDefault()} // <-
    >
      {item.label}
    </DropdownMenu.Item>
  }
/>
```
This will prevent context menu from closing `onClick`.

2. Add `onClick` event to the trigger in `DeleteFileDialog`:

```ts
return (
  <AlertDialog.Root open={open} onOpenChange={handleOpenChange}>
    <AlertDialog.Trigger onClick={() => handleOpenChange(true)}>
      {trigger}
    </AlertDialog.Trigger>
    <AlertDialog.Content size="2" maxWidth="480px">
```  

Now you don't need to create `wrappedTrigger`

3. Revert changes to `useFileContextMenuItems` and keep delete action as `onClick`

```ts
return [
  { label: 'Rename', onClick: onRename },
  { label: 'Open containing folder', onClick: handleOpenFolder },
  { label: 'Delete', onClick: handleDelete, destructive: true },
]
```

4. Pass the `onClick` value as `onConfirm` prop to the `DeleteFileDialog`:

```ts
  item.destructive ? (
    <DeleteFileDialog
      key={item.label}
      file={file}
      onConfirm={item.onClick} // <-
      trigger={
        <DropdownMenu.Item
          color="red"
          onClick={(e) => e.preventDefault()}
        >
          {item.label}
        </DropdownMenu.Item>
      }
    />
```

5. Remove all the file deletion logic from `DeleteFileDialog` and simply call `onConfirm` when confirmation button is clicked:

```ts
<AlertDialog.Action>
  <Button variant="solid" color="red" onClick={onConfirm}>
    {actionLabel}
  </Button>
</AlertDialog.Action>
```

6. Refactor to remove duplication between `FileActionsMenu` and `FileContextMenu`, and to avoid ternary conditions inside JSX

Extract `MenuItem` component

```ts
function MenuItem({ destructive, label, file, onClick }) {
  if (destructive) {
    return (
      <DeleteFileDialog
        key={label}
        file={file}
        onConfirm={onClick}
        trigger={
          <DropdownMenu.Item color="red" onClick={(e) => e.preventDefault()}>
            {label}
          </DropdownMenu.Item>
        }
      />
    )
  }

  return (
    <DropdownMenu.Item key={label} onClick={onClick}>
      {label}
    </DropdownMenu.Item>
  )
}
```

7. Now that `DeleteFileDialog` is decoupled from any file deletion logic it can be made more generic, like `DestructiveActionDialog` and read title/description/action from props, but that's optional


I did like your addition of toasts, though! This could be added to the `handleDelete` function in `useFileContextMenuItems`.

## UI

The label "file" and extension badge don't add much value to the confirmation dialog and don't look great spanning multiple lines. I'd go with something simple like this:

Image

src/views/Validator/ValidatorControls.tsx
```diff
@@ -1,31 +1,46 @@
...
   onRunInCloud,
   onSelectScript,
   onStopScript,
+  onAfterDelete,
 }: ValidatorControlsProps) {
   const proxyStatus = useProxyStatus()
 
+  const handleDelete = useDeleteFile({
+    file,
+    navigateHomeOnDelete: false,
```
e-fisher: You are disabling navigation here, but pass a `onAfterDelete` prop that does exactly that. Perhaps `onAfterDelete` could be removed?

src/views/Validator/ValidatorControls.tsx
```diff
@@ -67,9 +82,22 @@ export function ValidatorControls({
...
+                    cursor: 'pointer',
+                    padding: '4px 6px',
+                  }}
+                  onClick={(e) => e.preventDefault()}
+                >
+                  Delete
+                </div>
+              }
+            />
           )}
```
e-fisher, line 94: The styling doesn't match `DropdownMenu`:

Image

Any reason not to do it the same way as in sidebar and pass `DropdownMenu.Item` with `e.preventDefault` default?

src/views/RecordingPreviewer/RecordingPreviewerControls.tsx
```diff
@@ -38,15 +40,12 @@ export function RecordingPreviewControls({
...
-  const handleDeleteRecording = async () => {
-    await window.studio.ui.deleteFile(file)
-    navigate(getRoutePath('home'))
-  }
+  const navigateRecorder = () => navigate(getRoutePath('recorder'))
 
-  const handleDiscard = async () => {
-    await window.studio.ui.deleteFile(file)
-    navigate(getRoutePath('recorder'))
-  }
```
e-fisher, line 49: While the Discard and Delete buttons currently do almost the same thing (and this will be revisited in the future), I don’t think changes in this PR should affect the redirect logic. It’s common to start a new recording after discarding, so Discard should take the user to the `recorder` path, while Delete should redirect to `home`.

src/hooks/useDeleteFile.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    expect(navigate).toHaveBeenCalledWith('/home')
+  })
+
+  it('should show error toast if deletion fails', async () => {
+    // So there is a try catch from my useDeleteFile hook
+    // and the expected error is printed via console.error inside the hook. So I am testing this bit
+    // Vitest prints it to stderr even if the test passes,
+    // because the hook calls console.error during the simulated failure.
+    // Suppressing this with spyOn
```
e-fisher: Verifying that toast is shown should be enough, I wouldn't tie this test to a console log.

mohswell: I have removed the console.error from the hook

src/views/Validator/ValidatorControls.tsx
mohswell: I have fixed, forgot i wrapped a div because DropdownMenu.Item automatically closes the menu on selection

src/views/RecordingPreviewer/RecordingPreviewerControls.tsx
mohswell: Yes they do the same thing, I decided to combine the two but now have separated the 2 redirections

src/views/Validator/ValidatorControls.tsx
mohswell: yes i have set `navigateHomeOnDelete: true` and let the hook handle navigation directly and removed this onAfterDelete prop

e-fisher (commented):
@mohswell looking good, one last thing before approving this: file type should be capitalized in the toast, taking screenshot as example it should be "Generator deleted". 

Image

e-fisher:
@mohswell looks good, just need to update the title in test as well

e-fisher (approved):
LGTM 🚀

mohswell:
Force pushed this branch with latest changes from main and my changes.

Issue #922, opened 2025-12-04
External script validation failing due to `execution` missing


Llandy3d:
### Describe the bug

Opening an external script in the Validator fails to run due to the lack of the `execution` import.

### Steps to reproduce the problem

- Create a valid k6 script, ex.
```
import http from 'k6/http';
import { sleep } from 'k6';

export const options = {
  iterations: 10,
};

export default function () {
  http.get('https://quickpizza.grafana.com');
  sleep(1);
}
```

- Open the validator and select this script in your system
- Press the validate button and check the error in logs

```
ReferenceError: execution is not defined
	at instrumentArguments (file:///Users/nobody/dev/.ir6ql5__tmp-k6studio__.js:41:41(16))
	at request (file:///Users/nobody/dev/.ir6ql5__tmp-k6studio__.js:8:61(12))
	at get (file:///Users/nobody/dev/.ir6ql5__tmp-k6studio__.js:17:26(12))
	at bound get (native)
	at default (file:///Users/nobody/dev/.ir6ql5__tmp-k6studio__.js:67:11(4))
```

### Expected behavior

Should be able to validate an external script without errors.

### Actual behavior

Not able to validate an external script.

### Grafana k6 Studio version

master

### OS

macOS 15.3.1

mohswell:
Hi @Llandy3d can you maybe assign me this. I can fix this in a pr, I have experienced it in my windows 11....I think its the resources file that has group snippets.

The injected group shim was using execution.vu.tags.group without importing k6/execution for external runs. that global wasn’t defined, so the shim threw ReferenceError: execution is not defined before the scripts could run. 

i was reading grafana k6 docs: https://grafana.com/docs/k6/latest/javascript-api/k6-execution/#vu

PR #946, opened 2025-12-17
fix: Scripts without k6/execution import cannot be run in Validator

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/922

## Description
- Add an entry script, that imports the user's script and use instrumented versions of the http and browser modules for validation/debug run
- Remove `group_snippet.js` and `checks_snippet.js`

grafik

## How to Test
Verify that scripts can be run both in Validator itself and in the Validator dialog in Generator. All these should work as expected:
- http scripts with- and without checks
- http script, that doesn't have an explicit http import (e.g. [this httpx example](https://grafana.com/docs/k6/latest/javascript-api/jslib/httpx/#example))
- browser scripts
- Any script that doesn't have the `k6/execution` import (the httpx example should cover this case)

## Checklist

### DIFF
--- a/src/main/runner/entrypoint.ts
+++ b/src/main/runner/entrypoint.ts
@@ -14,7 +14,7 @@ const DEFAULT_SCENARIOS: Options['scenarios'] = {
   },
 }
 
-const scenarios: Options['scenarios'] = Object.fromEntries(
+const scenarios: Required<Options>['scenarios'] = Object.fromEntries(
   Object.entries(
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     (userScript.options as Options)?.scenarios ?? DEFAULT_SCENARIOS
@@ -41,12 +41,9 @@ export const options: Options = {
 export default async function () {
   // Always prefer the 'default' export if configured, otherwise fallback to the
   // first scenario.
-  const exec = Object.values(scenarios ?? {}).find(
-    ({ exec }) => exec === 'default'
-  )
+  const exec = Object.values(scenarios).find(({ exec }) => exec === 'default')
     ? 'default'
-    : Object.values(scenarios ?? {})[0]?.exec
-
+    : Object.values(scenarios)[0]?.exec
   if (exec === undefined) {
     throw new Error('No scenario found to execute')
   }
--- a/src/main/runner/shims/browser/index.ts
+++ b/src/main/runner/shims/browser/index.ts
@@ -21,7 +21,6 @@ const locatorProxy = (
 export const browser = createProxy(nativeBrowser, {
   newPage: {
     proxy: (_target) => {
-      console.log('Shimmed!')
       return {
         goto: {
           track(url) {


vite.script.config.ts
going-confetti, line 1: This is copied from the browser debugger PR with the addition of an the http shim input

src/main/runner/entrypoint.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// @ts-expect-error - Path will be replaced at runtime
+// eslint-disable-next-line import/no-unresolved
+import * as userScript from '__USER_SCRIPT_PATH__'
+
+export default function () {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+  userScript['default']()
+}
+
```
going-confetti, line 53: @allansson I made two changes to simplify the implementation:
1. Only the default scenario is supported since we don't have a way to select other scenarios from the UI
2. The `options` export is missing in the entry script, but is added in `intrumentation.ts`. We also don't merge the user's option with the default debug options, but that's in line with the current behavior.

src/main/runner/entrypoint.ts
```diff
@@ -0,0 +1,14 @@
+// Import shims first to mutate k6 modules globally
+import './shims/http'
+import './shims/browser'
```
allansson, line 3: Since we no longer replace the `k6/browser` imports in the user's script, the browser shim needs to modify the `browser` export of `k6/browser`.

I did a quick test, and modifying it seems to work:

```ts
// ./shim.js
import { browser } from 'k6/browser'

const newPageOriginal = browser.newPage

browser.newPage = async function () {
  console.log('Shimmed!')

  return newPageOriginal.apply(this, arguments)
}

// ./test.js
import './shim.js'
import { browser } from 'k6/browser'

export const options = {
  scenarios: {
    default: {
      executor: 'shared-iterations',
      options: { browser: { type: 'chromium' } },
    },
  },
}

export default async function () {
  const page = await browser.newPage()

  await page.goto('https://k6.io/')
}
```

This logs "Shimmed!" for me.

But since the browser debugger PR has pretty significant changes to the browser shim, I think it might be better to get this merged and do the changes there instead. WDYT?

src/main/runner/entrypoint.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// @ts-expect-error - Path will be replaced at runtime
+// eslint-disable-next-line import/no-unresolved
+import * as userScript from '__USER_SCRIPT_PATH__'
+
+export default function () {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+  userScript['default']()
```
allansson: We should `await` this just in case.

src/main/runner/entrypoint.ts
allansson: > Only the default scenario is supported since we don't have a way to select other scenarios from the UI

Makes sense. I think adding a scenario selector in the UI would be pretty cool but it shouldn't be a part of this.

However, the current solution _requires_ that your test is using a default export. I'm thinking we could do something like this:

```ts
// The user might not have exported options or used the `iteration`, `duration` or `stages`
// properties. In these cases we want to fallback to using the `default` function.
const DEFAULT_SCENARIOS = {
  "default": {
    exec: "default"
  }
}

const scenarios = Object.values(userScript.options?.scenarios ?? DEFAULT_SCENARIOS)
  .map(scenario => scenario.exec ?? "default")

// Always prefer the 'default' export if configured, otherwise fallback to the
// first scenario.
const exec = scenarios.includes("default") ? "default" : scenarios[0]

// Alternatively we throw if there are multiple scenarios:
if (scenarios.length > 1) {
  throw new Error("Multi-scenario tests are currently not supported in k6 Studio")
}

await userScript[exec]()
```

allansson: > The options export is missing in the entry script, but is added in intrumentation.ts. We also don't merge the user's option with the default debug options, but that's in line with the current behavior.

One really nice thing about this entry script approach is that we don't need to rely on static analysis and can do this more reliably at runtime with little effort:

```ts
export const options = {
  scenarios: Object.fromEntries(
    Object.entries(userScript.options?.scenarios ?? {}).map(([name, scenario]) => {
      return {
        // Always use 1 VU, 1 iteration in the debugger
        executor: "per-vu-iterations",
        vus: 1,
        iteration: 1,
        exec: scenario.exec,
        // Make sure the browser options are carried over
        options: scenario.options
      }
    })
  )
}
```

It means we can avoid parsing the user's script altogether.

src/main/runner/summary.ts
```diff
@@ -1,7 +1,16 @@
...
+  checks: Check[]
+  groups: Group[]
+}
+
+export function handleSummary(data: {
+  root_group: { checks: Check[]; groups: Group[] }
+}) {
+  const checks: Check[] = []
+
+  function traverseGroup(group: Group) {
```
allansson, line 11: It won't let me comment on unchanged lines, but below the `traverseGroup` function:

```ts
data.root_group.checks.forEach((check) => {
  checks.push(check)
})

data.root_group.groups.forEach((group) => {
  traverseGroup(group)
})

// =>

traverseGroup(data.root_group)
```

src/main/runner/shims/http/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// Mutate the native http module to add instrumentation
+http.request = function <RT extends http.ResponseType | undefined>(
+  method: string,
+  url: string | HttpURL,
+  body?: http.RequestBody | null,
+  params?: http.RefinedParams<RT> | null
+) {
+  return originalRequest<RT>(method, url, body, instrumentParams(params))
+}
```
allansson: Typescript is able to infer these types correctly for me.

```suggestion
http.request = function (
  method
  url,
  body?,
  params?
) {
  return originalRequest(method, url, body, instrumentParams(params))
}
```

src/main/runner/shims/http/utils.ts
```diff
@@ -0,0 +1,18 @@
+import execution from 'k6/execution'
+import type { RefinedParams, ResponseType } from 'k6/http'
+
+export function instrumentParams<RT extends ResponseType | undefined>(
+  params: RefinedParams<RT> | null = {}
+) {
+  // Trim and remove prefix
+  const group = `${execution.vu.metrics.tags.group}`.trim().replace(/^::/, '')
```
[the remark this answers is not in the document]
allansson: No need to support <v0.38.0

src/main/runner/shims/http/index.ts
going-confetti: Indeed, I must've changes something because I had issues before. Glad you noticed this, now it looks cleaner 👍

src/main/runner/entrypoint.ts
going-confetti: >But since the browser debugger PR has pretty significant changes to the browser shim, I think it might be better to get this merged and do the changes there instead. WDYT?

Sounds reasonable!

src/main/runner/entrypoint.ts
going-confetti: Done! Added both changes

src/main/runner/shims/browser/index.ts
```diff
@@ -21,6 +21,7 @@ const locatorProxy = (
 export const browser = createProxy(nativeBrowser, {
   newPage: {
     proxy: (_target) => {
+      console.log('Shimmed!')
```
allansson: Mistake?

src/main/runner/entrypoint.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { Options } from 'k6/options'
+
+const DEFAULT_SCENARIOS: Options['scenarios'] = {
+  default: {
+    executor: 'per-vu-iterations',
+    exec: 'default',
+  },
+}
+
+const scenarios: Options['scenarios'] = Object.fromEntries(
```
allansson: Nit: technically more correct and lets us remove a couple of `??`.

```suggestion
const scenarios: Required<Options>['scenarios'] = Object.fromEntries(
```

allansson (approved):
Two minor things.

PR #949, opened 2025-12-18
internal: Sum token usage and show input/output separately

e-fisher:
## Description

Token usage was being replaced instead of accumulated. With `sendAutomaticallyWhen: lastMessageIsToolCall`, `useChat` makes multiple API calls (one per tool call round), and each reports its own usage. What we displayed was "context usage" but not total tokens used.

- Show accumulated token usage
- Show input and output token usage separately because they have different cost

## How to Test

I've created a new project to verify token usage:

screencapture 2025-12-18 at 12 40 58@2x

screencapture 2025-12-18 at 12 40 17@2x

Token usage displayed in k6-studio now matches the one displayed in OpenAI usage dashboard

I don't think you need to go through the same steps, just verify there are no regressions and token usage is displayed as expected.

## Checklist

### DIFF
--- a/src/views/Generator/AutoCorrelation/utils/sumTokenUsage.ts
+++ b/src/views/Generator/AutoCorrelation/utils/sumTokenUsage.ts
@@ -1,0 +1,16 @@
+import { TokenUsage } from '@/handlers/ai/types'
+
+export function sumTokenUsage(
+  current: TokenUsage | undefined,
+  next: TokenUsage
+): TokenUsage {
+  return {
+    inputTokens: (current?.inputTokens ?? 0) + (next.inputTokens ?? 0),
+    outputTokens: (current?.outputTokens ?? 0) + (next.outputTokens ?? 0),
+    totalTokens: (current?.totalTokens ?? 0) + (next.totalTokens ?? 0),
+    cachedInputTokens:
+      (current?.cachedInputTokens ?? 0) + (next.cachedInputTokens ?? 0),
+    reasoningTokens:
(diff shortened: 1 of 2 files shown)


src/views/Generator/AutoCorrelation/utils/sumTokenUsage.ts
```diff
@@ -0,0 +1,16 @@
...
+): TokenUsage {
+  return {
+    inputTokens: (current?.inputTokens ?? 0) + (next.inputTokens ?? 0),
+    outputTokens: (current?.outputTokens ?? 0) + (next.outputTokens ?? 0),
+    totalTokens: (current?.totalTokens ?? 0) + (next.totalTokens ?? 0),
+    cachedInputTokens:
+      (current?.cachedInputTokens ?? 0) + (next.cachedInputTokens ?? 0),
+    reasoningTokens:
+      (current?.reasoningTokens ?? 0) + (next.reasoningTokens ?? 0),
+  }
```
going-confetti, line 15: Maybe something like this to avoid repetition?
```ts
  const result: TokenUsage = {
    inputTokens: 0,
    outputTokens: 0,
    totalTokens: 0,
    cachedInputTokens: 0,
    reasoningTokens: 0,
  }

  for (const key in result) {
    const k = key as keyof TokenUsage
    result[k] = (current?.[k] ?? 0) + (next[k] ?? 0)
  }
```

e-fisher: Thought about this, but went with the simplest approach, although with some duplication

Issue #848, opened 2025-09-19
Parameterization/Correlation rule replaces all occurences of matched value


going-confetti:
### Describe the bug

To put it simply, the replacement logic consists of two steps:
1. Match the value user the selector
2. use `replaceAll` to replace the occurences of this value

This means that the actual replacement doesn't take the selector configuration into account: if `admin` is matched using the begin-end selector, the same substring will be replaced in other places (where provided begin-end values may not be present)

<img width="962" height="456" alt="Image" src="https://github.com/user-attachments/assets/8fe6eed2-31d4-423b-844f-91656c49511c" />

### Steps to reproduce the problem

1. Start a recording
2. Navigate to https://quickpizza.grafana.com/admin
3. Log in
4. Stop the recording and create a test generator
5. Create a parameterization rule, that replaces the username ("admin") with e.g. an arbitrary text value:
    - Select `URL` as `Target`, `Begin-End` as `Type` 
    - Type `user=` into `Begin` and `&` into `End`
    - Type anything into the `Value` field, e.g. `HELLO`
6. Find the affected request in the request list

### Expected behavior

The URL path should be `/api/admin/login?user=HELLO&password=admin`

### Actual behavior

The URL path is `/api/HELLO/login?user=HELLO&password=HELLO`

### Grafana k6 Studio version

1.7.0

### OS

-

PR #966, opened 2025-12-29
fix: Rules with regex and begin-end selectors replace all occurences of matched value

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/848

## Description

> [!NOTE]  
> Empty regex expressions/expressions with empty groups are supported as well as begin-end pairs that match empty strings. This is an intentional change, but do let me know if there's something I missed when making it.

- Change the replacement logic used in rule preview to fix the issue
- Add tests for this edge case
- Refactor the code across `rules/correlation.utils.ts`, `rules/shared.ts`, and `rules/selectors/text.ts`

grafik

## How to Test

> [!IMPORTANT]  
> Make sure to test this thoroughly and let me know if you think something should be covered by tests

1. Verify that https://github.com/grafana/k6-studio/issues/848 is addressed (also check that the generated test works as expected
2. Verify that the following functionality works as expected:
    - Rule preview for correlation and parameterization rules
    - Correlation rules with- and without custom replacers
    - All types of selectors (begin-end, regex, header-name, JSON, text) in all kinds of targets (URL, headers, body)

## Checklist

### DIFF
--- a/src/rules/selectors/beginEnd.ts
+++ b/src/rules/selectors/beginEnd.ts
@@ -7,15 +7,15 @@ import { exhaustive } from '@/utils/typescript'
 export function replaceBeginEnd(
   selector: BeginEndSelector,
   request: Request,
-  variableName: string
+  newValue: string
 ): Request {
   switch (selector.from) {
     case 'body':
-      return replaceBeginEndBody(selector, request, variableName)
+      return replaceBeginEndBody(selector, request, newValue)
     case 'headers':
-      return replaceBeginEndHeaders(selector, request, variableName)
+      return replaceBeginEndHeaders(selector, request, newValue)
     case 'url':
-      return replaceBeginEndUrl(selector, request, variableName)
+      return replaceBeginEndUrl(selector, request, newValue)
     default:
       return exhaustive(selector.from)
   }
@@ -46,7 +46,7 @@ function replaceBeginEndPattern(
 function replaceBeginEndBody(
   selector: BeginEndSelector,
   request: Request,
-  variableName: string
+  newValue: string
 ): Request {
   const match = matchBeginEnd(
     request.content ?? '',
@@ -59,7 +59,7 @@ function replaceBeginEndBody(
     request.content ?? '',
     selector.begin,
     selector.end,
-    variableName
+    newValue
   )
   return { ...request, content }
 }
@@ -67,7 +67,7 @@ function replaceBeginEndBody(
 function replaceBeginEndHeaders(
   selector: BeginEndSelector,
   request: Request,
-  variableName: string
+  newValue: string
 ): Request {
   for (const [key, value] of request.headers) {
     const match = matchBeginEnd(value, selector.begin, selector.end)
@@ -78,7 +78,7 @@ function replaceBeginEndHeaders(
       value,
       selector.begin,
       selector.end,
-      variableName
+      newValue
     )
 
     const headers: Header[] = request.headers.map(([k, v]) =>
@@ -94,7 +94,7 @@ function replaceBeginEndHeaders(
 function replaceBeginEndUrl(
   selector: BeginEndSelector,
   request: Request,
-  variableName: string
+  newValue: string
 ): Request {
   const match = matchBeginEnd(request.url, selector.begin, selector.end)
   if (match === undefined) return request
@@ -103,19 +103,19 @@ function replaceBeginEndUrl(
     request.url,
     selector.begin,
     selector.end,
-    variableName
+    newValue
   )
   const path = replaceBeginEndPattern(
     request.path,
     selector.begin,
     selector.end,
-    variableName
+    newValue
   )
   const host = replaceBeginEndPattern(
     request.host,
     selector.begin,
     selector.end,
-    variableName
+    newValue
   )
   return { ...request, url, path, host }
 }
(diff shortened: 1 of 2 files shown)


src/rules/correlation.utils.ts
```diff
@@ -35,31 +37,19 @@ export function replaceCorrelatedValues({
...
-    extractedValue,
-    variableName
-  )
-  const cookies: Cookie[] = replaceCookies(
-    request.cookies,
-    extractedValue,
-    variableName
-  )
+  oldValue: string,
+  newValue: string
```
going-confetti, line 41: Should I have kept `variableName` here? I figured it wasn't important in this context, but I'm not sure anymore.

src/rules/shared.ts
```diff
@@ -97,51 +97,39 @@ export const getJsonObjectFromPath = (json: string, path: string) => {
...
-  headers: Header[],
-  value: string,
-  newValue: string
-): Header[] => {
-  return headers.map(([key, headerValue]) => {
-    const replacedValue = headerValue.replaceAll(value, newValue)
-    return [key, replacedValue]
-  })
+  const regex = new RegExp(`${escapeRegExp(begin)}(.*?)${escapeRegExp(end)}`)
+  return content.replace(regex, `${begin}${newValue}${end}`)
```
going-confetti: This is enough to replace the first occurrence of the extracted value surrounded within the begin-end boundaries.

src/rules/shared.ts
```diff
@@ -97,51 +97,39 @@ export const getJsonObjectFromPath = (json: string, path: string) => {
...
+    // If no capture group, replace the entire match
+    return newValue
   })
 }
 
-export const replaceBeginEndBody = (
+const replaceBeginEndBody = (
   selector: BeginEndSelector,
   request: Request,
   variableName: string
```
going-confetti: Here, `variableName` does not look like the correct param name (it only does when this is used for correlation). This is out of scope of this PR, but since I've already made so many changes here, should I rename it to `value`?

src/rules/shared.ts
```diff
@@ -284,318 +307,77 @@ const replaceHeaderByName = (
...
-      scheme: 'http',
-      host: 'localhost:3000',
-      content,
-      path: '/api/v1/foo',
-      timestampStart: 0,
-      timestampEnd: 0,
-      contentLength: 0,
-      httpVersion: '1.1',
-    }
+export function replaceAllBody(
```
going-confetti: `replaceAllBody`, `replaceAllUrl`, and others are basically replacements for the old `replaceContent` `replaceURL`, etc. with a few minor changes (such as their names being more explicit about what the function does (although at this point I'd prefer all names to follow the "replace X in Y pattern", but that's a separate discussion), and the return type being `Request`).

src/rules/shared.ts
```diff
@@ -97,155 +98,189 @@ export const getJsonObjectFromPath = (json: string, path: string) => {
...
+const replaceRegexUrl = (
   selector: RegexSelector,
   request: Request,
   variableName: string
-) => {
-  const valueToReplace = matchRegex(request.url, selector.regex)
-  if (!valueToReplace) return
+): Request | undefined => {
+  const match = matchRegex(request.url, selector.regex)
+  if (match === undefined) return
```
going-confetti: IMO it's perfectly valid to want to use an empty group in some cases. Off the top of my head, I can think of a few examples, such as:
- Add a new query param
- Modify the URL to include prefix/suffix depending on the environment, etc.
- Parameterize emails for plus-addressing (e.g. `()@`)

Image

Image

The downside is that it now matches the beginning of the string (see the screenshot below). I'd argue it's actually good because it provides immediate feedback (whereas before if you had an empty regex, the rule would not work) - but I realize that I might be biased.

Image

src/rules/correlation.utils.ts
e-fisher: Old/new makes more sense in this context 👍

src/rules/shared.ts
e-fisher: I think empty regex input shouldn't produce any matches - it just gives an impression of a bug to me. Do you know any examples where this would be useful?

going-confetti: Yeah, the more I think about it, the more it looks like a bug. I fixed this 👍

src/rules/selectors/beginEnd.ts
```diff
@@ -0,0 +1,121 @@
...
+    case 'headers':
+      return replaceBeginEndHeaders(selector, request, newValue)
+    case 'url':
+      return replaceBeginEndUrl(selector, request, newValue)
+    default:
+      return exhaustive(selector.from)
+  }
+}
+
+export function matchBeginEnd(value: string, begin: string, end: string) {
```
going-confetti, line 24: Petition to rename all these `matchX` functions to `extractX` 👀

src/rules/selectors/beginEnd.ts
going-confetti, line 1: Oops, splitting `shared.ts` into multiple files made it harder to review the changes, I apologize.

src/views/Generator/RuleEditor/SelectorField.tsx
```diff
@@ -187,7 +187,17 @@ function SelectorContent({
...
     case 'regex':
       return (
-        <FieldGroup name={`${field}.regex`} errors={errors} label="Regex">
+        <FieldGroup
+          name={`${field}.regex`}
+          errors={errors}
+          label="Regex"
+          hint={
+            <Text size="1">
+              First match only. Use a capture group (e.g.,{' '}
```
going-confetti, line 196: Image

PR #985, opened 2026-01-08
fix: close websocket server when browser recording exits

### LINKED ISSUE #984: When the browser exits, the WebSocket server is not disposed
### Describe the bug

In the process of a recording with browser events, closing the browser window doesn’t dispose the WebSocket server that listens on port 7554. 
So when you start a new recording with browser events enabled, it tries to start another server on the same port and it fails, producing this error:
This causes the user to experience this error in the studio app:
Image

Actual error is:
```
browser:stop event received
browser:start event received
An unexpected error occurred while starting recording:  BrowserLaunchError: WebSocket server transport error
    at launchBrowserWithExtension (C:\Users\mohswell\Desktop\projects\k6-studio\.vite\build\main.js:35750:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async C:\Users\muhammad_solutech\Downloads\k6-studio\.vite\build\main.js:35891:49
    at async Session.<anonymous> (node:electron/js2c/browser_init:2:107024) {
  source: 'websocket-server-error',
  cause: WebSocketServerError: WebSocket server transport error
      at WebSocketServer.<anonymous> (C:\Users\mohswell\Desktop\projects\k6-studio\.vite\build\main.js:28157:16)
      at WebSocketServer.emit (node:events:518:28)
      at Server.emit (node:events:518:28)
      at emitErrorNT (node:net:1976:8)
      at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
    [errors]: [ [Error] ]
  }
}
```

In the extension-based recorder, the server isn’t stopped on browser process exit. The electron process is still bound to the server port, I ran this pwsh command after i closed the headless browser:
```
Get-NetTCPConnection -State Listen | Where-Object { $_.LocalPort -in 7554,9222 } | ForEach-Object { $p = Get-Process -Id $_.OwningProcess -ErrorAction SilentlyContinue; [PSCustomObject]@{ LocalAddress=$_.LocalAddress; LocalPort=$_.LocalPort; State=$_.State; PID=$_.OwningProcess; ProcessName=$p.ProcessName } } | Sort-Object LocalPort | Format-Table -Auto
````

this is the result:
```
LocalAddress LocalPort  State   PID ProcessName
------------ ---------  -----   --- -----------
::1               7554 Listen 13912 electron
```

1. Launch k6 studio app
2. Tap on recordings, then tap on + 
3. Enter url and start recording
4. Exit the headless chromium browser by clicking the top right icon `x`
5. Now attempt to start a new recording
6. Observe

### Expected behavior

Dispose the server when the browser process exits or errors. While there is a workaround for running taskkill for the PID thats just temporary. 

### Actual behavior

In the extension recorder, the process exit function does not stop the Browser Server, so the WebSocket server remains bound even after the browser closes.

The server is only stopped when it emits a stop event via recording stopping, but it is not stopped on process exit or browser errors along this path.

Funny enough, the CDP recorder already does this. 

### Grafana k6 Studio version

1.9.0

### OS

Windows 10, Windows 11

mohswell:
## Description
In the extension recorder, the process exit function does not stop the Browser Server, so the WebSocket server remains bound even after the browser closes.
The server is only stopped when it emits a stop event via recording stopping, but it is not stopped on process exit or browser errors along this path.

## Checklist

## Related PR(s)/Issue(s)
Resolves #984

### DIFF
--- a/src/handlers/browser/recorders/extension.ts
+++ b/src/handlers/browser/recorders/extension.ts
@@ -33,6 +33,13 @@ class BrowserExtensionRecordingSession
 
     process.once('exit', () => {
       this.emit('stop', undefined)
+      // Ensure the WebSocket server is disposed when the browser exits
+      this.#server.stop()
+    })
+
+    this.#process.on('error', (error) => {
+      this.emit('error', { error })
+      this.#server.stop()
     })
 
     server.on('stop', () => {


mohswell:
@Llandy3d since its merged. Quick one

random question not related to this PR. I faced another issue but it only occured once. in server file there is this part that references localhost
`
    const transport = await WebSocketServerTransport.create('localhost', 7554)
`

This is error I got `listen EACCES ::1:7554`
On some Windows machines, the system picks IPv6 (::1) by default, which can cause permission errors or binding failures.

> on windows OS whenever you write localhost, the system can resolve it to either:
> 1. IPv4: 127.0.0.1 (older, simpler)
> 2. IPv6: ::1 (newer, but sometimes causes issues on Windows)


My thought process is why not use 127.0.0.1 instead of localhost to prevent any strange windows OS bugs

PR #1016, opened 2026-01-15
feat: Move files to Trash bin on deletion

going-confetti:
## How to Test
1. Delete a file (can be done from the sidebar or from the actions menu in the top right corner when the file is open)
2. Verify that it's removed from the sidebar
    - If you deleted the file from the actions menu, you should be redirected to the home view
3. Open Trash and restore the file
4. Verify that it's displayed in the sidebar again and you can open it

## Checklist


src/handlers/ui/index.ts
```diff
@@ -46,7 +46,13 @@ export function initialize() {
     console.info(`${UIHandler.DeleteFile} event received`)
 
     const filePath = getFilePath(file)
-    return unlink(filePath)
+
+    try {
+      await shell.trashItem(filePath)
+    } catch (error) {
```
going-confetti, line 52: It is possible to disable recycle bin on windows (most likely, on other platforms as well), so `unlink` is still there as a fallback.

src/hooks/useDeleteFile.tsx
```diff
@@ -0,0 +1,75 @@
...
+        // If deletion fails, restore the file
+        addFile(file)
+        showToast({
+          title: `Failed to delete ${FileTypeToLabel[file.type]}`,
+          description: file.displayName,
+          status: 'error',
+        })
+      }
+    }, TOAST_DURATION_MS)
+  }
```
going-confetti: There are some edge cases with this approach, the most obvious one is when the user closes the app before the timeout, and the deletion hasn't completed yet - the file won't be deleted.
Of course this can be fixed, but I couldn't think of a simple way that wouldn't require even more changes than this PR already has. It's _probably_ not a big deal, but let me know what you think

PR #1031, opened 2026-01-20
internal(browser): Browser action editing MVP

going-confetti:
## Description
image

What I'd like to do next:
- Add file type icons
- Improve the action editing UX
- Undo/redo: the new state hook already supports this (thanks to @allansson 👏), but I'd like to explore it in a separate PR, as getting it to work together with the built-in undo functionality when editing text fields is a bit tricky

## How to Test
1. In the sidebar, look for the `TESTS` section and click on the `+` icon button
2. Select `Browser test`
3. Once opened, add a new "Navigate to URL" action in the browser actions list on the right
4. An action with prefilled value will be added, and you'll see `page.goto("https://example.com")` in the script preview
5. Test the debug/export/save/run in the cloud functionality

## Checklist

### DIFF
--- a/src/views/BrowserTestEditor/BrowserTestEditor.hooks.ts
+++ b/src/views/BrowserTestEditor/BrowserTestEditor.hooks.ts
@@ -120,23 +120,12 @@ export function useBrowserTestState(
   browserTestFile: BrowserTestFile | undefined
 ) {
   const { actions = [] } = browserTestFile ?? {}
-  const { state, undo, redo, push, reset } = useStateWithUndo<
-    BrowserActionWithId[]
-  >(
+  const { state, undo, redo, push } = useStateWithUndo<BrowserActionWithId[]>(
     actions.map((action) => ({
       id: crypto.randomUUID(),
       action,
     }))
   )
-
-  useEffect(() => {
-    reset(
-      actions.map((action) => ({
-        id: crypto.randomUUID(),
-        action,
-      }))
-    )
-  }, [reset, actions])
 
   const addAction = (method: AnyBrowserAction['method']) => {
     const action = createNewAction(method)
@@ -153,15 +142,6 @@ export function useBrowserTestState(
   const removeAction = (id: string) => {
     const newActions = state.filter((actionWithId) => actionWithId.id !== id)
     push(newActions)
-  }
-
-  const resetActions = (newState: AnyBrowserAction[]) => {
-    reset(
-      newState.map((action) => ({
-        id: crypto.randomUUID(),
-        action,
-      }))
-    )
   }
 
   const plainActions = useMemo(() => {
@@ -184,7 +164,6 @@ export function useBrowserTestState(
     isDirty,
     undo,
     redo,
-    resetActions,
   }
 }
 
--- a/src/views/BrowserTestEditor/EditableAction.tsx
+++ b/src/views/BrowserTestEditor/EditableAction.tsx
@@ -16,7 +16,7 @@ import {
   Trash2Icon,
   TriangleAlertIcon,
 } from 'lucide-react'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import { AnyBrowserAction, PageGotoAction } from '@/main/runner/schema'
 import { exhaustive } from '@/utils/typescript'
@@ -155,6 +155,7 @@ interface GoToActionBodyProps {
 function GoToActionBody({ action, onUpdate }: GoToActionBodyProps) {
   const [isPopoverOpen, setIsPopoverOpen] = useState(false)
   const [url, setUrl] = useState(action.url)
+
   const [error, setError] = useState<string | null>(null)
 
   const handleSubmit = (e: React.FormEvent) => {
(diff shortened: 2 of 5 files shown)


src/views/BrowserTestEditor/EditableAction.tsx
```diff
@@ -0,0 +1,194 @@
...
+  })
+
+  const handleBlur = async () => {
+    await trigger('url')
+
+    const data = getValues()
+    if (data.url !== action.url) {
+      onUpdate(data)
+    }
+  }
```
going-confetti: Not sure how I feel about react-hook-form, especially in this case, but I tried not to step too far away from patterns established in other places. Of course, any feedback is welcome.

src/views/BrowserTestEditor/EditableAction.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  return (
+    <>
+      Navigate to{' '}
+      <TextField.Root
+        size="1"
+        color={errors.url ? 'red' : 'gray'}
```
going-confetti: When the user types an invalid URL, should we still put it in the script (current behavior) or should we "skip" the action?

src/views/BrowserTestEditor/EditableAction.tsx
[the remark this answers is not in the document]
going-confetti: This is _kinf of_ intentional. See my other comment

src/views/BrowserTestEditor/EditableAction.tsx
allansson: I say it should go in the script as-is, but we invalidate the field or show a warning. I think omitting things from the script should be reserved for when the input would produce invalid syntax, in which case I think we should emit a comment.

src/views/BrowserTestEditor/EditableAction.tsx
allansson: I feel like this is targeted at me directly. 😆

Introducing johan-forms:

```tsx
// Define the different inputs upfront. Keeps the component code clean + they can be reused in different forms.
const username = input<string>({
  name: 'username',
  // Not using Zod here because it is overkill when validating primitive values
  validate(value) {
    if (value.trim().length === 0) {
      return { error: 'Username cannot be empty' }
    }

    return { result: value.trim() }
  },
})

const password = input<string>({
  name: 'password',
  validate(value) {
    if (value.length < 6) {
      return { error: 'Password must be at least 6 characters long' }
    }

    return { result: value }
  },
})

export function MyComponent() {
  // The form data is just normal component state
  const [data, setData] = useState(() => {
    return {
      username: username(''),
      password: password(''),
    }
  })

  // Values are updated 
  const handleUsernameChange = (username: Input<string>) => {
    setData((prev) => ({ ...prev, username }))
  }

  const handlePasswordChange = (password: Input<string>) => {
    setData((prev) => ({ ...prev, password }))
  }

  // The form is constructed from its inputs, not the other way around
  // like react-hook-forms where inputs are created from the form.
  const form = useForm(data, {
    onSubmit: (data) => {
      console.log('Form submitted with data:', data)
    },
  })

  return (
    <form {...form}>
      <Input
        label="Username"
        value={data.username}
        onChange={handleUsernameChange}
      />
      <Input
        label="Password"
        value={data.password}
        onChange={handlePasswordChange}
      />
    </form>
  )
}

interface InputProps<T> {
  label: ReactNode
  // Inputs have very simple types so they are easy to pass around as props
  value: Input<T>
  onChange: (value: Input<T>) => void
}

function Input<T>({ label, value, onChange }: InputProps<T>) {
  return (
    <div>
      <label {...value.label}>{label}</label>
      <input
        // Values have props for inputs and labels for convenience
        {...value.input}
        // Values are immutable and can be updated via `change`, `blur` and `set`.
        onChange={(e) => onChange(value.change(e.target))}
        onBlur={(e) => onChange(value.blur(e.target))}
      />
      {value.dirty && value.error && (
        <div style={{ color: 'red' }}>{value.error}</div>
      )}
    </div>
  )
}
```

src/views/BrowserTestEditor/BrowserTestEditor.hooks.ts
```diff
@@ -20,36 +33,189 @@ export function useBrowserTestFile(): StudioFile {
...
+  const { state, undo, redo, push, reset } = useStateWithUndo<
+    BrowserActionWithId[]
+  >(
+    actions.map((action) => ({
+      id: crypto.randomUUID(),
+      action,
+    }))
+  )
+
+  useEffect(() => {
```
allansson: This is to reset the state if the user selected a different file, right? You can do this by setting a key on the `BrowserTestEditor`:

```tsx
function BrowserTestEditorView({ fileName }) {
  // Same logic as the current `BrowserTestEditor`
}

function BrowserTestEditor() {
  const { fileName } = useParams()

  return <BrowserTestEditorView key={fileName} fileName={fileName} />
}
```
This will ensure that state is completely reset and it removes the footgun of forgetting to add a `useEffect` in some component deep down in the hierachy.

Issue #954, opened 2025-12-18
Add session replay to browser debugger


allansson:
### Feature description

To fully understand what's going on in the browser we need to add a session replay to the debugger.

### Suggested solution (optional)

Based on a small proof of concept, this is roughly how it would work:

1. The proxied `newPage` and `newContext` functions inject a script with `rrweb`.
2. `rrweb` sends its events to the tracking server hosted by k6 Studio.
3. Events are synced to the frontend of the app.
4. The session is replayed in the frontend using rrweb's Replayer-class.

The UI will be a video player with a play/pause and a timeline for scrubbing. 

Actions are highlighted as the recording plays. Clicking an action takes the recording to that point in time.

PR #1035, opened 2026-01-22
feat(browser): Session replay while debugging browser tests

allansson:
## Description

This PR adds session replay to the browser debugger. 

https://github.com/user-attachments/assets/4abf1d87-f0fc-49bd-9bea-fb7592823cab

**Features**:

* Streams live while debugging
* Play, pause, seek
* Content scales to fit viewport

**Known issues and limitations**:

* There are noticeable stutters while scrubbing the timeline because rrweb has to rebuild the DOM. I tried to throttle it, but in a way that made it worse because the gaps that rrweb had to replay became bigger whenever the throttle was released. I don't want to disable scrubbing altogether because it's nice to be able to see where you are in the recording. 
* You cannot pause or seek during live streaming. Since the max of the timeline keeps increasing while streaming, the thumb would jitter a bunch and given the short length of a normal session I didn't find the effort worth it.
* After the session has stopped, there is a brief moment where the timeline is enabled but keeps playing to the end. This is because rrweb keeps a buffer to make sure the playback is smooth. I tried disabling the timeline until the `recording-end` event is received, but then the "Debug script" button and "Script stopped"-notification would appear before the timeline is enabled. I find it less jarring when everything happens at once, even if it's a bit weird.

**Future improvements**:

* Keep track of page transitions and show proper loading states on page navigations
* Highlight actions in sidebar during playback
* Seek to a specific time by clicking on an action in the sidebar
* Add markers to timeline for each action
* Highlight elements while hovering over selectors in the sidebar
* Zoom settings: fit to view, original size
* Playback speed
* Support for multiple pages/tabs

## How to Test

1. Make sure you have a browser script
2. Open the debugger
3. Click "Debug script"
    * The view should switch to the Replay tab
4. You should see a replay of the browser while the test is running
5. When the script has completed, try interacting with the player:
     * Play
     * Pause
     * Seek

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6-studio/issues/954

### DIFF
--- a/src/views/Validator/Browser/SessionPlayer/PlaybackControls.tsx
+++ b/src/views/Validator/Browser/SessionPlayer/PlaybackControls.tsx
@@ -78,6 +78,7 @@ export function PlaybackControls({
         size="1"
         css={css`
           white-space: nowrap;
+          font-variant-numeric: tabular-nums;
         `}
       >
         <Flex align="center" justify="end" minWidth="80px">


src/main/runner/shims/browser/proxies/page.ts
```diff
@@ -1,11 +1,32 @@
...
+      return
+    }
+
+    trackLog({
+      level: type === 'log' ? 'info' : type,
+      msg: msg.text(),
+      time: new Date().toISOString(),
+      source: 'browser',
+    })
+  })
```
allansson: This started out as a way of debugging the replay script, but now that it's there I don't want to remove it. I will create a separate PR to improve the console component to include the `source` of the message so that it's possible to distinguish browser logs from k6 logs.

src/main/runner/shims/browser/index.ts
```diff
@@ -1,7 +1,18 @@
 import { BrowserContext, browser } from 'k6/browser'
 
 import { pageProxy } from './proxies/page'
-import { createProxy, ProxyOptions } from './utils'
+import { createProxy, ProxyOptions, TRACKING_SERVER_URL } from './utils'
+
+const SESSION_REPLAY_SCRIPT =
```
allansson: I couldn't get vite to build `replay.ts` and then import it as a raw string. I also tried using the `open` function to read the file but there were some path issues that I couldn't figure out. And I didn't want to pass the entire script as an environment variable, so here we are...

src/main/runner/rrweb.ts
```diff
@@ -0,0 +1,57 @@
...
+    height: z.number(),
+  }),
+})
+
+const CustomReplayEventSchema = z.discriminatedUnion('tag', [
+  RecordingEndEventSchema,
+  PageStartEventSchema,
+])
+
+const RrwebCustomEventSchema = z.object({
```
allansson, line 26: rrweb loves typing things as `unknown` with no way of specifying types yourself. Zod to the rescue.

src/main/runner/schema.ts
```diff
@@ -296,6 +303,7 @@ export type ActionBeginEvent = z.infer<typeof ActionBeginEventSchema>
 export type ActionEndEvent = z.infer<typeof ActionEndEventSchema>
 
 export type BrowserActionEvent = z.infer<typeof BrowserActionEventSchema>
+export type BrowserReplayEvent = eventWithTime
```
allansson, line 314: For some reason rrweb uses camel-case for type names and I can't stand seeing that so I've aliased it here.

src/utils/k6/testRun.ts
```diff
@@ -88,11 +88,38 @@ export interface TestRunLogEvent {
 }
 
 interface TestRunEventMap {
+  /**
```
allansson, line 91: Couldn't remember what each of the events were doing so I looked it up and added some comments. You're welcome, future me!

src/views/Validator/Browser/BrowserActionsPanel.tsx
```diff
@@ -0,0 +1,77 @@
+import { css } from '@emotion/react'
```
allansson, line 1: Extracted from `BrowserDebugger`. No real changes here.

allansson (commented):
Self-review

src/utils/k6/tracking.ts
```diff
@@ -56,7 +66,17 @@ export async function createTrackingServer(): Promise<TestRunTrackingServer> {
   const httpServer = new Server(app)
   const trackingServer = new TestRunTrackingServer(httpServer)
 
-  app.use(json())
+  app.use(
+    cors({
+      origin: '*',
+    })
+  )
```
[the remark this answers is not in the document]
allansson: Since browser scripts can navigate to any page, we do need to keep it open. Users would have to navigate to malicous web pages in their script for this to be a real issue.

src/utils/k6/tracking.ts
```diff
@@ -56,7 +66,17 @@ export async function createTrackingServer(): Promise<TestRunTrackingServer> {
...
+
+  app.use(
+    json({
+      limit: '100mb',
+    })
+  )
```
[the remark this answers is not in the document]
allansson: Maybe it's a bit much, but given the use-case of k6 Studio it's not really a concern so I'm going to keep it.

src/views/Validator/Browser/SessionPlayer/PlaybackControls.tsx
```diff
@@ -0,0 +1,89 @@
...
+        onPause={onPause}
+      />
+
+      <Slider
+        size="1"
+        disabled={streaming}
+        value={[currentTime]}
+        step={0.001}
+        min={0}
+        max={totalTime}
```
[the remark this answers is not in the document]
allansson: This is deliberate. If we set `currentTime` to `0` then the slider will be rendered as empty (at the beginning) but we want it to render as full (at the end).

src/views/Validator/Browser/SessionPlayer/Viewport.tsx
```diff
@@ -0,0 +1,76 @@
...
+      const parentWidth = target.parentElement.clientWidth
+      const parentHeight = target.parentElement.clientHeight
+
+      const elementWidth = target.offsetWidth
+      const elementHeight = target.offsetHeight
+
+      const scaleX = parentWidth / elementWidth
+      const scaleY = parentHeight / elementHeight
+
+      const scale = Math.min(scaleX, scaleY, 1)
```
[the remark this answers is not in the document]
allansson: The behaviour when dividing by zero is undefined so I'm going to leave the behaviour of this function undefined when dividing by zero, i.e. I'm going to ignore it.

PR #1037, opened 2026-01-23
feat(browser): Add comments during browser recordings

### LINKED ISSUE #901: Add comments during recording
### Feature description

Sometimes you want to annotate what you are doing while recording a browser test (and protocol test for that matter). We should implement a way to add a comment to through the browser UI which would later be emitted in the generated script.

Add a new tool to the toolbar to add a comment. Clicking the tool would open a small popup with a text field. When a script is generated from the recording, the comment should appear as a code comment above the next recorded action.

_No response_

mohswell:
## Description
This pull request introduces a new "Add Comment" feature to the browser recording whereby users can now add comments directly in the browser UI, which are then recorded as comment events and displayed as comments in the js test scripts. Users can annotate their browser-based test recordings with comments, improving scripts readability. e.t.c

I had a few approaches to tackle this, but for an interactive experience, a popover that follows the cursor click/placement would be best. That's a comment popover similar to what Figma uses.
image
    
The comment added is then shown in the script file:
image

## My TODO's
- [x] I understand the implementation does not have a design mockup
- [ ] I am open to changing the implmentation to a dropdown in the toolbar instead of the cursor popover

## Improvements
- Supporting multi line comments maybe I should use a text area instead of text field input
- Linking the annotation label next to the element to indicate where the comment is placed
- Editing, deleting the comment added by the user 
- limiting the number of characters that a inline comment has using a constant like MAX_COMMENT_LENGTH
- improving the cursor popover wth some animations 
- I cant figure out whats causing an extra line after adding comment (the space below the comments and the next event).

## Related PR(s)/Issue(s)
Resolves #901


extension/src/frontend/view/Comment/CommentPopover.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (e.key === 'Enter') {
+      e.preventDefault()
+      handleSubmit()
+    }
+  }
+
+  return (
+    <div
+      css={css`
+        position: absolute;
```
mohswell: absolute position to ensure when the page is scrolled, the cursor is still able to track and place a comment anywhere

src/codegen/browser/codegen.test.ts
```diff
@@ -776,6 +776,63 @@ it('should assert that aria input is not checked', async ({ expect }) => {
...
+          source: 'address-bar',
+          inputs: {
+            page: { nodeId: 'page' },
+          },
+        },
+        {
+          type: 'comment',
+          nodeId: 'comment-1',
+          text: 'Expecto Patronum! This test now magically passes',
+          inputs: {
```
mohswell, line 800: how did this get in here ?😀

src/codegen/browser/code/options.ts
```diff
@@ -59,6 +59,9 @@ function isBrowserScenario(scenario: ir.Scenario) {
       case 'ExpectExpression':
         return visit(node.actual) || visitAssertion(node.expected)
 
+      case 'Comment':
+        return false
+
```
mohswell, line 64: Returning false means comments don't represent actual browser operations

extension/src/frontend/view/Comment/CommentPopover.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+              &:focus {
+                outline: none;
+                box-shadow: none;
+              }
+            }
+          `}
+        />
+        <IconButton
+          onClick={handleSubmit}
```
allansson: Nit: I'm a stickler for this but event handlers go last among the props. 🙈

allansson (changes requested):
Hey! Thanks for the PR!

I've taken a look and I have a few thoughts.

First of all, I don't like how the comment has space after it in the generated output. I've created a [separate PR](https://github.com/grafana/k6-studio/pull/1038) to make the comments stick to the next statement. You can merge that in once it's in main.

Next, I think we should remove the "Click anywhere to write a comment" step. I think the UX is a bit confusing because it doesn't really matter where you click and what you click is not connected to the comment in any way. Clicking the comment icon could just display a popover below the toolbar:

Image

Or even better would be to have a centered dialog with a dimmed background:

Image

Like you say in the PR decriptions, I think it'd be a good idea to make the input a text area. You can add multi-line comments by modifying [this code](https://github.com/mohswell/k6-studio/blob/37b6cb71f21bd25461e1fc1679086ef62912832c/src/codegen/browser/formatting/formatter.ts#L72).

```ts
const comments = node?.comment.split("\n")
  .flatMap((line, index) => {
    const comment = ["// ", line]

    if (index > 0) {
      return [hardline, ...comment]
    }

    return comment
  })

const doc =
  comments === undefined
    ? estree.print(path, options, print)
    : comments
```

Finally, while I think the design of the editor popover is nice, it's not consistent with the rest of the UI. In time we might want to take another look at the designs, but until then keeping things consistent is more important. Here's what the text assertion editor looks like:

Image

extension/src/frontend/view/ToolBox/index.tsx
```diff
@@ -150,6 +152,11 @@ export function ToolBox({
               <TextCursorIcon />
             </Toolbar.ToggleItem>
           </ToolBoxTooltip>
+          <ToolBoxTooltip content="Click anywhere on the page to add a comment">
```
mohswell, line 155: To update this content description

e-fisher:
Closing as inactive

PR #1041, opened 2026-01-28
deps: Migrate to React 19

going-confetti:
## Description
Looks like we're _finally_ ready to do it.

## How to Test
First, test individual dependency updates:
- `testing-library/react`: `npm run test`/CI check should work as expected
- `allotment`: it's a patch bump, but do check that panels that use allotment work as before
- `zustand`: it's a patch bump, verify that the app works as expected(specifically Recorder, Generator, the sidebar, etc)
- `@emotion/react`: it's a minor bump, shouldn't really require much testing, but check that everything looks okay and `npm start`/`npm run make` work

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@dnd-kit/modifiers": "^9.0.0",
         "@dnd-kit/sortable": "^10.0.0",
         "@emotion/cache": "^11.14.0",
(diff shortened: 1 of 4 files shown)


allansson (approved):
Tried everything I could think of. Couldn't find anything. Let's go! 🚀

PR #1075, opened 2026-02-03
internal(fix): Session replay does not work in production builds

allansson:
## Description

Session replay doesn't work in production builds. This is because the variable names in the entrypoint script are mangled during minification and we rely on there being a specific variable called `SESSION_REPLAY_SCRIPT` where we inject the script to be loaded into the browser.

## How to Test

1. Run `npm run make`
2. Check that the `resources/entrypoint.js` script contains a variable called `SESSION_REPLAY_SCRIPT` and that its value is an empty string.

## Checklist

### DIFF
--- a/vite.script.config.ts
+++ b/vite.script.config.ts
@@ -9,6 +9,7 @@ export default defineConfig((env) => {
     mode: env.mode,
     build: {
       sourcemap: false,
+      minify: false,
       target: 'esnext',
       lib: {
         entry: 'src/main/runner/entrypoint.ts',


vite.script.config.ts
```diff
@@ -9,6 +9,7 @@ export default defineConfig((env) => {
     mode: env.mode,
     build: {
       sourcemap: false,
+      minify: false,
```
allansson, line 12: Minification serves no purpose for the entrypoint script since it's executed locally by k6.

going-confetti (approved):

PR #1090, opened 2026-02-16
fix(browser): Pages are not disposed of properly at end of test

### LINKED ISSUE #885: Exported browser script does not close page on completion
### Describe the bug

Browser scripts should call `page.close` before the end of each iteration. This is usually done inside a `try-finally` clause to make sure the page is always closed.

Exported scripts don't do this however.

1. Create a recording
2. Export a browser script

### Expected behavior

The code should be wrapped in a try-finally and the page should be properly closed:

```ts
export default async function () {
  const page = await browser.newPage()

  try {
    await page.goto()
  } finally {
    await page.close()
  }
}
```

### Actual behavior

The code is not wrapped and no call to `page.close` is done.

```ts
export default async function () {
  const page = await browser.newPage()

  await page.goto()
}
```

### Grafana k6 Studio version

1.8.0

### OS

Any

allansson:
## Description

Browser scripts should close pages at the end of the script to make sure that all metrics are collected correctly, resources are freed, etc. ([docs](https://grafana.com/docs/k6/latest/using-k6-browser/recommended-practices/clean-up-test-resources-page-close/))

The script exported by k6 Studio wasn't emitting a try-finally statement closing opened pages, but this PR changes that.

## How to Test

1. Record a test
2. Export a browser script
3. Make sure that the script has a try-finally around all uses of any page variable
4. Run the script

## Checklist

## Related PR(s)/Issue(s)

Fixes #885

### DIFF
--- a/src/codegen/browser/__snapshots__/browser/merged-try-finally-blocks.ts
+++ b/src/codegen/browser/__snapshots__/browser/merged-try-finally-blocks.ts
@@ -25,7 +25,7 @@ export default async function () {
 
     await page1.goto("https://example2.com");
   } finally {
-    await page1?.close();
     await page2?.close();
+    await page1?.close();
   }
 }


src/codegen/browser/intermediate/index.ts
```diff
@@ -12,22 +12,30 @@ function emitPageNode(context: IntermediateContext, node: m.PageNode) {
...
 
 function emitGotoNode(context: IntermediateContext, node: m.GotoNode) {
-  const page = context.reference(node.inputs.page)
-
   // Skip goto for implicit navigation, as it will be triggered by preceding action
   if (node.source === 'implicit') {
     return
   }
 
+  const page = context.reference(node.inputs.page)
```
allansson, line 37: This had to be moved after the condition, otherwise it would decrease the reference count of the page even when it wasn't used (i.e. `source` was `'implicit'`)

src/codegen/browser/intermediate/index.ts
```diff
@@ -204,6 +211,8 @@ function emitClickNode(context: IntermediateContext, node: m.ClickNode) {
   }
 
   if (node.triggersNavigation) {
+    const page = context.reference(node.inputs.page)
```
allansson, line 214: Same here. Any click node that didn't trigger a navigation would decrease the reference count, even though the page wasn't being used.

src/main/runner/shims/browser/proxies/page.ts
```diff
@@ -53,6 +53,12 @@ export function pageProxy(target: Page): ProxyOptions<Page> {
         }
       },
 
+      close() {
```
allansson, line 56: I'm not sure if the `page.close` call should be displayed in the browser action list. We could just ignore tracking it in the `$default` handler. WDYT?

e-fisher: Could this help with debugging, perhaps a race condition when some action happens after page is closed? Also, if we keep it displayed, should we add corresponding "Open page" event as well? 🤔

e-fisher (dismissed):
LGTM 🚀 
Test snapshot needs to be updated after change of order

going-confetti (approved):
LGTM

PR #1093, opened 2026-02-24
internal(browser): Add wait for element action

going-confetti:
Closes https://github.com/grafana/k6-studio/issues/1046
Closes https://github.com/grafana/k6-studio/issues/1051

## Description
This PR adds support for adding and configuring the `locator.waitFor()` (all options supported) and `page.reload()` (no options supported yet) actions.

https://github.com/user-attachments/assets/35ee25be-2211-448a-b699-b4819faa29b4

## How to Test
- Refer the the attached screen recording
- Verify that the changes are reflected in the script preview
- Try to break it by entering invalid options (non-numeric timeouts, etc)

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Expands the browser test editor/action schema and codegen pipeline (actions → node graph → emitted script), so regressions could affect script generation and recorded selector matching. UI changes are mostly additive but touch core editor state handling and validation.
> 
> **Overview**
> Adds first-class Browser Test Editor support for **`locator.waitFor()`** (with `state`/`timeout` options) and **`page.reload()`**, including new action editors/forms, option popovers, validation, and a small options summary UI.
> 
> Refactors editor state to use `BrowserActionInstance` (adds IDs + locator option metadata) with adapters for persistence, debounced script preview generation, and updates action→test conversion to emit `reload`/`wait-for` nodes while reusing locator nodes across sequential actions.
> 
> Updates selector/codegen plumbing to support richer role locator options (`name` + optional `exact`), adjusts equality/DOM query behavior accordingly, and removes the unused `useStateWithUndo` hook/tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be62175b2c5ef163798dd8ef6fc8bf5130b3bee6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/hooks/useStateWithUndo.ts
going-confetti, line 1: I decided to get rid of this for now as it wasn't originally in scope.

src/views/BrowserTestEditor/ActionForms/fields/role.ts
```diff
@@ -0,0 +1,49 @@
...
+  'link',
+  'checkbox',
+  'radio',
+  'switch',
+  'textbox',
+  'searchbox',
+  'combobox',
+  'listbox',
+  'option',
+].map((role) => ({ value: role, label: role }))
```
going-confetti: There are [too many roles to list here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles), so the idea is to display some of the common ones, while giving the user an option to enter an arbitrary value.

I'm not sure this list makes sense though, so please let me know if something important is missing (or if the opposite is the case).

src/views/BrowserTestEditor/ActionForms/forms/LocatorForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+    setTouchedTypes((prev) => {
+      if (prev.has(current)) {
+        return prev
+      }
+      const next = new Set(prev)
+      next.add(current)
+      return next
+    })
+  }
+
```
e-fisher, line 85: NIT: same logic 'add to Set if not present' repeated multiple times, could be extracted into helper
```ts
const addToSet = <T,>(set: Set<T>, value: T) =>
  set.has(value) ? set : new Set(set).add(value)
```

src/views/BrowserTestEditor/ActionForms/forms/LocatorForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+                  <RadioGroup.Item value={type} key={type}>
+                    {label}
+                  </RadioGroup.Item>
+                ))}
+            </RadioGroup.Root>
+          </FieldGroup>
+
+          <Separator orientation="vertical" size="4" decorative />
+          <Flex direction="column" gap="2" align="stretch">
+            {fields.map((field) => (
```
allansson: I have my qualms about this. We're expressing static structure through dynamic evaluation, and I think that's bad for a couple of reasons:

* It is harder to reason about. To understand what the final result will look like I have to evaluate in my mind what the code will do, whereas if I explicitly implement these as forms I can understand it at a glance:
    ```tsx
    {currentLocator.type === "role" && <GetByRoleForm />}
    ```
* There will inevitably come a point where we hit some limitation of the generic approach and, through [our tendency to solve problems by adding more things](https://en.wikipedia.org/wiki/Additive_bias), we'll start extending upon it until it becomes its own framework.

  As an example, the `click` action can be configured with an X and a Y offset. It seems resonable to render these as a `Position` label with an input for X and Y each. Here's some ASCII art:

    ```
               ____   ____    
    Position: | 10 | | 50 |
               ----   ---- 
    ```

    The temptation here is to solve the problem by adding new constructs to the `FieldConfig` interface (maybe some kind of recursive multi-input or a layout type?), but it might just be that the problem was that we were trying to solve the problem generically to begin with.

Writing these out as components is tedious, boring and repetitive but it is explicit and easy to understand.

That being said, I do like the fields and their definitions being defined separately and I don't want to push too hard for changing anything as it is still pretty simple and I respect that others might have a different opinion on this.

src/views/BrowserTestEditor/ActionForms/fields/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  name: string
+  label: string
+  input: FieldInput
+  clearable?: boolean
+  placeholder?: string
+  options?: Array<{ label: string; value: string }>
+  getValue(model: TModel): TValue
+  setValue(model: TModel, value: TValue): TModel
+  validate?(value: TValue, model?: TModel): string | null
+}
```
allansson: Because I'm a typescript zealot:

```suggestion
interface FieldConfigBase<TValue, TModel> {
  name: string
  label: string
  input: FieldInput
  clearable?: boolean
  placeholder?: string
  getValue(model: TModel): TValue
  setValue(model: TModel, value: TValue): TModel
  validate?(value: TValue, model?: TModel): string | null
}

interface SelectFieldCOnfig<TValue, TModel> extends FieldConfigBase<TValue, TModel> {
  input: "select" | "combobox"
  options: Array<{ label: string; value: string }>
}

interface InputFieldConfig<TValue, TModel> {
  input "text" | "number" | "textarea"
}

export type FieldConfig<TValue, TModel = TValue> = 
  | InputFieldConfig<TValue, TModel>
  | SelectFieldConfig<TValue, TModel>
```

allansson (commented):
Everything works like a charm.

I left a couple of comments on the code. Do with them what you will.

Either way, this is looking great. 🙌

src/views/BrowserTestEditor/ActionForms/forms/LocatorForm.tsx
going-confetti: I gave it some thought and I agree. Pushed some changes 🙌

src/views/BrowserTestEditor/ActionForms/forms/locators/GetByRoleForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+          name="name"
+          value={locator.options?.name || ''}
+          onChange={(e) => {
+            const value = e.target.value
+            onChange({
+              ...locator,
+              options: value.trim()
+                ? { ...locator.options, name: value }
+                : locator.options,
+            })
```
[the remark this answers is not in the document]
going-confetti: I know about this (it's also the case with the state select in the waitFor options). I'll address this issue in a future PR.

src/codegen/browser/selectors.ts
```diff
@@ -109,6 +110,62 @@ export function getNodeSelector(selector: ElementSelector): NodeSelector {
...
+      return {
+        type: 'css',
+        selector: locator.selector,
+      }
+
+    case 'role':
+      return {
+        type: 'role',
+        role: locator.role,
+        name: locator.options?.name ?? '',
```
allansson: Right now, the editor says `name` is optional, but it still outputs `{ name: "", exact: true }` in the script.

I think we should propagate the `undefined` here and edit the codegen to only emit the options argument if `name` was the empty string.

```ts
interface GetByRoleName {
  value: string
  exact: boolean
}

interface GetByRoleNodeSelector {
  // ...
  name: GetByRoleName | undefined
}
```

Then update the codegen to conditionally emit the options object:

```ts
const options = node.name && ObjectBuilder.from({
  name: emitExpression(context, node.name.value),
  exact: node.name.exact ? true : undefined // `ObjectBuilder.from` will ignore properties if they are undefined
})
```

Sorry for the backseat coding! 😓

src/views/BrowserTestEditor/ActionForms/forms/locators/GetByRoleForm.tsx
```diff
@@ -68,13 +70,42 @@ export function GetByRoleForm({
...
+                ...locator.options,
+                name: value.trim() ? value : undefined,
+              },
             })
           }}
           onBlur={onBlur}
-        />
+        >
+          <TextField.Slot side="right">
+            <Tooltip content="Match case">
```
allansson: The `exact` option acts like a combination of whole string and case-sensitive search so this is a little bit misleading.

src/codegen/browser/selectors.ts
```diff
@@ -22,7 +22,7 @@ function getRoleSelector(
   return {
     type: 'role',
     role: selectors.role.role,
-    name: selectors.role.name,
+    name: selectors.role.name ? { value: selectors.role.name } : undefined,
```
allansson: This should have `exact: true` to avoid changing the generated output.

PR #1096, opened 2026-02-25
build: Use `bundler` and `esnext` for module resolution

allansson:
## Description

So far we've been using `commonjs` modules in `tsconfig.json`, but there's no technical reason for us to do so. Electron is based on CommonJS, but Vite is transforming ES moduels to CJS for us so typescript doesn't really have to worry about it.

The biggest drawback has been that we've had to add `@ts-expect-error` to any place where we want to use `import.meta`, but I recently tried adding a new package and typescript couldn't find the correct types without setting `moduleResolution` to `bundler`.

The changes are pretty small. Besides the changes to `tsconfig.json`, all of the `@ts-expect-error` directives related to `import.meta` have been removed. In a couple of places there's been some changes to type imports.

## How to Test

* Start k6 Studio and check that everything works
* Run the type checker

## Checklist

### DIFF
--- a/src/components/Monaco/ConstrainedCodeEditor.tsx
+++ b/src/components/Monaco/ConstrainedCodeEditor.tsx
@@ -4,7 +4,7 @@ import constrainedEditor, {
   ConstrainedEditorInstance,
   RestrictionObject,
 } from 'constrained-editor-plugin'
-import * as monacoTypes from 'monaco-editor/esm/vs/editor/editor.api'
+import type * as monacoTypes from 'monaco-editor'
 import { useEffect, useRef, useState } from 'react'
 
 import { ReactMonacoEditor } from './ReactMonacoEditor'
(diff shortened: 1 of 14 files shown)


src/components/Monaco/ConstrainedCodeEditor.tsx
```diff
@@ -4,7 +4,7 @@ import constrainedEditor, {
   ConstrainedEditorInstance,
   RestrictionObject,
 } from 'constrained-editor-plugin'
-import * as monacoTypes from 'monaco-editor/esm/vs/editor/editor.api'
+import type * as monacoTypes from 'monaco-editor'
```
allansson, line 7: Import the types directly from `monaco-editor` instead. The `type`-only import is to make sure we don't pull in the entire `monaco-editor`.

PR #1097, opened 2026-02-26
build: Fix macos build error and improve local build/hmr performance

e-fisher:
This PR fixes 2 issue:

### Problem
The Electron 40 upgrade bumped Node from v22 to v24.11. After that, the macOS release jobs fail with:
```
Cannot find module 'appdmg'
```

appdmg is an optional transitive dependency (@electron-forge/maker-dmg -> electron-installer-dmg -> appdmg). When its native code fails to compile, npm silently skips it. At make time, require('appdmg') crashes.

### Cause
1. appdmg -> ds-store -> macos-alias@0.2.12 depends on nan@^2.4.0, which resolved to nan@2.20.0. NAN 2.20 does not support Node 24's V8 engine -- only NAN 2.25.0+ does. The V8 API changes cause C++ compilation errors in NAN's headers (IndexedPropertyHandlerConfiguration, ScriptOrigin constructors).
2. node-gyp requires Python's setuptools to orchestrate native builds. macOS GitHub Actions runners ship with Python 3.12+ where setuptools was removed 1. The previous workaround installed Python 3.11.4 on all platforms, even though appdmg is only needed on macOS.

### Fix
- package.json: Add "nan": "^2.25.0" to overrides, forcing all transitive NAN consumers to use a version that supports Node 24.
- release.yml / release-test-version.yml: Replace actions/setup-python (Python 3.11.4) with brew install python-setuptools, scoped to macOS only via if: runner.os == 'macOS'.

### Problem

After the Forge v7 upgrade, the `external` list was replaced with `builtins` (only Node built-ins + electron), forcing Vite to bundle all ~80+ npm dependencies on every build. This caused:

- Terribly slow dev builds - every file edit triggers rebundling of the entire dependency tree
- HMR crashes the bundler runs out of memory or times out during hot module replacement

Reverting to `external` (all deps externalized) broke the packaged app because `@electron-forge/plugin-vite` v7+ strips `node_modules` from the ASAR -- externalized packages are missing at runtime.

### Solution

Use different externalization strategies for dev vs production:

- **Dev** (`npm start`): externalize all CJS-compatible dependencies for fast builds and stable HMR. ESM-only packages (detected automatically by reading each dependency's `package.json` for `"type": "module"`) are kept bundled since CJS `require()` can't load them.
- **Production** (`electron-forge make`): only externalize Node built-ins, so Vite bundles everything. No `node_modules` needed in the ASAR, keeping app size unchanged.

Also:
- `tsconfig.json`: changed `"module"` from `"commonjs"` to `"ESNext"`. This is the correct setting for Vite projects -- it only affects type checking, not the actual build output (which is still CJS). This allows `import.meta.env` and `import.meta.vitest` to type-check without workarounds.
- Removed 11 `@ts-expect-error` comments across 9 files that existed solely to suppress `import.meta` errors under the old `commonjs` module setting.

## How to Test
**Local:**
1. Run the app, verify everythings works
2. Veify live code reload works
3. Modify electron file (e.g. main.ts), verify app reloads (previously crashed)

**Built app:**
Download build from custom release and verify everythings works as expected: https://github.com/grafana/k6-studio/releases/tag/untagged-33405c035a26900f92b0

## Checklist


tsconfig.json
allansson: We should change this to `bundler` as well.

.github/workflows/release-test-version.yml
```diff
@@ -51,12 +51,12 @@ jobs:
...
-      # https://github.com/electron/forge/issues/3371#issuecomment-2105195302
-      - name: Install Python 3.11.4
-        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
-        with:
-          python-version: "3.11.4"
+      # appdmg (used for DMG builds) depends on native modules compiled via node-gyp,
+      # which requires Python setuptools. macOS runners ship with Python 3.12+ where
+      # setuptools was removed. See https://github.com/electron/forge/issues/3371
+      - name: Install setuptools for appdmg
+        if: runner.os == 'macOS'
```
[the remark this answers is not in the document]
e-fisher: `runner.os` is actually the more robust check -- it doesn't depend on matrix naming. Using `startsWith(matrix.platform, ...)` is a workaround for cases where you need to distinguish between matrix entries that run on the same OS (e.g., `macos-latest` vs `macos-14` for different architectures).

If the other checks in the workflow use `startsWith(matrix.platform, ...)` because they need to differentiate between matrix variants, that's fine for those cases. But suggesting to replace `runner.os` with a string match purely for "consistency" would make the check less robust. I'd push back on that review comment.

tsconfig.json
e-fisher: Looks like you commented on older commit - it's updated 👍

PR #1164, opened 2026-04-13
internal(browser): Add Select Option action to Browser Test Editor

going-confetti:
## Description

- Add support for the Select option action
- ⚠️ This PR doesn't include any action options

grafik

## How to Test
- Add a new `Select option` action
- Configure the locator and add some options to be selected
- Verify that the code is properly generated

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Touches the browser test editor, action-to-test conversion, and codegen IR/emitter types, so regressions could impact generated scripts for select actions. Changes are additive and scoped, with minimal impact outside select-option handling.
> 
> **Overview**
> Adds a new **Select option** action to the Browser Test Editor, including UI to configure one or more option matchers by *value*, *label*, or *index*.
> 
> Updates the browser test graph/types and action-to-test conversion to carry structured select values (deduped) and set `multiple` automatically.
> 
> Extends the browser codegen intermediate AST and scenario emitter with a new `SelectOptionValueExpression` so generated `locator.selectOption(...)` calls can emit either strings or `{ value, label, index }` objects, and wires this through variable substitution and browser-scenario detection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51ba3a76b3d73efa3c8bf62017b97f6c5d3186cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/views/BrowserTestEditor/Actions/SelectOptionAction/SelectOptionActionBody.tsx
```diff
@@ -0,0 +1,48 @@
...
+      gap="2"
+      align="center"
+      width="100%"
+    >
+      Select
+      <SelectOptionValuesForm
+        values={action.values}
+        onChange={handleChangeValues}
+      />
+      in
```
going-confetti, line 40: This order is different from other actions, but feels more natural/easier to read IMO

src/codegen/browser/test.ts
```diff
@@ -423,12 +423,29 @@ function buildBrowserNodeGraphFromActions(
             locator: getLocator(action.locator),
           },
         }
+      case 'locator.selectOption': {
+        const nonEmpty = action.values.filter(
+          (v) =>
+            (v.value !== undefined && v.value !== '') ||
+            (v.label !== undefined && v.label !== '') ||
+            v.index !== undefined
+        )
```
allansson: `value` is allowed to be an empty string according to the HTML specification (but `label` is not). 
```suggestion
        const nonEmpty = action.values.filter(
          (v) =>
            (v.label !== undefined && v.label !== '') ||
            v.value !== undefined ||
            v.index !== undefined
        )
```

allansson: Maybe this is a dumb optimization, but we could dedupe the values for the user:

```ts
function getKey(v) {
  if (v.value !== undefined) {
    return `value:${v.value}`
  }

  if (v.label !== undefined) {
    return `label:${v.label}`
  }

  return `index:${v.index}`
}

const deduped = Object.values(keyBy(nonEmpty, getKey))
```

allansson (changes requested):
Works great! UI feels good.

I think we should allow empty strings for `value` because it is valid HTML, but I'm happy to hear counter arguments.

src/codegen/browser/test.ts
going-confetti: No, I think it's a good suggestion 👍

src/views/BrowserTestEditor/ActionForms/forms/SelectOptionValuesForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+        <ValuePopoverBadge
+          displayValue={
+            <span
+              css={css`
+                text-overflow: ellipsis;
+                white-space: nowrap;
+                overflow: hidden;
+              `}
+            >
+              <SelectOptions options={values} />
```
going-confetti, line 106: As noted by @allansson, empty values are not necessarily invalid, so `SelectOptions` needs to handle those in a better way. There should probably also be a better distinction between labels, values, and indexes. I'll try to create a separate PR for that change

Image

src/codegen/browser/test.ts
```diff
@@ -431,12 +433,31 @@ function buildBrowserNodeGraphFromActions(
...
         }
+      case 'locator.selectOption': {
+        const deduped = Object.values(
+          keyBy(action.values, (v) => {
+            if (v.value !== undefined) return `value:${v.value}`
+            if (v.label !== undefined) return `label:${v.label}`
+            return `index:${v.index}`
+          })
+        )
+        const selected = deduped.length > 0 ? deduped : ['']
```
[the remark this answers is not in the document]
allansson: After some internal discussion, this is the desired behavior.

allansson (approved):
LGTM! 👍 

One small thing: maybe we should render `value` and `label` with quotes to make it clear if it's an empty value. We could also maybe use icons to indicate whether its a label, value or index.

But I think it's fine as it is and we can add polish later.

going-confetti:
> One small thing: maybe we should render `value` and `label` with quotes to make it clear if it's an empty value. We could also maybe use icons to indicate whether its a label, value or index.
> 
> But I think it's fine as it is and we can add polish later.

Yes, I did mention this implicitly in a comment above. I'll try to open a PR later today.

PR #1238, opened 2026-05-13
refactor: Store paths to recording and data files as relative in HTTP tests

allansson:
## Description

This PR updates the file format for HTTP tests so that they store file references as relative to the HTTP test file.

The benefits of doing so is that HTTP tests become path independent, which enables:

* Referencing recordings and data files from any directory without copying them into `k6-studio/`
* Saving and opening HTTP tests to and from any directory
* More advanced organization of files, i.e. a workspace

Since we've locked everything down to the `k6-studio/` directory these changes should be safe. The only way I can think of where this could be a problem is:

1) The user has initialized a git repository in the `k6-studio` directory, and
2) Someone is trying to open the generator files in an older version of k6 Studio

## How to Test

1. Open an HTTP test
2. Add and remove data files
3. Setup parameterization rules using the data files
4. Switch recordings
5. Save and re-open the test

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Updates the on-disk generator file format and path serialization/deserialization logic, which can break opening/saving existing tests or resolving files if edge cases in relative path handling occur (especially across platforms).
> 
> **Overview**
> **Generator file format is bumped to `3.0`** and migrations are updated so older generator files (v0–v2) upgrade to v3.
> 
> **Recording and data file paths are now persisted as POSIX-style relative paths** (relative to the generator file) and converted back to native absolute paths on load; save/open/create handlers were updated to pass the generator `filePath` through `serializeGenerator`/`deserializeGenerator`.
> 
> Adds `v3` schemas (rules/options/thresholds/load zones) and a `toNative()` path helper with tests, and updates factories/selectors/tests to expect `version: '3.0'` and migrated relative paths (e.g. `../Recordings/...`, `../Data/...`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83d5e46b395c564d2842bbc7a465050dfb9164aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/handlers/generator/index.ts
```diff
@@ -16,12 +16,14 @@ import { GeneratorHandler } from './types'
...
     console.log(`${GeneratorHandler.Create} event received`)
-    const generator = createNewGeneratorFile(
-      recordingPath ? path.basename(recordingPath) : undefined
-    )
+    const generator = createNewGeneratorFile(recordingPath || undefined)
 
     const filePath = await createFileWithUniqueName({
-      data: JSON.stringify(generator, null, 2),
+      data: JSON.stringify(
+        serializeGenerator(getGeneratorPath('Generator'), generator),
```
allansson, line 23: This is a chicken and egg situation where we don't know the filename until `createFileWithUniqueName` returns but we need to provide the filename to `serializeGenerator` to create relative paths. Luckily we only need the `dirname` to do so, so we can just pretend that there will never be a collision.

If we ever get around to implementing workspaces, the user will have to pick where to store the file and then this problem goes away.

src/schemas/generator/v3/loadZone.ts
allansson, line 1: Unlike previous migrations, this one copies the entire schema. This prevents us from accidentally introducing breaking changes into older schema versions.

allansson (commented):
Self-review

allansson:
bugbot run

allansson:
bugbot run

e-fisher (approved):
LGTM

PR #1248, opened 2026-05-19
refactor: Use posix paths as internal representation

allansson:
## Description

This PR refactors the code to always use posix paths internally, converting them to native paths only at the application boundary.

Keeping a single internal representation of paths simplifies the application logic when dealing with them. For instance, we don't need to worry about which kind of path it is when storing them or when generating data file imports in scripts. It also allows us to do path manipulations in the renderer when the native `path` module is not available.

There are only two cases where native paths are used:

* In APIs that in some way operate with the shell, e.g. `spawn` and `shell.open`.
* In user facing representations, e.g in the browser executable path input.

The biggest change is the introduction of a file system abstraction which proxies the `fs/promises`, `@electron/dialog` and `chokidar` modules, converting paths to posix as needed. 

We also export a path module that uses `pathe` to provide path manipulations with automatic posix normalization. The path module provides a helper function `toNativePath` that lets you convert back to native style paths as needed.

Lastly, the PR also tightens up path comparisons by making sure that they respect file system case-sensitivity (with the caveat that it only assumes the default sensitivity for each platform, custom configurations are not supported).

The next step is where all the fun begins: adding open, save and save as to the file menu.

## How to Test

Test everything that deals with paths:

* Open files
* Run and validate tests
* Import
* Rename files
* Trash files
* etc.

I have tested this on Linux and Window. A test build called v1.13.0-posix is available in the releases.

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Touches most file/path I/O and process-spawn boundaries; incorrect conversion or keying could break file selection, renames, watchers, proxy/k6 execution, or settings persistence across platforms.
> 
> **Overview**
> This PR **standardizes on POSIX-style paths as the internal representation** and centralizes OS-specific conversions at the edges via new `@/utils/path` (now backed by `pathe`) and `@/utils/fs` wrappers.
> 
> It refactors handlers, main-process utilities, and UI/store logic to use the new helpers for path joins/relatives, file dialogs, chokidar watchers, and `fs/promises` access, and introduces case-aware path comparison/keys (`path.key`, `path.equal`) to avoid mismatches on Windows/macOS.
> 
> Settings and user-facing path inputs now convert between internal POSIX storage and native display (`SettingsDialog.utils`), and all shell/spawn call sites (proxy, browser launchers, k6 client, log folder) explicitly convert to native paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 434087bc7c31a8bd3ba6ef0e7fa634e9a36bda26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/components/Settings/SettingsDialog.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+export function settingsToFormValues(settings: AppSettings | undefined) {
+  return settings ? convertPaths(settings, toNativePath) : settings
+}
+
+export function formValuesToSettings(data: AppSettings): AppSettings {
+  return convertPaths(data, toPosixPath)
+}
```
allansson, line 33: Convert paths to and from native for the settings dialog.

src/handlers/ui/index.ts
```diff
@@ -44,47 +43,45 @@ export function initialize() {
 
   ipcMain.handle(UIHandler.TrashFile, async (_, file: StudioFile) => {
     console.info(`${UIHandler.TrashFile} event received`)
-    const filePath = getFilePath(file)
-    return shell.trashItem(filePath)
+
+    return shell.trashItem(path.toNativePath(file.path))
```
allansson, line 47: By using the path from the `StudioFile` instead of `getFilePath` these functions becomes decoupled from our hardcoded project structure.

src/handlers/ui/index.ts
```diff
@@ -44,47 +43,45 @@ export function initialize() {
...
-    const filePath = getFilePath(file)
-    return shell.openPath(filePath)
+
+    return shell.openPath(path.toNativePath(file.path))
   })
 
   ipcMain.handle(UIHandler.GetFiles, async () => {
     console.info(`${UIHandler.GetFiles} event received`)
-    const recordings = (await readdir(RECORDINGS_PATH, { withFileTypes: true }))
+    const recordings = (await readdir(RECORDINGS_PATH))
```
allansson, line 64: I'm not sure when you would ever _not_ want `withFileTypes` so the proxied function will always return `DirEntry[]`.

src/handlers/ui/index.ts
```diff
@@ -105,12 +102,7 @@ export function initialize() {
 
   ipcMain.handle(
     UIHandler.RenameFile,
-    async (
-      e,
-      oldFileName: string,
-      newFileName: string,
-      type: StudioFile['type']
-    ) => {
+    async (e, file: StudioFile, newFileName: string) => {
```
allansson, line 105: Another case of making the handler independent of the project location. This means we could rename a file at any path, which enables features like workspaces.

src/main/deepLinks.ts
```diff
@@ -19,7 +19,7 @@ function registerCustomProtocol() {
   if (process.defaultApp) {
     if (process.argv.length >= 2) {
       app.setAsDefaultProtocolClient(CUSTOM_APP_PROTOCOL, process.execPath, [
-        path.resolve(process.argv[1] ?? ''),
+        path.toNativePath(path.resolve(process.argv[1] ?? '')),
```
allansson, line 22: The path in `argv[1]` can be relative, but has to be an absolute native path when registered. That's why we need to do this little dance with resolving + converting back to native.

src/main/file.ts
```diff
@@ -1,96 +1,59 @@
...
-  BROWSER_TESTS_PATH,
-} from '@/constants/workspace'
 import { StudioFile } from '@/types'
-import { exhaustive } from '@/utils/typescript'
+import * as path from '@/utils/path'
 
 export function getStudioFileFromPath(
   filePath: string
 ): StudioFile | undefined {
+  const parsedPath = path.parse(filePath)
```
allansson, line 13: Simplified to parse the path once and then reuse the result.

src/main/file.ts
```diff
@@ -1,96 +1,59 @@
...
+    displayName: parsedPath.name,
+    fileName: parsedPath.base,
     path: filePath,
   }
 
-  if (
-    filePath.startsWith(RECORDINGS_PATH) &&
-    path.extname(filePath) === '.har'
-  ) {
+  if (parsedPath.ext === '.har') {
```
allansson, line 21: I removed all the path prefix checks here because they seem redundant. By looking at just the extenion, the code becomes more portable and doesn't restrict files to being in any particular directory.

The one caveat is that a `.har` file imported as a data file will still be using the `RecordingPreviewer` for visualization... which seems ok?

src/utils/path.ts
allansson, line 1: Our custom logic has been replaced almost entirely by the `pathe` package.

src/utils/plist.ts
```diff
@@ -1,9 +1,10 @@
 import log from 'electron-log/main'
-import { readFileSync } from 'fs'
 import { parse } from 'plist'
 
-export function getExecutableNameFromPlist(plistPath: string) {
-  const plistContent = readFileSync(plistPath, 'utf-8')
+import { readFile } from './fs'
+
+export async function getExecutableNameFromPlist(plistPath: string) {
+  const plistContent = await readFile(plistPath, 'utf-8')
```
allansson, line 7: Use async for consistency.

src/utils/workspace.ts
```diff
@@ -24,12 +23,10 @@ const REQUIRED_FOLDERS = [
 
 export const setupProjectStructure = async () => {
   for (const folder of REQUIRED_FOLDERS) {
-    if (!existsSync(folder)) {
-      await mkdir(folder)
-    }
+    await mkdir(folder, { recursive: true })
```
allansson, line 26: When using `recursive: true`, `mkdir` is a noop if the directory already exists. This means it's no longer necessary to check if the directory exists.

.eslintrc.cjs
```diff
@@ -76,11 +76,70 @@ module.exports = {
...
+            name: 'node:path',
+            message: "Import from '@/utils/path' instead.",
+          },
+          {
+            name: 'pathe',
+            message: "Import from '@/utils/path' instead.",
+          },
+        ],
+      },
+    ],
```
allansson, line 122: We restrict all the native modules from being used to enforce the usages of our proxies modules instead.

allansson (commented):
Self-review

allansson:
bugbot run

src/utils/fs.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  if (useNativePaths) {
+    return result
+  }
+
+  return {
+    ...result,
+    filePaths: result.filePaths.map(normalize),
+  }
+}
```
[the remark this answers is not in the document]
allansson: @cursor push 9e2234baa3

e-fisher (commented):
Data files don't seem to work, says it's missing when trying to add to an http test. Also, the test fails, but I expect for the same reason. 

Image

Image

Otherwise, everything works as expected!

allansson:
@e-fisher I've fixed the data files warning (I've run bugbot and claude on this PR I don't know how many times and neither of them caught that).

The second issues seems to be that the CSV(?) you used has a column that is quoted incorrectly.

e-fisher (approved):
LGTM 🚀

PR #1250, opened 2026-05-21
feat: Export network traffic from the browser debugger

### LINKED ISSUE #969: Create a protocol test from a browser test
### Feature description

From a given browser test, extract API calls (HTTP at least, ideally ws too) into a new, protocol test following the same logic as an original, just without browser dependency and browser-related logic.

Main use case. Sometimes, there are "heavy" browser tests: people write a browser script with large number of VUs, basically trying to run a load test with browser. Such browsers tests are resource-heavy and may be impossible or at least impractical to execute. Standard recommendation is to rewrite it to protocol test (load testing of APIs should be done without browsers) and use browsers only for UX flow with small number of VUs. This recommendation requires manual effort from a tester now.

k6-studio could help with this recommendation by "automatically" creating a protocol test from a browser one. Then, adjusting VUs to reasonable values for both browser and protocol scripts would help achieve the necessary result: load testing done via protocol test, UX testing done via browser test.

@allansson suggested adding a button that saves recorded requests into a non-browser script, since all the traffic is already captured.

Additionally, IMO, we should assume that people might already have a browser script written somewhere, so this browser script is not necessarily generated by k6-studio. AFAIS, I can load a pre-existing browser.js and have it executed in Studio's Validator, so I assume the same logic as above is applicable.

Related [Slack discussion](https://raintank-corp.slack.com/archives/C06LTHD0708/p1766135334194409).

allansson:
## Description

This PR adds the ability to export the network traffic recorded during a browser debugger session to various formats:

* as a raw recording
* an HTTP test with the network traffic as its recording
* as a script using either an existing generator or an allowlist

https://github.com/user-attachments/assets/df85fe93-7a79-45e5-b33a-8ef51bef6db7

## How to Test

Run a browser test in the debugger and then:

* Export a recording
* Export an HTTP test
* Export as a script using a generator
* Export as a script using an allowlist

## Checklist

## Related PR(s)/Issue(s)

Closes https://github.com/grafana/k6-studio/issues/969

---

> [!NOTE]
> **Medium Risk**
> Changes exported script input data by decoding potentially base64-encoded request/response bodies, which could subtly alter generated rules/assertions or payloads for some recordings. Scope is limited to the browser export flow.
> 
> **Overview**
> Exported scripts now **decode captured request/response `content`** before passing the recording into `generateScript`, ensuring generator rules apply to the real body text rather than base64 strings.
> 
> The export flow also defers request filtering until after the save-path is chosen, avoiding unnecessary processing when the user cancels the dialog.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit baa09671f05f01054acd3aa262a084dfe85f4e14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


allansson:
bugbot run

allansson:
bugbot run

e-fisher (commented):
There's an issue when exporting and applying generator: the correlation and parameterization rules don't get applied because POST body is encoded.

This is a direct export from generator:

Image

And this one from network traffic:

Image


Attaching both scripts:

[scripts.zip](https://github.com/user-attachments/files/28147711/scripts.zip)

e-fisher (approved):
LGTM 🚀

PR #1253, opened 2026-05-25
build: Override @electron/node-gyp with vanilla node-gyp

allansson:
## Description

This PR adds an override from `@electron/node-gyp` to `node-gyp`.

Electron Forge is using an old version of the @electron/rebuild package which has a git dependency on https://github.com/electron/node-gyp. This package has since been deprecated in favor of the vanilla version but Electron Forge probably won't update this package until v8 is released.

This is the only package that we need that uses a git dependency and it has forced us to disable `blockExoticSubDeps` opening us up for supply chain attacks. `pnpm` doesn't have a way of whitelisting specific packages. Luckily the `node-gyp` package should be compatible so we can just override the dependency.

## How to Test

1. Run a test build
2. Test that the built binaries work

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Native rebuilds for Electron still depend on this override until Forge 8; wrong incompatibility would break packaged binaries, though risk is mainly build/runtime on supported platforms.
> 
> **Overview**
> Replaces Electron Forge’s **GitHub tarball** `@electron/node-gyp` with **`node-gyp@12.3.0`** via a pnpm **`overrides`** alias (`'@electron/node-gyp': 'npm:node-gyp@12.3.0'`), so `@electron/rebuild` still resolves the same package name but pulls the published npm build.
> 
> **`blockExoticSubdeps: false`** is removed from `pnpm-workspace.yaml` now that the only exotic (git) dependency is gone, tightening install policy against unvetted git subdeps.
> 
> The lockfile drops the old node-gyp stack (cacache, make-fetch-happen, `encoding@0.1.13` peer wiring on Forge/Sentry, etc.) and wires **`@electron/rebuild`** to **`node-gyp@12.3.0`** instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bddebafba28c284867e1e749322e5df7a66e518e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,6 +10,7 @@ overrides:
   basic-ftp: ^5.2.0
   tar: ^7.5.13
   dompurify: ^3.4.0
+  '@electron/node-gyp': npm:node-gyp@12.3.0
 
 patchedDependencies:
   rrweb: 026a8a870fbc04ea8b1d0048f39e6543189a4a762caf48af27db81ba0ff87d40
@@ -101,7 +102,7 @@ importers:
         version: 5.12.0
       '@sentry/vite-plugin':
         specifier: 2.23.1
-        version: 2.23.1(encoding@0.1.13)
+        version: 2.23.1
       '@tabler/icons-react':
         specifier: 3.42.0
         version: 3.42.0(react@19.2.5)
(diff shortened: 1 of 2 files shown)


allansson (commented):
I've tested this on:

* Windows 11 (x64)
* Linux (.rpm, x64)
* macOS (x64)
* macOS (arm64)

All work.

e-fisher (approved):
LGTM :electron:

PR #1259, opened 2026-05-26
ci: Harden release workflow cleanup and input handling

going-confetti:
## Description

- Change the macOS signing cleanup so that it runs with `if: always()` and explicitly deletes the build keychain - previously, a failing `pnpm run publish` would skip cleanup entirely, leaving `apple_api_key.p8`, `certificate.p12`, and an unlocked keychain on the runner.
- Extract shared steps from `release.yml` and `release-test-version.yml` into a reusable `release-base.yml` to prevent drift in security-sensitive configuration - the cleanup fix above would otherwise have needed to land in two places.
- Add an upfront regex check on `release-test-version.yml`'s `version` input (`^[a-zA-Z0-9._-]{1,20}$`) so malformed values can't land in `package.json` and produce broken release tags or auto-update channel names - the 20-character limit was previously documented only in the input description, not enforced.
- Quote `$APPLE_CERTIFICATE_P12` and use `printf '%s'` instead of `echo` when materializing the cert - guards against whitespace-induced corruption and the stray trailing newline `echo` adds.

## How to Test

- [x] Dispatch `release-test-version.yml` from this branch with `version="bad value"` - the `validate` job should reject it within seconds (confirms version input validation).
- [x] Dispatch `release-test-version.yml` from this branch with a valid version (e.g. `verify-refactor`) - all four matrix jobs (macOS arm64/x64, Windows, Linux) should complete, with a signed pre-release published, and the macOS jobs' `cleanup macos certificates` step shown as run. Delete the pre-release afterward (confirms the reusable-workflow wiring, cert materialization, and cleanup on success).
- [x] On a throwaway branch, inject a `- run: exit 1` step between `setup macos keychain` and `publish macOS arm64` in `release-base.yml`, dispatch, watch the macOS jobs fail, and confirm `cleanup macos certificates` still ran without errors. Revert.

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Touches signing, vault secrets, and release publishing paths; behavior should match prior flows but mistakes could break releases or mishandle certificates.
> 
> **Overview**
> Release CI is consolidated into a reusable **`release-base.yml`** (`workflow_call`), and **`release.yml`** / **`release-test-version.yml`** now delegate to it instead of duplicating the full multi-platform publish matrix.
> 
> **Test releases** gain a **`validate`** job that rejects `version` unless it matches `^[a-zA-Z0-9._-]{1,20}$` before any signing or publish runs. The shared workflow optionally runs `update_version.py` when `version-suffix` is set.
> 
> **macOS signing** hardening in the base workflow: P12 materialization uses **`printf '%s'`** with a quoted cert env var; cleanup runs with **`if: always()`**, removes key material with **`rm -f`**, and **`security delete-keychain build.keychain`** so failed publishes do not leave secrets on the runner.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit be440df0ab2183d6d96dc57b7d7b8e1314b6e1bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/release-base.yml
```diff
@@ -0,0 +1,191 @@
...
+        run: brew install python-setuptools
+
+      - name: install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: setup macos keychain
+        if: startsWith(matrix.platform, 'macos-')
+        run: |
+          echo "$APPLE_API_KEY" > ./apple_api_key.p8
+          printf '%s' "$APPLE_CERTIFICATE_P12" | base64 --decode > certificate.p12
```
going-confetti, line 72: The base64-encoded cert was piped through echo $VAR unquoted - if the secret ever contained whitespace, the decoded cert file could end up corrupted.
`printf '%s' "$VAR"` quotes the value safely and avoids the trailing newline `echo` would otherwise add.

.github/workflows/release-base.yml
```diff
@@ -0,0 +1,191 @@
...
+          # There is a bug on Ubuntu 22.04 (ubuntu-latest) regarding stripping a binary for a different architecture.
+          # https://github.com/electron/forge/issues/3102
+          # https://github.com/electron/forge/issues/3701
+          echo "%__strip /bin/true" >> ~/.rpmmacros
+
+          pnpm run publish
+          pnpm run publish --arch=arm64
+
+      - name: cleanup macos certificates
+        if: always() && startsWith(matrix.platform, 'macos-')
```
going-confetti, line 188: By default, a step is skipped if any earlier step in the job failed - so the previous cleanup was being silently skipped whenever `pnpm publish` errored, leaving the cert files and unlocked keychain on the runner. `if: always()` run cleanup on both success and failure, and `security delete-keychain` now removes the keychain itself (the old version only deleted the cert files)

Llandy3d (approved):
Looks good!

PR #1274, opened 2026-06-04
internal: Add iframe DOM utilities and relocate locator derivation

e-fisher:
## Description

Groundwork for iframe support in the recorder, browser tests, and session replay. First in a stacked split, with no behavior change of its own.

It does two things. First, it moves the locator derivation (`toLocatorOptions`, `getElementLocator`, and the per-type helpers) out of `src/codegen/browser` into `src/utils/locator`, so the recorder event list and the browser test editor stop importing locator logic from the codegen module. Second, it adds the realm-safe DOM helpers the rest of the iframe work builds on: type guards in `src/utils/dom/realm` that classify an element by its tag rather than `instanceof`, since an element inside an iframe lives in that frame's JS realm where a top-frame `instanceof` silently returns `false`; plus a shared iframe-ancestor walk and the locator-resolution and coordinate helpers built on it.

Everything here is covered by unit tests; the later PRs in the stack wire it up.

## How to Test

No behavior to verify by hand. The relocation is a pure move with repointed imports, and the new DOM utilities ship with unit tests.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have added tests for my changes.
- [x] I have commented on my code, particularly in hard-to-understand areas.

## Related PR(s)/Issue(s)

First in the iframe-support stack; the recorder PR builds on this, followed by codegen and replay/editor.

---

> [!NOTE]
> **Low Risk**
> Mostly a pure move and additive utilities with broad unit tests; limited runtime impact until follow-up PRs connect them, though incorrect realm guards could misclassify iframe elements later.
> 
> **Overview**
> **Groundwork for iframe support** with no intended product behavior change: shared utilities and import cleanup that later stack PRs will wire up.
> 
> **Locator derivation** moves from `src/codegen/browser/locator` into `src/utils/locator` (`toLocatorOptions` joins existing helpers). Codegen, the recorder event list, and replay context-menu code import from `@/utils/locator` instead of codegen. `cssLocatorOptions` is added on the locator schema for plain CSS frame locators.
> 
[...]

PR #1275, opened 2026-06-04
feat: Record interactions inside iframes

e-fisher:
## Description

Records DOM interactions that happen inside iframes, including nested and cross-origin ones. The recorder previously captured only the top frame and dropped everything inside iframes, which is a real gap since apps put auth, payment, and embedded widgets in cross-origin iframes.

Each recorded event is now tagged with the chain of iframe locators it happened in, and the event list shows that chain inside the element's selector badge so an interaction inside an iframe is recognizable. To reach cross-origin iframes, the recorder browser forces them into the parent process so the injected script runs in them (it already runs with web security disabled; this browser is authoring-only).

Second PR in the iframe-support stack, building on #1274. Generating `frameLocator` code from these recordings and replaying them come in later PRs.

screencapture 2026-06-04 at 12 42 46@2x

## How to Test

1. Record a page with form in an iframe, you can use this codepen example: https://codepen.io/BaylorRae/full/oNxeBQJ
2. Interact inside the iframe: type inputs, use element picker to add conditions, add assertions by selecting text
3. Stop recording and verify iframe events are recorded

Each interaction captured inside the iframe shows the frame chain in its selector badge (an "in iframe" marker plus the iframe locator), top-frame events look the same as before.

## Checklist

## Related PR(s)/Issue(s)

Builds on https://github.com/grafana/k6-studio/pull/1274

---

> [!NOTE]
> **Medium Risk**
> Touches core recording, CDP browser launch flags, and event ordering across frames; mis-framed paths or sort regressions would corrupt saved recordings, though behavior is heavily tested.
> 
> **Overview**
> **Iframe-aware recording** extends the recorder so interactions inside nested (including cross-origin) iframes are captured, not only the top frame. Each element-targeted event can carry an optional **`frames`** chain (outermost iframe first); the v2 schema adds this on targeted events only. Chrome is launched with **`--disable-site-isolation-trials`** plus existing **`IsolateOrigins` / `site-per-process`** disables so the injected script runs in child frames and can walk **`frameElement`** for paths. **`startRecording`** now runs in every frame; child frames forward inspection/text-selection via **`__K6_STUDIO_*`** bridges while the top frame keeps the UI.
> 
> **Ordering and live lists** stop appending event batches: CDP **`mergeRecordedEvents`**, HAR export, and **`useListenBrowserEvent` / `useRecordedEvents`** replace state with the full timestamp-sorted buffer so late iframe sockets do not scramble the timeline.
> 
> **Highlighting and bounds** thread optional frame locators through IPC/messaging into **`findElementsByFrameChain`**, use cross-realm **`isElement`**, and recompute overlay bounds with iframe offsets and **`collectLayoutShiftWindows`**. Locator badges show **"in &lt;frame&gt;"** (collapsed to **"in N frames"** when deep) and pass frame options on hover.
> 
> **Inspector tools** pick elements and text inside iframes (child-frame detection, top-frame coordinates via **`frameGeometry`**), attach **`frames`** on manual assert/wait-for/text assertions, and skip recording while the top-frame inspector/selection bridge is active.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bdeac39a971e1cde1fe4be97e71cc9aa33d87c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/recorder/browser/messaging/types.ts
+++ b/src/recorder/browser/messaging/types.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod/v4'
 
-import { ElementLocatorSchema } from '@/schemas/locator'
+import { ElementLocatorSchema, LocatorOptionsSchema } from '@/schemas/locator'
 import { BrowserEventSchema } from '@/schemas/recording'
 
 export const InBrowserSettingsSchema = z.object({
@@ -37,6 +37,9 @@ export const EventsRecordedSchema = z.object({
 export const HighlightElementsSchema = z.object({
   type: z.literal('highlight-elements'),
   locator: ElementLocatorSchema.nullable(),
+  // Chain of iframe locators the element lives in, so highlighting can resolve
+  // into nested frames. Absent for top-frame elements.
+  frames: LocatorOptionsSchema.array().optional(),
 })
 
 export const NavigateSchema = z.object({
--- a/src/recorder/browser/view/EventDrawer.tsx
+++ b/src/recorder/browser/view/EventDrawer.tsx
@@ -4,7 +4,7 @@ import { XIcon } from 'lucide-react'
 
 import { BrowserEventList } from '@/components/BrowserEventList'
 import { useContainerElement } from '@/components/primitives/ContainerProvider'
-import { ElementLocator } from '@/schemas/locator'
+import { ElementLocator, LocatorOptions } from '@/schemas/locator'
 import { BrowserEvent } from '@/schemas/recording'
 import { RecordingContext } from '@/views/Recorder/RecordingContext'
 
@@ -42,10 +42,14 @@ export function EventDrawer({ open, events, onOpenChange }: EventDrawerProps) {
   const container = useContainerElement()
   const [settings, setSettings] = useInBrowserSettings()
 
-  const handleHighlight = (locator: ElementLocator | null) => {
+  const handleHighlight = (
+    locator: ElementLocator | null,
+    frames?: LocatorOptions[]
+  ) => {
     client.send({
       type: 'highlight-elements',
       locator,
+      frames,
     })
   }
 
--- a/src/recorder/launchers/cdp/index.ts
+++ b/src/recorder/launchers/cdp/index.ts
@@ -2,7 +2,7 @@ import { ChildProcess, spawn } from 'child_process'
 import logger from 'electron-log/main'
 
 import { BrowserServer } from '@/recorder/server'
-import { ElementLocator } from '@/schemas/locator'
+import { ElementLocator, LocatorOptions } from '@/schemas/locator'
 import { BrowserEvent } from '@/schemas/recording'
 import { ChromeDevToolsClient, Transport } from '@/utils/cdp/client'
 import { PipeTransport } from '@/utils/cdp/transports/pipe'
@@ -20,6 +20,7 @@ import {
 import { getBrowserLaunchArgs } from '../utils'
 
 import { BrowserSession } from './browser'
+import { mergeRecordedEvents } from './events'
 import { Script } from './script'
 
 type InitState = 'init' | 'spawned' | 'ready'
@@ -31,6 +32,12 @@ const BROWSER_CDP_ARGS = [
   // Disable web security to allow our script to be executed in sandboxed iframes.
   '--disable-web-security',
   '--allow-running-insecure-content',
+  // Keep cross-origin iframes in the parent's process so the injected recording
+  // script runs inside them and can walk up to the top frame. Disabling the
+  // IsolateOrigins/site-per-process features (see FEATURES_TO_DISABLE) is not
+  // enough on its own in current Chrome; the site-isolation trials must also be
+  // turned off.
+  '--disable-site-isolation-trials',
 ]
 
 const BROWSER_CDP_WEBSOCKET_ARGS = [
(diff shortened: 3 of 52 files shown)


src/recorder/browser/messaging/types.ts
```diff
@@ -37,6 +37,9 @@ export const EventsRecordedSchema = z.object({
 export const HighlightElementsSchema = z.object({
   type: z.literal('highlight-elements'),
   locator: ElementLocatorSchema.nullable(),
+  // Chain of iframe locators the element lives in, so highlighting can resolve
+  // into nested frames. Absent for top-frame elements.
```
allansson, line 41: Wouldn't an empty array signal that it's the top-level frame? 🤔

src/recorder/browser/view/EventDrawer.tsx
```diff
@@ -42,10 +42,14 @@ export function EventDrawer({ open, events, onOpenChange }: EventDrawerProps) {
   const container = useContainerElement()
   const [settings, setSettings] = useInBrowserSettings()
 
-  const handleHighlight = (locator: ElementLocator | null) => {
+  const handleHighlight = (
+    locator: ElementLocator | null,
+    frames?: LocatorOptions[]
```
allansson, line 47: Just thinking out loud here:

```ts
// Old ElementLocator
type LocatorDefinition = CssLocator | RoleLocator | // etc.

// Old LocatorOptions
interface ElementLocator {
  current: LocatorDefintion["type"]
  values: {
    css?: CssLocator
    role?: RoleLocator
  }
  parentFrame?: ElementLocator
}
```

Basically, ElementLocator is a linked list where the first element is the target element.

An alternative is:

```ts
type LocatorDefinition = CssLocator | RoleLocator | // etc.

// Old LocatorOptions
interface FrameLocator {
  current: LocatorDefintion["type"]
  values: {
    css?: CssLocator
    role?: RoleLocator
  }
}

interface ElementLocator extends FrameLocator {
  frames: FrameLocator[]
}
```

The point being that everything that locates an element is stored in the same data structure.

src/schemas/recording/browser/v2/index.ts
```diff
@@ -31,6 +31,14 @@ const BrowserEventBaseSchema = z.object({
   timestamp: z.number(),
 })
 
+// Events that target an element also record the chain of iframe elements from
+// the top frame down to the frame the element lives in, outermost first. Absent
+// or empty means the top frame. Page-level events (navigation, reload) have no
+// element target and so no frame scope; they extend the base directly.
+const TargetedEventBaseSchema = BrowserEventBaseSchema.extend({
+  frames: BrowserEventTargetSchema.array().optional(),
```
allansson, line 39: Note to self: I would love to refactor the `BrowserEventTargetSchema` to use the same `LocatorOptionsSchema` as the rest of the app before release.

src/recorder/launchers/cdp/index.ts
```diff
@@ -31,6 +32,12 @@ const BROWSER_CDP_ARGS = [
   // Disable web security to allow our script to be executed in sandboxed iframes.
   '--disable-web-security',
   '--allow-running-insecure-content',
+  // Keep cross-origin iframes in the parent's process so the injected recording
+  // script runs inside them and can walk up to the top frame. Disabling the
+  // IsolateOrigins/site-per-process features (see FEATURES_TO_DISABLE) is not
+  // enough on its own in current Chrome; the site-isolation trials must also be
+  // turned off.
+  '--disable-site-isolation-trials',
```
allansson, line 40: This must be the reason I couldn't get it to work in my attempts. And Claude swore that it wasn't possible to disable this. 😏

src/recorder/launchers/cdp/index.ts
```diff
@@ -124,11 +134,14 @@ class CDPRecordingSession
   }
 
   #record(events: BrowserEvent[]) {
-    this.#events.push(...events)
+    // Frames record over separate WebSocket connections, so events can arrive
+    // out of order. Keep the buffer sorted by timestamp so the saved recording
+    // and generated test reflect the real interaction order.
+    this.#events = mergeRecordedEvents(this.#events, events)
```
allansson, line 140: In my attempt, I bubbled events up using `window.postMessage`, collecting frame locators along the way. When it reached the top-level frame it sent the event to the server. It solves the ordering issues.

That said, this works just as good so we'll keep it.

allansson (commented):
Really great work! 😍

Left some food for thought, but I don't think it's something we need to address (at least not in this PR).

A suggestion on the UI:

Keep the target element in focus, separate frames with `in` or `inside`:

Image

Shorthand when frame hierarchy is very deep:

Image

src/recorder/browser/messaging/types.ts
e-fisher: Empty array would signal it too, page events have it as undefined, adding frames as optional allowed a bit fewer changes. View layer covers this with frames ?? []

src/recorder/browser/view/EventDrawer.tsx
e-fisher: `submit-form` has two targets (form + submitter) sharing the one event-level frames. This would duplicate that chain across both targets.

e-fisher:
bugbot run

e-fisher:
@cursor push cabcb711d5

e-fisher:
bugbot run

allansson (approved):
🥳

PR #1277, opened 2026-06-04
refactor: Move proxy data parsing to main process

allansson:
## Description

This PR moves the conversion from HAR to ProxyData to the main process. I was hoping that it'd improve the performance of loading large recordings (and to some degree it has) but it didn't have as much impact as I hoped.

The changes make sense though. There's no point in doing this processing in the renderer and if we can unblock the renderer even a few ms that's a win.

The big culprit is the full-text search indexing where we parse the body of all requests and then throw them into Fuse.js. My proposed solution to this would be to:

1. Do the parsing of bodies once when loading the file.
2. Use [`FuseWorker`](https://www.fusejs.io/web-workers.html) for indexing and search

So this PR could be seen as one step closer to doing that.

## How to Test

Open recordings in both the previewer and a generator. Both should work fine.

## Checklist

## Related PR(s)/Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Changes the IPC contract for recording files and shifts parsing to the main process; regressions would affect every recording load in generator and previewer, though scope is localized to deserialization and consumers.
> 
> **Overview**
> **Recording open path** now parses HAR into **`ProxyData[]`** in the main process (`deserializeContent`) instead of in the renderer. **`RecordingContent`** returns **`data: ProxyData[]`** plus a top-level **`browserEvents`** array, rather than the raw **`Recording`** schema.
> 
> The **Generator** (`loadHarFile`, **`RecordingSelector`**) and **Recording previewer** no longer call **`harToProxyData`**; they use **`content.data`** and **`content.browserEvents`** from **`openFile`**. **`harToProxyData`** switches from **`self.crypto`** to **`crypto`** for UUID generation where it still runs on the main side.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 023ae74d0933320710d93d1af90f96185fa4af55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/handlers/fs/serialization.ts
+++ b/src/handlers/fs/serialization.ts
@@ -10,6 +10,7 @@ import { RecordingSchema } from '@/schemas/recording'
 import { StudioFileType } from '@/types'
 import { DataFilePreview } from '@/types/testData'
 import { parseDataFile } from '@/utils/dataFile'
+import { harToProxyData } from '@/utils/harToProxyData'
 import { K6Client } from '@/utils/k6/client'
 import * as path from '@/utils/path'
 import {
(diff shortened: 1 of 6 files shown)


allansson:
bugbot run

e-fisher (approved):
🙌

e-fisher:
FuseWorker looks promising!

PR #1290, opened 2026-06-10
internal: Add cross-OS E2E smoke tests

e-fisher:
## Description

Adds E2E smoke tests that package the real app with `electron-forge package` and launch it on macOS (Apple Silicon + Intel), Windows, and Linux, asserting the Home view renders. Nothing in CI currently builds the packaged app, so per-OS packaging, startup, and signing regressions only surface at release. The tests run via a new `smoke.yml` workflow, one native-arch leg per runner, triggered manually (`workflow_dispatch`), on PRs that touch packaging/startup/native files, and on release-please PRs.

Getting a creds-free build to launch needed a few changes: signing and notarization are gated to release, and creds-free macOS builds get a deep ad-hoc signature in a `postPackage` hook (Apple Silicon kills an invalidly-signed app on launch, and the fuses plugin re-signs the binary after packaging). This also fixes a latent crash where `SENTRY_DSN` was left undefined in the bundle when built without Sentry credentials.

Playwright drives the packaged app over Chromium's remote debugging port rather than `_electron.launch`, since the production fuses disable the Node inspector.

## How to Test

Locally, against your runner's native arch:

1. `pnpm package`
2. `pnpm test:e2e`

The app should package, launch, and the test pass. In CI, all four legs run automatically on this PR (it touches `forge.config.ts`, `e2e/**`, and the workflow); verify they go green.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -1,0 +1,63 @@
+name: E2E Smoke
+
+# Packages the real artifact and launches it on every supported OS to catch
+# build/startup regressions release builds would miss.
+on:
+  workflow_dispatch:
+  # TEMPORARY: runs the 4-OS matrix on this PR for first-time verification.
+  # Remove this trigger before merging -- workflow_dispatch is the intended
+  # long-term entry point.
+  pull_request:
+    branches:
+      - main
+
+permissions: {}
+
+jobs:
+  smoke:
+    permissions:
+      contents: read
+    strategy:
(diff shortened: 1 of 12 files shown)


allansson (approved):
Workflow took ~7 mins when I tested it. I think that's roughly what bugbot takes as well. It's also a good start towards writing tests for critical paths through the application. 💯

PR #1308, opened 2026-06-26
feat: Add an Assistant-guided setup wizard for HTTP tests

e-fisher:
## Description

Adds an Assistant-guided **Test Configuration Wizard** for HTTP tests. Instead of dropping a recording straight into the generator, the wizard walks through five steps, Select hosts, Autocorrelation, Parameterization, Thresholds, and Run test, each backed by the Grafana Assistant, which analyzes the recording and proposes configuration the user reviews and adjusts. The final Run test step summarizes the plan (hosts, thresholds, an estimated VU-hours cost, and a load-stage timeline) and launches the test in Grafana Cloud.

The Assistant steps drive renderer-owned tool definitions over the A2A transport; launch reuses the existing `RunInCloudDialog`, and the VU-hours estimate comes from the cloud `validate_options` endpoint. The wizard opens on guided setup from a new recording, or via "Configure with Assistant" on an existing generator, and requires Grafana Cloud sign-in.

screencapture 2026-07-14 at 11 06 40@2x

screencapture 2026-07-14 at 10 56 13@2x

screencapture 2026-07-14 at 10 55 58@2x

## How to Test

1. Sign in to Grafana Cloud. Record a session (or open a HAR), create an HTTP test, and choose **Configure with Assistant**.
2. Step through Select hosts → Autocorrelation → Parameterization → Thresholds, reviewing the Assistant's suggestions at each (toggle/edit rules, adjust thresholds). Navigate back and forth and re-run a step to confirm state holds.
3. On **Run test**, confirm the plan summary (requests, thresholds, ~VUh, stage timeline), then **Save & Run** to launch in Grafana Cloud. Try a too-large VU count or duration and confirm the VUh line shows the project-limit message.

## Checklist


src/utils/assistant/useAssistantAgent.ts
```diff
@@ -0,0 +1,201 @@
...
+        output: output as never,
+      })
+    },
+  })
+
+  const syncMessagesToLog = actionsLog.syncFromMessages
+
+  useEffect(() => {
+    syncMessagesToLog(messages, statusRef.current === 'running')
+  }, [messages, syncMessagesToLog])
```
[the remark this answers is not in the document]
e-fisher: Expected

e-fisher:
@cursor push 4f2a067b03

e-fisher:
bugbot run

src/views/Generator/SetupWizard/steps/AutocorrelationStep/AutocorrelationStep.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  const handleStatusChange = (status: CorrelationStatus) => {
+    if (status === 'not-started' || stepState.status === 'running') {
+      return
+    }
+
+    if (!TERMINAL_STATUSES.includes(status)) {
+      dispatch({ type: 'stepRunStarted', stepId: 'autocorrelation' })
+    }
+  }
```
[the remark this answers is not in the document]
e-fisher: AutoCorrelation renders its own error state

e-fisher:
bugbot run

e-fisher:
bugbot run

src/views/Generator/SetupWizard/steps/AutocorrelationStep/AutocorrelationStep.tsx
```diff
@@ -0,0 +1,341 @@
...
+    dispatch({ type: 'stepRunReset', stepId: 'autocorrelation' })
+  }
+
+  if (stepState.status === 'completed') {
+    return <CompletedStep />
+  }
+
+  if (proxyStatus !== 'online') {
+    return <ProxyOfflineStep onSkip={handleSkip} />
+  }
```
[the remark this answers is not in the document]
e-fisher: That's fine

e-fisher:
bugbot run

src/views/Generator/SetupWizard/steps/useStepAgentLifecycle.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (state.steps[stepId].status !== 'running') {
+      return
+    }
+
+    if (status === 'completed') {
+      // A finished agent step means the wizard actually configured something;
+      // opening and closing the wizard, skips, and failures do not count. The
+      // flag persists in the generator file and script header for reporting.
+      setWizardUsed(true)
+      onCompleted()
```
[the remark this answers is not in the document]
e-fisher: @cursor push d9e208ec89

e-fisher:
bugbot run

src/components/Assistant/useActionsLog.ts
```diff
@@ -22,7 +25,7 @@ export function useActionsLog() {
   }
 
   const syncFromMessages = useCallback(function syncFromMessages(
-    messages: Message[],
+    messages: UIMessage[],
```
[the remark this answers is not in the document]
e-fisher: Expected, we don't show any messages after finish tool call

e-fisher:
bugbot run

src/views/Generator/GeneratorControls/GeneratorControls.tsx
```diff
@@ -82,6 +85,13 @@ export function GeneratorControls({
         >
           <CircleCheckBigIcon /> Validate
         </ButtonWithTooltip>
+        <ButtonWithTooltip
```
going-confetti, line 142: This is likely an old problem made apparent by adding another button, but the buttons no longer fit the window when the window is small

Image

src/views/Generator/SetupWizard/Stepper.tsx
```diff
@@ -0,0 +1,139 @@
...
+  activeStep: WizardStep,
+  isCompleted: boolean
+): StepDisplayState {
+  if (isCompleted) return 'done'
+  if (stepId === activeStep) return 'active'
+
+  return 'todo'
+}
+
+const circleStyles: Record<StepDisplayState, Record<string, string>> = {
```
going-confetti: Nit: `Record<string, string>` seems pointless

src/views/Generator/SetupWizard/Stepper.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  alignItems: 'center',
+                  gap: 9,
+                  flexShrink: 0,
+                  cursor: isClickable ? 'pointer' : 'default',
+                }}
+              >
+                <StepCircle displayState={displayState} number={index + 1} />
+                <Text
+                  size="1"
+                  weight={stepId === state.activeStep ? 'bold' : 'medium'}
```
going-confetti: Nit: using different font weight causes small but noticeable layout shift when moving between steps

https://github.com/user-attachments/assets/bdb94ffe-f196-4fff-949b-3754518c5387

src/components/Assistant/AssistantAuthGate.tsx
```diff
@@ -0,0 +1,220 @@
...
+            setIsCloudSigningIn(false)
+            void invalidateAssistantAuthStatus()
+          }}
+          onAbort={() => setIsCloudSigningIn(false)}
+        />
+      </GateLayout>
+    )
+  }
+
+  if (!isSignedIn) {
```
going-confetti, line 63: Nit: can we handle this on the choice screen instead? I find the experience of seeing this extra step quite frustrating if you're not logged in

Image
Image

src/components/Assistant/AssistantAuthGate.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  if (isAwaitingApproval) {
+    return (
+      <GateLayout>
+        <GateSpinner label="Waiting for approval" />
+        {signIn.verificationCode !== null && (
+          <Text size="2" color="gray">
+            Verification code{' '}
+            <Text weight="bold">{signIn.verificationCode}</Text>
```
going-confetti, line 101: Nit: IMO this needs some additional copy and perhaps better presentation so that it's clearer what's going on if/when the user accidentally switches back to k6 Studio before completing the process
Image

src/views/Generator/SetupWizard/steps/ThresholdsStep/ThresholdsStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+            hideRemove
+          />
+        </Flex>
+      </StepFrame>
+      <WizardFooter canContinue onBack={goBack} onContinue={goNext} />
+    </>
+  )
+}
+
+export function ThresholdsStep() {
```
going-confetti, line 95: You can add new thresholds, but there's no way to remove them (only disable, which essentially means removing them from the generator config, but doesn't look like it to the user):
Image

Additionally, the default value for the default metric doesn't make sense (most likely a preexisting issue) - `response time < 0ms`

src/views/Generator/SetupWizard/steps/ParameterizationStep/ParamRow.tsx
```diff
@@ -0,0 +1,231 @@
...
+    <SuggestionRow
+      isLast={isLast}
+      dimmed={!rule.enabled}
+      icon={<Icon size={14} />}
+      name={meta.field}
+      secondary={
+        <>
+          <MethodBadge method={meta.location.method}>
+            {meta.location.method}
+          </MethodBadge>
```
going-confetti, line 198: IMO this is quite confusing for people no super familiar with the methods/icons. As someone who hasn't worked on k6 Studio for a couple of months now, I spent a few moments to understand what those were and even then I wasn't 100% sure. Do we even need to surface the method to the user considering they have no way to change it?

Image

src/views/Generator/SetupWizard/Stepper.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { WIZARD_STEPS, WizardStep } from './state/types'
+
+type StepDisplayState = 'done' | 'active' | 'todo'
+
+function getStepDisplayState(
+  stepId: WizardStep,
+  activeStep: WizardStep,
+  isCompleted: boolean
+): StepDisplayState {
+  if (isCompleted) return 'done'
```
going-confetti: Have you thought about adding the skipped state? Feels a bit unintuitive to have a step marked as "completed" when you've just skipped it.

src/views/Generator/SetupWizard/Stepper.tsx
```diff
@@ -0,0 +1,139 @@
...
+      flexShrink="0"
+      align="center"
+      css={{
+        height: 64,
+        borderBottom: '1px solid var(--gray-4)',
+        backgroundColor: 'var(--gray-2)',
+        padding: '0 24px',
+      }}
+    >
+      <Flex align="center" width="100%" maxWidth="860px" mx="auto">
```
going-confetti, line 110: The min width of the whole window is 800px, so it should be supported unless we decide to increase that number

Image

src/views/Generator/SetupWizard/SetupWizard.tsx
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  return (
+    <Badge color="orange" variant="soft" size="1">
+      Experimental
+    </Badge>
+  )
+}
+
+function StepIndicator() {
```
going-confetti: IMO this component is redundant: we only have 5 steps, and their state is clearly communicated to the user.

src/views/Generator/SetupWizard/SetupWizard.tsx
```diff
@@ -0,0 +1,183 @@
...
+        <>
+          <StepIndicator />
+          <Button
+            variant="ghost"
+            color="gray"
+            size="1"
+            onClick={handleClose}
+            css={{ margin: 0 }}
+          >
+            Close
```
going-confetti: This purpose of this button is a bit confusing to me: it makes perfect sense if the user landed here by using the "Configure with Assistant" button, but if they've used the new test menu/choice screen, wouldn't it be expected to go back to the previous screen instead? 

It's not a perfect solution, but I would suggest renaming it to something like "Open in Generator" instead, but hopefully someone has better ideas.

src/views/Generator/SetupWizard/ChoiceScreen.tsx
```diff
@@ -0,0 +1,160 @@
...
+      flexGrow="1"
+      p="6"
+      overflowY="auto"
+    >
+      <Flex direction="column" align="center" gap="2" mb="6" maxWidth="560px">
+        <Heading size="6" align="center">
+          How do you want to configure this test?
+        </Heading>
+        <Text size="2" color="gray" align="center">
+          Turn your recording into a load test. Let the Grafana Assistant guide
```
going-confetti: Nit
```suggestion
          Turn your recording into a load test. Let Grafana Assistant guide
```

going-confetti: Might also be worth linking the assistant docs here for those unfamiliar with it.

src/views/Generator/SetupWizard/ChoiceScreen.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <Button
+              size="3"
+              css={{ width: '100%' }}
+              onClick={onStartGuidedSetup}
+            >
+              Start guided setup <ArrowRightIcon size={17} />
+            </Button>
+          }
+        />
+        <ChoiceCard
```
going-confetti, line 138: `Familiar generator workspace` only applies to a subset of the existing users. 
Image

src/views/Generator/SetupWizard/ChoiceScreen.tsx
```diff
@@ -0,0 +1,160 @@
...
+          accent
+          badge="Experimental"
+          visual={
+            <img src={grotMagic} alt="" css={{ height: 64, width: 'auto' }} />
+          }
+          title="Configure with Assistant"
+          description="The Assistant analyzes your recording and proposes the hosts, rules and thresholds your test needs."
+          bullets={[
+            'Suggests which hosts to include',
```
going-confetti, line 123: Nit: technically, we do that in Generator as well by preselecting one of the options

src/views/Generator/SetupWizard/constants.ts
```diff
@@ -0,0 +1,61 @@
...
+  icon: LucideIcon
+}
+
+export const STEP_CONFIG: Record<WizardStep, StepConfig> = {
+  hosts: {
+    label: 'Hosts',
+    hint: 'Choose what to test',
+    title: 'Select hosts',
+    description:
+      'The Assistant identified which hosts carry the load you care about. Include or exclude any before continuing.',
```
going-confetti: This is also shown while the process hasn't finished yet, which is a bit confusing

Image

The same applies to other descriptions that mention the result of the step regardless of whether the process is finished. Another example is when the recording doesn't need anything parameterized or correlated - "created rules" won't make sense in such cases.

src/views/Generator/SetupWizard/steps/ParameterizationStep/ParamRow.tsx
going-confetti: There's also something wrong with the underscore in `admin_user` and `admin_password` here.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,11 @@ jobs:
         with:
           persist-credentials: false
       - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+        with:
+          # Pinned like release-base.yml: without it the runner's pnpm
+          # self-updates to the newest 11.x, whose installer can crash on
+          # freshly published versions.
+          version: 11.5.2
```
Llandy3d: On main release-base is on `11.8.0`, this seems to be a "regression"

Llandy3d (changes requested):

going-confetti (changes requested):
I've added some comments, including nitpicks and personal opinions - would love to hear your though about them.

The PR/feature is quite big, so I've only done the first iteration so far, focusing on the UI. I also encountered a bug during the autocorrelation step when it wasn't able to fix the correlation issues, which eventually led to an error:

Image

Image

I had a suspicion it was caused by the wizard configuration from before - and turns out that was the case. The wizard/assistant added multiple parameterization rules that were incorrect - the quickpizza backend expects `maxCaloriesPerSlice`, `minNumberOfToppings`, and `maxNumberOfToppings` to be integer, but test variables are always strings.

I'm not sure how what would be the best fix for this, but maybe for now it's worth adjusting the prompt to skip parameterization if the value is non-string.

src/views/Generator/SetupWizard/steps/ThresholdsStep/ThresholdsStep.tsx
e-fisher: I was debating between "removing" or "disabling" generated artifacts and wanted it be consistent between steps. I went with toggles because it's non-destructive and user can always toggle it back on after accidental click. I see your point though, this step allows creating new thresholds while correl/param only shows what Assistant suggested. Do you think it would make sense witch to remove action on other steps as well?

going-confetti: Good question. I _think_ it's fine to keep the toggle for auto-generated rules/thresholds (as mistakenly removing them would mean the user needs to re-run the step) and add the remove button for thresholds added manually.

src/views/Generator/SetupWizard/steps/ParameterizationStep/ParamRow.tsx
e-fisher: I've included method for cases when recording might have same routes with different methods, e.g. `GET /user/1` and `PATCH /user/1`, without method those 2 requests would look identical in this list. 

Agree on icons though, not very intuitive, should we just drop them?

As for missing underscore - assistant generates those names, just to have user-friendly description of the rule rather than showing actual selectors. You already had username and password so it prepended next two with "admin "

src/components/Assistant/AssistantAuthGate.tsx
e-fisher: I've had the login gate in the choice screen initially, but this needed for running wizard on existing generators, they skip choice screen

src/views/Generator/GeneratorControls/GeneratorControls.tsx
e-fisher: I've added logic to collapse those button into icons when there's not enough space and bumped min window size to 1000. There are other parts of the app that still overflow when app window at minimum width and sidebar is at max width - but it might require additional work to autohide sidebar and I'd rather do it in a separate PR

screencapture 2026-07-16 at 10 40 11@2x

src/views/Generator/SetupWizard/Stepper.tsx
e-fisher: Good call, added skipped state, additionally stepper will indicate skipped steps

src/views/Generator/SetupWizard/SetupWizard.tsx
e-fisher: I've kept "Close" but gave it a tooltip that spells out where it lands and that finished-step changes are kept. Going back to the choice screen is already covered by Back on the first step, so close stays as an "exit" for both paths. WDYT?

screencapture 2026-07-16 at 10 56 08@2x

going-confetti: I think it's just not an established pattern in this app, that's why the outcome of clicking "Close" doesn't match my personal expectations. So while not ideal, I don't think it's critical.

The choice screen is a special case, because the user hasn't made a choice at that point so:
1. The tooltip shouldn't say "Close guided setup"
2. It feels weird to land in generator and not in recording, but maybe it can be less confusing with better tooltip/label copy?

image

src/views/Generator/GeneratorControls/GeneratorControls.tsx
```diff
@@ -70,25 +115,46 @@ export function GeneratorControls({
...
         </ButtonWithTooltip>
-        <RunInCloudButton
+        <ButtonWithTooltip
+          variant="ghost"
+          aria-label="Configure with Assistant"
+          tooltip={actionsCollapsed ? 'Configure with Assistant' : ''}
+          onClick={onConfigureWithAssistant}
+        >
+          <WandSparklesIcon />
+          <span css={actionLabelStyles}>Configure with Assistant</span>
```
going-confetti, line 154: Minor: this is enabled in empty generators, but the assistant cannot really do anything when there's no recording, so I think it should be disabled.

There's a related issue where you can still start a cloud run even if there's no requests dispatched by the script:

Image

src/views/Generator/SetupWizard/Stepper.tsx
going-confetti, line 1: There's an edge-case bug that happens when you re-run one of the already skipped/completed steps and switch to a different step before it finishes:

https://github.com/user-attachments/assets/0151a167-3fff-46fa-831a-5eae7706fb4e

The last step then becomes inaccessible

going-confetti (changes requested):
Looks pretty good. I found a bug (let me know if you cannot reproduce it) and left a few comments.

I also think it's a bit weird to see the word "generator" when navigating away from an unfinished wizard, but it's not critical:
Image

The guided experience sometime has issues working with existing generators, presumingly because it doesn't take existing rules into account (?). For example, I faced an issue with Autocorrelation not being able to do anything meaningful for quite a while with an old-ish misconfigured Generator that I had saved (I'll share it with you):

Image

Image

src/views/Generator/SetupWizard/state/reducer.ts
```diff
@@ -0,0 +1,176 @@
...
+  const donePrefixLength = WIZARD_STEPS.findIndex(
+    (step) => !isStepDone(state.steps[step])
+  )
+
+  if (donePrefixLength === -1) {
+    return true
+  }
+
+  return index <= donePrefixLength
+}
```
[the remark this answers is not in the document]
e-fisher: not reproducible

src/views/Generator/SetupWizard/steps/ParameterizationStep/useParameterizationAgent.ts
```diff
@@ -0,0 +1,188 @@
...
+          ? 'No values need parameterization'
+          : `Suggested ${proposals.length} parameterization rule${proposals.length === 1 ? '' : 's'}`,
+    })
+  }
+
+  // Re-running the step withdraws the previously committed rules and the
+  // variables this run introduced before starting a fresh analysis.
+  function cleanupCommittedProposals() {
+    withdrawStepArtifacts(stepState)
+  }
```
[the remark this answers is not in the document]
e-fisher: Expected behavior

src/views/Generator/SetupWizard/steps/AutocorrelationStep/AutocorrelationStep.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  const { entries } = stepState.result
+
+  // Re-running withdraws the rules accepted by the previous run; the fresh
+  // mount of AutoCorrelation auto-starts a new analysis.
+  const handleRerun = () => {
+    withdrawStepArtifacts(stepState)
+    dispatch({ type: 'stepRunReset', stepId: 'autocorrelation' })
+  }
```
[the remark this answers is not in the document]
e-fisher: Expected

going-confetti (approved):
LGTM

Llandy3d (approved):

PR #1330, opened 2026-07-23
fix: Unicode payloads become corrupted after stopping a recording

### LINKED ISSUE #1322: [k6 Studio] Encoding issue / Unicode corruption in HTTP payload upon stopping recorder
### Describe the bug

When using the built-in k6 Studio Recorder (via the proxy/extension) to capture HTTP traffic that contains non-ASCII characters (e.g., Vietnamese Unicode characters like Nguyễn Văn A or kiểm thử), the payload is correctly captured during the active recording phase.However, the moment the "Stop Recording" button is clicked, k6 Studio corrupts the Unicode encoding while finalizing and generating the script. The Vietnamese text in the HTTP request payload is replaced by corrupted character strings (e.g., converted to mojibake like Ã¡, đ»¥, or raw hex replacement blocks like \uFFFD)

1. Open k6 Studio and start a new recording session via the built-in recorder.
2. Trigger an HTTP POST/PUT request on a target website that includes Vietnamese Unicode characters in the payload body (e.g., {"name": "Nguyễn Văn A"}).
3. Verify that the request goes through successfully on the application side.
4. Click the "Stop Recording" button in k6 Studio.
5. Inspect the generated request payload inside the newly created visual flow or script editor. The Unicode characters are now corrupted/broken.

### Expected behavior

Upon stopping the recording, k6 Studio should accurately preserve the UTF-8/Unicode encoding of the captured HTTP payload in the generated script or flow.

### Actual behavior

Exporting a .har file directly from the Google Chrome/Edge Network tab and importing that HAR file into k6 Studio preserves the Unicode payload perfectly. This confirms that the bug lies specifically within k6 Studio's proxy recorder or its file-finalization encoding mechanism when stopping the record session.

### Grafana k6 Studio version

1.13.1

### OS

Windows 11, Ubuntu 26

Lileeid:
## Description

Fix Unicode corruption in recorded HTTP request payloads.

Request bodies were decoded with `atob()` during HAR serialization, which corrupted multi-byte UTF-8 characters. This change uses the existing UTF-8-safe decoder and adds a regression test for mixed Unicode payloads.

## How to Test

Run:

```bash
pnpm test -- --run src/utils/proxyDataToHar.test.ts
```

### Manually

- Record a POST request with Unicode characters in the payload.
- Stop the recording.
- Reopen the request and verify that the payload is unchanged.

## Checklist

## Related PR(s)/Issue(s)

Resolves #1322

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -30,8 +30,14 @@ function atobUTF8(base64Str: string) {
     bytes[i] = binaryStr.charCodeAt(i)
   }
 
-  // decode bytes as UTF-8
-  return new TextDecoder('utf-8').decode(bytes)
+  try {
+    return new TextDecoder('utf-8', {
+      fatal: true,
+      ignoreBOM: true,
+    }).decode(bytes)
+  } catch {
+    return binaryStr
+  }
 }
 
 export function safeBtoa(content: string) {
--- a/src/utils/proxyDataToHar.test.ts
+++ b/src/utils/proxyDataToHar.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from 'vitest'
 
 import { BrowserEvent } from '@/schemas/recording'
+import {
+  createProxyDataWithoutResponse,
+  createRequest,
+} from '@/test/factories/proxyData'
+import { safeBtoa } from '@/utils/format'
 
 import { proxyDataToHar } from './proxyDataToHar'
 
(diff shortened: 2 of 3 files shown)


e-fisher (commented):
@Lileeid we've found few issues in the PR:

- Binary request corruption: An `application/octet-stream` payload containing `00 C3 A9 0A` replays correctly on main, but becomes `00 E9 0A` with this PR. The UTF-8 decoder converts `C3 A9` to `é`, which `safeBtoa()` then encodes as a single byte.
- JSON response previews disappear: A UTF-8 BOM-prefixed JSON response displays correctly on main, but shows “No content” with this PR. `ignoreBOM: true` preserves the BOM, causing the preview’s `JSON.parse()` to fail.

The original Unicode request fix works. Request decoding needs to preserve binary bytes without changing the shared response-preview decoder’s behavior.

PR #1331, opened 2026-07-25
feat: Add custom browser launch arguments to Recorder

### LINKED ISSUE #852: Be able to add custom Chrome flags.
#### What ?

Add an option in the Recorder tab to configure custom browser flags when launching Chrome from K6 Studio.

#### Why ?

At work we use a large monorepo application where both frontend and backend run on localhost (e.g. `localhost:3000` for frontend, `localhost:5000` for backend).

Currently, as noted in the docs, K6 Proxy cannot intercept requests made to localhost. A common workaround is to replace every localhost occurrence in the codebase with an alias (like app.local), but this is not feasible for us as it would require modifying application files every time we want to run a recording (and other issues related to auth we have when doing so).

Instead, I found a working solution:

- Enable manual proxy in Windows with the K6 proxy and add `<-loopback>` to the exception list.
- Launch Chrome with a shortcut that includes K6’s standard flags plus the extra `--proxy-bypass-list=<-loopback>` flag. This is the final shortcut I have:

```
"C:\Program Files\Google\Chrome\Application\chrome.exe" 
  --new 
  --args 
  --proxy-bypass-list=<-loopback> 
  --user-data-dir="C:\Users\Polyconseil\Desktop\tmp" 
  --hide-crash-restore-bubble 
  --test-type 
  --no-default-browser-check 
  --no-first-run 
  --disable-background-networking 
  --disable-component-update 
  --disable-search-engine-choice-screen 
  --disable-features=OptimizationGuideModelDownloading,OptimizationHintsFetching,OptimizationTargetPrediction,OptimizationHints 
  --proxy-server="http://localhost:6000"
```

This allows me to:

- Start recording in K6 Studio (backend URL on `http://localhost:5000`).
- Launch my custom Chrome shortcut.
- Navigate to my frontend at `http://localhost:3000`.
- Use the app normally while K6 captures backend requests correctly.

It would be very useful if K6 Studio allowed users to specify additional custom Chrome flags when launching the Recorder. That way, this workflow could be supported out of the box without manual shortcuts. I can see this just under `Browser Path` Textbox. I've seen how the browser is spin up in the codebase I think we can use the spread syntax to append them.

_No response_

Lileeid:
## Description

Adds support for configuring additional Chrome/Chromium command-line arguments in the Recorder settings.

Users can enter one argument per line. The configured arguments are persisted in the application settings and passed to Chrome/Chromium when k6 Studio starts a recording session.

To protect the recording flow, arguments that conflict with switches managed internally by k6 Studio, such as `--proxy-server` and `--user-data-dir`, are rejected by validation.

Validation for other arguments is intentionally minimal. Custom Chromium arguments may affect browser security, browser behavior, or recording stability, so the settings UI displays a warning explaining that their use is the user's responsibility.

The default value is empty, preserving the existing browser launch behavior for users who do not configure additional arguments.

This PR also includes:

* A settings schema version update and migration for existing settings.
* Validation for malformed and internally managed browser switches.
* A warning when custom browser arguments are configured.
* Tests covering the settings migration and argument validation.

### Screenshot

image

## How to Test

1. Open the k6 Studio settings.

2. Navigate to the Recorder settings.

3. Enter one or more Chromium arguments, one per line, for example:

   ```text
   --proxy-bypass-list=<-loopback>
   --ignore-certificate-errors
   --user-agent=Custom browser agent
   ```

4. Save the settings and reopen them.

5. Verify that the configured arguments are preserved.

6. Start a recording and verify that Chrome/Chromium launches successfully with the configured arguments.

7. Remove all custom arguments and verify that recording continues to work with the existing default behavior.

8. Enter an argument managed by k6 Studio, such as:

   ```text
   --proxy-server=http://localhost:8080
   ```

9. Verify that the settings validation rejects the conflicting argument.

10. Run the relevant automated tests

## Checklist

* [x] I have performed a self-review of my code.
* [x] I have added tests for my changes.
* [x] I have commented on my code, particularly in hard-to-understand areas.

## Related PR(s)/Issue(s)

Resolves #852


src/schemas/settings/v6/index.ts
allansson, line 1: The new schema version is unnecessary since adding a property can be done in a non-breaking way. Add the new property to the `RecorderSettingsSchema` in `v5` instead and make it `optional().default([])`.

```
// src/schemas/settings/v1/index.ts
export const CustomBrowserLaunchArgsSchema = z
  .array(z.string()) 
  .transform((args) => args.map((argument) => argument.trim()).filter(Boolean))
  .superRefine((args, ctx) => {
    const message = findCustomBrowserArgumentError(
      args,
      ALWAYS_MANAGED_BROWSER_SWITCHES
    )

    if (message) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message,
      })
    }
  })

const BaseRecorderSettingsSchema = z.object({
  chromeLaunchArgs: CustomBrowserLaunchArgsSchema.optional().default([])
})

const RecorderDetectBrowserPathSchema = BaseRecorderSettingsSchema.extend({
  detectBrowserPath: z.literal(true),
  enableBrowserRecorder: z.boolean().default(true),
})

const RecorderBrowserPathSchema = RecorderDetectBrowserPathSchema.extend({
  detectBrowserPath: z.literal(false),
  browserPath: z.string().optional(),
})
```

Note that I removed the default from `CustomBrowserLaunchArgsSchema` and added it on the `chromeLaunchArgs` property instead. By calling `default` after `optional` we make the property optional when read from disk but non-optional in the code, which is what we want.

src/components/Settings/RecorderSettings.tsx
```diff
@@ -96,6 +102,72 @@ export const RecorderSettings = () => {
...
+            label="Additional Chrome/Chromium arguments"
+            errors={errors}
+            hint="Enter one Chromium command-line switch per line. Values may contain spaces and do not require shell quotes"
+            hintType="text"
+          >
+            <TextArea
+              placeholder={
+                '--proxy-bypass-list=<-loopback>\n' +
+                '--ignore-certificate-errors\n' +
+                '--user-agent=Custom browser agent'
```
allansson: I think it would be better if the text said "`e.g. --user-agent=Custom browser agent`" otherwise the user might mistake the placeholder for default values.

src/components/Settings/RecorderSettings.tsx
```diff
@@ -96,6 +102,72 @@ export const RecorderSettings = () => {
...
+        <Callout.Root color="amber" mb="4">
+          <Callout.Icon>
+            <AlertTriangleIcon />
+          </Callout.Icon>
+          <Callout.Text>
+            Custom browser arguments may reduce browser security or interfere
+            with recording.
+            <br />
+            k6 Studio managed switches, such as <Code>--proxy-server</Code> and{' '}
+            <Code>--user-data-dir</Code>, cannot be overridden
```
allansson, line 146: @radix-ui/themes already has a `Code` component which should be used here.

src/utils/browserLaunchArgs.ts
```diff
@@ -0,0 +1,96 @@
...
+  customArgs: readonly string[],
+  managedArgs: readonly string[]
+): void {
+  const message = findCustomBrowserArgumentError(
+    customArgs,
+    buildRuntimeManagedSwitches(managedArgs)
+  )
+
+  if (message) {
+    throw new BrowserLaunchArgumentsError(message)
```
allansson: This could throw `BrowserLaunchError` directly instead of catching and mapping it in the caller.

allansson (changes requested):
Thank you for your contribution.

I think the features makes sense and all in all it looks good. I have left some comments and the biggest blocker is the bump in schema version which is not necessary.

We're currently in the final stages of releasing a new version, so we won't be able to merge this until after that.

Anyway, if you make the changes and re-request a review I can approve and merge after the release is done.

Lileeid:
@allansson Thanks for the detailed review and helpful suggestions. I’ve addressed the feedback and requested another review

PR #1332, opened 2026-07-27
internal(refactor): Make LocatorOptions a recursive data type with frame information

allansson:
## Description

This PR restructures how frames are represented in locators. Previously, information about the frame chain of an element was stored directly on the action, alongside the locator. Since we need both frame chain and locator to identify an element it makes sense to co-locate this information.

This PR turns the locator + array of frame locators into single recursive structure, where the locator can contain a parent locator. It makes the locators entirely self-contained, needing a lot less logic to handle.

Future work that this PR enables:

* Capture both iframes and nested locators in debugger
* Re-use for element highlighting instead of the additional data type needed now

> ⚠️ NOTE: All existing browser tests will break. No migration has been added because the feature has not been released yet so it won't affect end users.

## How to Test

Create and edit a browser test. Add actions with locators targeting frames. Check that the script looks correct and make sure it runs.

Record interactions with elements inside frames. Export a browser test and check that locators are converted correctly.

## Checklist


allansson:
bugbot run

allansson:
bugbot run

Llandy3d (approved):

PR #1336, opened 2026-07-30
build: Update test build version to be one minor ahead of stable release

allansson:
## Description

Test builds are currently always considered as older versions than the latest stable release because a suffixed version is a lower version than a non-suffixed. This will trigger the auto-updater to re-install the latest stable, meaning that if you close the app you will lose the test build and have to install it again.

This PR updates the test build logic to always put the test build version one minor version ahead of the stable release. To do this, we now use `git ls-remote` to fetch the latest version tag and update the package version accordingly (we can't trust `package.version` to be the latest version because a release might happen after the branch is created.)

## How to Test

Do a test build and verify that the version is the latest stable release with minor incremented by 1.

## Checklist

### DIFF
--- a/update_version.py
+++ b/update_version.py
@@ -1,17 +1,39 @@
-import os
 import json
-
+import os
+import re
+import subprocess
 
 version = os.environ['PACKAGE_VERSION']
 
+# package.json on a branch can lag behind the latest published release (e.g. a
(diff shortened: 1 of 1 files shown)


Llandy3d (approved):
nice!

PR #1352, opened 2026-08-06
feat: Export linear multi-tab recordings as a single browser test

e-fisher:
## Description

Recordings that open a new tab and continue there (e.g. a checkout or plan picker that opens in a popup) currently force a choice in the "Select a page" dialog, and the exported test covers only the chosen page. This teaches the export to recognize journeys that move from tab to tab without returning: the opening click gets a `continues in new tab` option, the generated script captures the popup with `context.waitForEvent('page')` and continues on it, and the whole journey lands in one test. The dialog still appears for recordings that branch between tabs.

Making the replay side of this work surfaced a few fixes that ride along:

- Pages obtained through `waitForEvent` are now proxied like any other page and get the session replay recorder injected, since k6 does not apply context init scripts to pages opened by the page itself.
- Includes the strict-CSP session replay fix (pull model over CDP instead of in-page `fetch`), plus fixes for two k6 v2 behaviors that broke it: an empty object argument to `page.evaluate` arrives as `undefined` in the page, and `context.pages()` returns fresh wrapper objects on every call, so pulls are now serialized per page id.
- `getByRole` names are no longer pre-escaped. k6 escapes them itself since v1.5.0, so the old workaround double-escaped names like `"I don't know"` into an `InvalidSelectorError`. This affects all browser tests, not just multi-tab ones.

## How to Test

1. Record a journey on any site where a link with `target="_blank"` opens a new tab, continue a few steps in the opened tab, then stop.
2. Create test > Browser test: no page picker appears and the opening click shows `continues in new tab`.
3. Validate the test: all actions pass and the session replay preview follows the journey into the new tab.

Recordings that return to the first tab after opening the popup still show the "Select a page" dialog.

## Checklist

## Related PR(s)/Issue(s)

Builds on the page picker from https://github.com/grafana/k6-studio/pull/1291. Includes the session replay CSP fix (b44ce285) from the `fix-session-replay-csp` branch; if that ships separately first, this rebases cleanly.


e-fisher:
bugbot run

Llandy3d (approved):

PR #1358, opened 2026-08-13
fix: Recording and session replay break on legacy pages

e-fisher:
## Description

Recording a site built with Prototype.js 1.6 produces no toolbar and no recorded browser actions, and session replay of the same site is blank. The recording script used to run in the page's main world, so the page's overrides of `toJSON`, `Function.prototype.bind` and `Array.from` corrupted both the recorder's event envelopes (silently rejected by the app) and the toolbar rendering, and pages that rewrite themselves with `document.open()` additionally wiped the injected script mid-recording.

- The recording script now runs in a CDP isolated world (the content-script model), so page-level JavaScript can no longer break it.
- `Page.documentOpened` triggers re-injection, so the toolbar and event capture survive `document.open()` rewrites. The runtime is reused across re-injections instead of leaking a connection each time.
- Each in-browser UI feature is wrapped in an error boundary, and a double-injection guard prevents two UI instances from fighting over the DOM (previously an infinite mutation loop that froze the recorded page).
- Invalid browser extension messages are reported to Sentry (validation issues only, never the payload), so this failure class is visible instead of silent.
- Session replay has to stay in the main world (k6 injects init scripts there), so it sidesteps the pollution instead: the event serializer ignores page-added `toJSON` methods, and `Array.from` is pinned to the native so rrweb's stylesheet inlining keeps working. Fixes blank and unstyled replays.

## How to Test

Save as `index.html`, serve with `python3 -m http.server 8000`:

```html
<!doctype html>
<html>
  <head>
    <script src="https://ajax.googleapis.com/ajax/libs/prototype/1.6.0.3/prototype.js"></script>
    <style>body { background: #e8f5e9; font-size: 18px }</style>
  </head>
  <body>
    <input id="user" placeholder="user" />
    <button onclick="document.open(); document.write('<input id=next><button>done</button>'); document.close()">
      rewrite
    </button>
  </body>
</html>
```

1. Record `http://localhost:8000`: the toolbar shows, typing and clicks are recorded. Before: no toolbar, no browser events, `Received invalid message` warnings in the app log.
2. Click `rewrite`: the toolbar survives the rewrite and actions on the new content keep recording. Before: toolbar and capture gone until the next navigation.
3. Generate a browser test from the recording and validate it: the session replay preview is populated and styled. Before: blank preview while the test itself passes.

## Checklist

## Related PR(s)/Issue(s)

Related: https://github.com/grafana/k6-studio/pull/1348 (restores the toolbar after app rehydration; does not cover main-world pollution or `document.open()`)

### DIFF
--- a/scripts/cdp/generateCdpClient.ts
+++ b/scripts/cdp/generateCdpClient.ts
@@ -425,16 +425,19 @@ function generateEventFunctions({ domain, events = [] }: cdp.Domain) {
           '+',
           params.name
         )
-        const listener = new ExpressionBuilder(identifier('listener')).as(
-          typeRef('ChromeEventListener')
-        )
+
+        // The transport is shared by every session, so the session-filtering
+        // wrapper is what gets registered, not the raw listener.
+        const filteredListener = new ExpressionBuilder(
+          identifier('filteredListener')
+        ).as(typeRef('ChromeEventListener'))
 
         return [
           ...body,
           new ExpressionBuilder($this())
             .member('transport')
             .member(name)
-            .call([eventName, listener])
+            .call([eventName, filteredListener])
             .returned(),
         ]
       })
--- a/scripts/cdp/generateCdpClient.ts
+++ b/scripts/cdp/generateCdpClient.ts
@@ -354,26 +354,46 @@ function generateEventFunctions({ domain, events = [] }: cdp.Domain) {
     return []
   }
 
+  // A client bound to a session only forwards the events of that session. A
+  // root client has no session id and forwards everything, including events
+  // that carry one. Dropping those would break the browser-level client, which
+  // receives Target.attachedToTarget with the session id of the parent target
+  // when a popup is auto-attached in flat session mode.
+  //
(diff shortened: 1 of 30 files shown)


e-fisher:
bugbot run

allansson (commented):
The comments made my head spin. I rewrote a bunch of them to make them more concise and, hopefully, easier to understand. I also removed some which I felt were unnecessary.

I also fixed the error in the generateCdpClient and generated a new client, so the TODO that Claude added is no longer.

One caveat with this change is that our code no longer lives in the same world as the page, so we're more limited in what we can do. 

One idea I have is to override `addEventListener` so that we can track which elements received a `click` listener. This would let us detect e.g. `div`s that have click handlers and treat them as clickable.

I think we can do this by loading a separate script into the page's realm though, so it's not a huge deal.

Other than that, great job! My approval isn't worth anything since I pushed some changes. You can approve and merge yourself, @e-fisher.

Llandy3d (approved):

PR #1367, opened 2026-08-25
fix: clarify HTTP search scope tooltip

### LINKED ISSUE #476: Update Search all request data tooltip to reflect the state
The search all request data toggle has the same tooltip regardless of its state.

Update the tooltip to make it clear what happens when the user toggles between the states.

Image

loglapa:
## What this changes

- make the HTTP search tooltip describe the active search scope
- keep tooltip and accessible-label text in sync as the search mode changes
- add regression coverage for both supported scopes

## Testing

- focused Vitest suite (2 tests)
- `npm run typecheck`
- `npm run format:check`
- `npm run lint`
- `git diff --check`

All project commands were run in Docker.

Fixes #476


e-fisher (approved):
LGTM

